PR #3, opened 2020-07-23
Clippy

sanket1729:
Creates temp README.

Adds better github actions

Clippy and 1.31.0 MSRV fixes


sanket1729:
For now, I have a custom policy allow/deny rules, we can change those according to feedback

PR #11, opened 2020-08-13
Add support for Dag representation. 

sanket1729:
This PR consists of two parts:
1) Refactor to disambiguate multiple uses of keyword `Node`. 
2) Implements Dag representation support of simplicity programs.


apoelstra (approved):
Nice

PR #12, opened 2020-08-14
Adds a policy compiler

sanket1729:
Adds a policy compiler. Builds on top of #11 . 

This is PR is majorly for design and overall architecture review. There are lot of unwraps, panics that would be later transferred into sane errors. The compiler is complete but assumes custom jets that are made only for testing. The lifting code is incomplete and only works for leaf policies. We can fix that once the jets are finalized.

### DIFF
--- a/src/core/mod.rs
+++ b/src/core/mod.rs
@@ -16,3 +16,22 @@
 /// Core Module for simplicity
 pub mod term;
 pub mod types;
+
+// handy function for converting bit vector to vec[u8]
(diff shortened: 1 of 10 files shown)


apoelstra (approved):
Gonna go ahead and merge this. I think the pubkey32 stuff will need to evolve as rust-bitcoin changes for Taproot, and this template-matching in `lift` might prove to be unsustainable, such that we'd have to write some macros or something to do it for us.

PR #17, opened 2022-01-25
Safe frame

chlewe:
Removes unsafe code from `frame.rs` (crate is now completely safe).
Bits are copied one by one from one frame to another. More efficient methods are possible, but require longer code _(let's discuss readability vs efficiency)_.

### DIFF
--- a/src/bit_machine/frame.rs
+++ b/src/bit_machine/frame.rs
@@ -148,8 +148,9 @@ impl Frame {
         }
     }
 
-    /// Adds
-    pub fn with_data<'a>(&self, data: &'a [u8]) -> FrameData<'a> {
+    /// Extend the present frame with a read-only reference the the data
+    /// and return the resulting struct.
+    pub fn to_frame_data<'a>(&self, data: &'a [u8]) -> FrameData<'a> {
         FrameData::new(self, data)
     }
 
@@ -321,10 +322,10 @@ mod tests {
     }
 
     #[test]
-    fn test_with_data_iter() {
+    fn test_to_frame_data_iter() {
         let bytes = (0..100).collect::<Vec<u8>>();
         let frame = Frame::new(0, 100 * 8);
-        let bits = frame.with_data(&bytes).collect();
+        let bits = frame.to_frame_data(&bytes).collect();
         let computed_bytes = bitvec_to_bytevec(bits);
 
         assert_eq!(bytes, computed_bytes);
--- a/src/bit_machine/frame.rs
+++ b/src/bit_machine/frame.rs
@@ -15,227 +15,283 @@
 
 //! # Simplicity Frame
 //!
-//! Implementation of the Frame in Simplicity BitMachine.
+//! Implementation of Frames in the Simplicity BitMachine.
 //! A frame is a, possibly empty, cell array with a cursor referencing
 //! a cell in the array.
 
-use std::{fmt, ptr};
+use std::fmt;
+use std::mem::size_of;
+use std::ops::{Add, Shl};
 
-/// A frame used internally by the Bit Machine to keep track of
-/// where we are reading or writing to
+/// Context to access a sub-slice of [`super::exec::BitMachine`]'s data.
+/// Read and write operations require a reference to the data,
+/// as it is not contained in this struct.
+#[derive(Debug, Eq, PartialEq)]
 pub(crate) struct Frame {
-    /// Base pointer to the data
-    pub(in crate::bit_machine) data: *mut u8,
-    /// Current position of the cursor. For read frames this points to
-    /// the next bit which is to be read. For write frames, this corresponds
-    /// to the next bit where data would be written
-    pub(in crate::bit_machine) abs_pos: isize,
-    /// Start index in of this frame in data
-    pub(in crate::bit_machine) start: isize,
+    /// Current position of the cursor.
+    /// For read frames, this is the next bit which is to be read.
+    /// For write frames, this is the next bit which is to be (over)written.
+    pub(crate) cursor: usize,
+    /// Start index of this frame in the referenced data.
+    pub(crate) start: usize,
     /// The total length of this frame.
-    pub(in crate::bit_machine) len: isize,
+    pub(crate) len: usize,
 }
 
-impl fmt::Debug for Frame {
-    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
-        unsafe {
-            for i in 0..self.len {
-                if i == self.abs_pos - self.start {
-                    f.write_str("^")?;
-                }
+impl Frame {
+    /// Create a new frame that starts at the given index and that is of given length.
+    pub(crate) fn new(start: usize, len: usize) -> Self {
+        Frame {
+            cursor: start,
+            start,
+            len,
+        }
+    }
 
-                let p = self.data.offset((self.start + i) / 8);
-                if *p & (1 << (7 - (self.start + i) % 8)) != 0 {
-                    f.write_str("1")?;
-                } else {
-                    f.write_str("0")?;
-                }
+    /// Reset the cursor to the start.
+    pub(crate) fn reset_cursor(&mut self) {
+        self.cursor = self.start;
+    }
+
+    /// Return the current bit.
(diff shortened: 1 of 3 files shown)


apoelstra:
Welcome @chlewe!

We currently have a MSRV of 1.31. On the phone recently I said we'd be happy to increase this, since rust-simplicity is something of a bleeding-edge crate and doesn't need to be as conservative as rust-bitcoin. So we will need to "officially" bump the MSRV by changing the version in CI -- and I will need to do some manual work in Github to allow this.

What is the minimum Rust version that you need for these changes? You can test locally by running e.g. `rustup install 1.51.0` then `cargo +1.51.0 test`.

cc @sanket1729 I think we want to bump the MSRV on this crate. Do you have any opposition?

chlewe:
Our dependency `byteorder` causes CI to fail: The crate requires `try_from` which was stabilised in Rust 1.34.0. My code doesn't require any new features and tests run fine on Rust 1.34.0. That being said, let's bump the MSRV to something reasonable like 1.51.0 so that dependencies don't cause breakages in the near future.

apoelstra:
There is no "reasonable" value that will not cause dependencies to break. Dependencies in the rust ecosystem are not dependable.

Let's stick with 1.41, which matches rust-bitcoin, since that is sufficient for now.

sanket1729:
I think @chlewe is on point here. This essentially boils down to performance vs safety. I think this may be a good point to decide the direction for `rust-simplicity`. Should we be aiming for efficiency at the cost of safety? 

If I recall correctly there were two main reasons why this unsafe was there in the first place. 
- The initial reason was to mimic the C implementation for easier debugging. I could check the exact buffers easily. I don't think this is a strong reason anymore. 
- It does only one allocation vs allocations for each new frame creation. For a simple bench for executing a sha256 hash block, the performance penalty is about 15-20%. 
Without this PR
```
running 1 test
test test_progs::benches::bench_hash ... bench:  27,852,396 ns/iter (+/- 711,968)
```

With this PR
```
running 1 test
test test_progs::benches::bench_hash ... bench:  32,237,754 ns/iter (+/- 2,494,923)
```
- Using a single buffer clearly uses the Dos limits: This current PR loses the following bound on the maximum size of cells during an execution. Although this is a guarantee of simplicity programs to execute within the desired number of cells, it is nice to clearly see enforced.
```diff
- data: vec![0; (io_width + prog.extra_cells_bound + 7) / 8],
```

> More efficient methods are possible, but require longer code

I did not think too deep about this, but I don't see any clear way to have safety while avoiding owning data in Frame.  I am in support of any complete safe code that can improve the performance of the bit machine. 

I am on the fence about valuing efficiency here because
- Verification efficiency is great to have for blockchain programs because this ultimately decides how fast you can verify a block. 
- This code will likely never change, so we would not have to worry about it rotting.

I would not complain if we make Frame safe and take the performance/allocation hit. I also dislike unsafe code and getting rid of raw pointers would be good. 
cc @apoelstra

sanket1729:
ACK for improving the MSRV to 1.41

apoelstra:
The `fuzz` directory does not build with this PR.

chlewe:
Thanks, @sanket1729, for your comment. I begin to see how, in the original code, `BitMachine` owned its data and how `Frame`s were a view on a slice of said data. It's also helpful to allocate memory up-front when we know an upper bound.

Could you tell me how I can reproduce your benchmarks on my end? The files seem to stem from outside the present repository.

I have an idea how to mimic the original code in a safe way: We keep the data in `BitMachine` and pass a reference to said data to `Frame` on each read/write call. This way, we avoid pointers while preserving the benefits of the original code. I would like to run benchmarks to see how performance is affected.

sanket1729:
@chlewe the branch is really a hack commit that I made in https://github.com/sanket1729/rust-simplicity/tree/temp_bench . You can get the benchmarks at branch HEAD and at branch HEAD~1 respectively. 


> We keep the data in BitMachine and pass a reference to said data to Frame on each read/write call

Is it possible to get multiple mutable references to different parts of the vector?

chlewe:
Thanks, I will try the benchmarks later today.

All operations of `BitMachine` work on one frame only, except for `copy`. For `copy`, the alignment of both src and dst frame is paramount: If either frame starts _inside_ some byte _(0 < bit offset < 8)_, then we must copy bit by bit. The code on `master` already does this. For this, we need only one reference, storing the read byte(s) in some variable. If both frames are aligned _(bit offset = 0)_, then we can memcopy. I would use something like [`split_at_mut`](https://doc.rust-lang.org/std/primitive.slice.html#method.split_at_mut) to split the mutable slice _(we know the frames are distinct)_ and [`copy_from_slice`](https://doc.rust-lang.org/std/primitive.slice.html#method.copy_from_slice) to do the copying.

src/bit_machine/frame.rs
```diff
@@ -15,227 +15,283 @@
...
+            self.write_bit(bit, data);
         }
     }
 
-    /// Write a big-endian u16 value to the current write frame
-    pub(in crate::bit_machine) fn write_u16(&mut self, data: u16) {
-        for idx in 0..16 {
-            self.write_bit(data & (1 << (15 - idx)) != 0);
+    /// Adds
+    pub fn with_data<'a>(&self, data: &'a [u8]) -> FrameData<'a> {
```
sanket1729, line 152: nit: This is a constructor of FrameData inside `impl Frame`. Perhaps move to this `impl FrameData` ? Or call this function `to_frame_data` ?

sanket1729 (commented):
This looks great. Left one minor nit.

code review ACK 8b4ae9dffe82912318d4b086a68ea0eeb9adc43f . I like the data structure changes. Bye bye unsafe!

sanket1729:
This can be rebased now that #18 has been merged

sanket1729 (approved):
ACK db58407344bc889aa8589e8adf1677896f681d5b. Left one minor nit. Feel free to update it here or address it in a later PR (up to you)

src/bit_machine/frame.rs
```diff
@@ -15,227 +15,283 @@
...
+    where
+        T: From<u8> + Shl<usize, Output = T> + Add<Output = T>,
+    {
+        let (mut self_byte_index, self_bit_index) = get_indices(self.cursor);
+        let mut read_number;
+
+        let number_bits = size_of::<T>() * 8;
+        let number_leading_bits = 8 - self_bit_index;
+        let number_full_bytes = (number_bits - number_leading_bits) / 8;
+        let number_trailing_bits = number_bits - number_leading_bits - number_full_bytes * 8;
```
apoelstra, line 167: In 372e8a313303e67ab42dc2fe3879850225ffe52b:

nit: it would be a bit more idiomatic to just write `(number_bits - number_leading_bits) % 8` rather than subtracting. (It is a common idiom to define something by `X / Y` then immediately following do `X % Y').

This is a very cool use of Rust generics. I should be more open to trying to use the numeric generics, which I tend to assume are useless because of the lack or sane mathematical guarantees (e.g. the stdlib implements `Add` non-commutatively on some types) and the inability to obtain a generic zero or one.

src/bit_machine/frame.rs
```diff
@@ -15,227 +15,283 @@
...
+    pub(crate) fn move_cursor_backward(&mut self, len: usize) {
+        self.cursor -= len;
     }
 
-    /// Write a big-endian u32 value to the current write frame
-    pub(in crate::bit_machine) fn write_u32(&mut self, data: u32) {
-        for idx in 0..32 {
-            self.write_bit(data & (1 << (31 - idx)) != 0);
+    /// Copy a bit string of given length from another frame into the present one.
+    pub(crate) fn copy_from(&mut self, other: &Self, len: usize, data: &mut [u8]) {
```
apoelstra, line 143: In 372e8a313303e67ab42dc2fe3879850225ffe52b:

nit: maybe add a `debug_assert` that the read and write frames don't overlap

apoelstra (approved):
ACK 3adc2035add4099276697b88e622790ec434f554

Neither of my nits are important. I think this is good to merge.

sanket1729:
It's sad that we have to pass data parameters for all read/write peek APIs, but this is definitely worth it because we have safety.

PR #18, opened 2022-01-26
A couple CI improvements

apoelstra:
Bump MSRV and add fuzztests to CI.


sanket1729:
Okay, I can see how clippy is really annoying here. The same code that passed Clippy checks in 2020 is failing now. I guess they added more checks. 

Seems like fuzz is already broken in the existing code

apoelstra:
Fixed clippy and fuzztest compilation. But the fuzztests fail now (`rtt_program` fails with `6b0f7f09000000000000000eff5500`) so I guess my compilation test fixes were incorrect :)

sanket1729:
I will investigate into this issue

.github/workflows/main.yml
```diff
@@ -75,3 +75,35 @@ jobs:
...
           command: clippy
           args: -- -D warnings
+
+  fuzz:
+    name: Fuzz Tests
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - nightly
```
sanket1729, line 85: Hongfuzz frequently breaks on nightly, so all repos in the rust-bitcoin ecosystem switched to stable for fuzzing. The CI also fails to download dependencies in nghtly here. I think we should also change this to stable

sanket1729:
I may have never ran the fuzz tests again. And this is probably some of my old code causing the crash. Looking into it

apoelstra:
@sanket1729 I think this is ready to review now. It seems that the fuzz test failures were quite simple bugs in the code.

apoelstra:
Lol, sigh, now honggfuzz isn't compiling on stable.

sanket1729:
Interesting, I think rust-bitcoin, rust-miniscript both use hongfuzz on stable now.

sanket1729:
I am able to build it locally on 1.58 ..

sanket1729:
@apoelstra, you need to install these. 
See rust-miniscript actions file for reference : https://github.com/rust-bitcoin/rust-miniscript/blob/b8d04accb1e5fcf02849113531783a860ef42bbc/.github/workflows/rust.yml#L16-L17

apoelstra:
CI passes :)

sanket1729 (approved):
ACK 31724a30be054ffc445b12250bdccb377270352e. We can have another PR to update these with C implementation

PR #19, opened 2022-02-22
Term tools

chlewe:
Adds methods to construct arbitrary Term DAGs inductively. As a consensus requirement, the resulting Simplicity programs must be compressed to have maximal sharing. To this end, Merkle roots received their own module, separate structs for each root and common traits to share methods. The serialization of values was also changed _(same bit string, different representation in Rust)_.

Removes annotated Merkle roots (AMR) and replaces them with identity Merkle roots (IMR), because the latter appears in the whitepaper while the former does not (seems more canonical).

The current implementation suffers from slow performance (1 order of magnitude slower during `cargo test`). Test case `extension::elements::jets::tests::sighash_all_cmr` is especially affected. I am in the process of debugging why this is, using a profiler. At the moment, I suspect that [hashing `FinalType` for our `HashMap`](https://github.com/chlewe/rust-simplicity/blob/term-tools/src/program.rs#L230) is taking too long. Maybe, we need AMRs, after all. Open to ideas.


src/extension/bitcoin.rs
```diff
@@ -189,35 +190,61 @@ impl extension::Jet for BtcNode {
...
+            BtcNode::ScriptCMR => {
+                Cmr::tag_iv(b"Simplicity-Draft\x1fPrimitive\x1fBitcoin\x1fscriptCMR")
             }
-            BtcNode::InputValue => Cmr::new(b"SimplicityPrimitiveBitcoinx1finputValue"),
-            BtcNode::InputSequence => Cmr::new(b"SimplicityPrimitiveBitcoinx1finputSequence"),
-            BtcNode::NumOutputs => Cmr::new(b"SimplicityPrimitiveBitcoinx1fnumOutputs"),
-            BtcNode::TotalOutputValue => Cmr::new(b"SimplicityPrimitiveBitcoinx1ftotalOutputValue"),
-            BtcNode::OutputValue => Cmr::new(b"SimplicityPrimitiveBitcoinx1foutputValue"),
-            BtcNode::OutputScriptHash => Cmr::new(b"SimplicityPrimitiveBitcoinx1foutputScriptHash"),
-            BtcNode::ScriptCMR => Cmr::new(b"SimplicityPrimitiveBitcoinx1fscriptCMR"),
```
chlewe, line 213: At the moment, the main method of this crate fails because the CMRs of Simplicity-Bitcoin programs are different. I added `\x1f` where it was missing. This seems to be in agreement with the whitepaper. Should I simply update the CMRs that are hard-coded in `main.rs`?

apoelstra: Yep, that sounds like the right thing to do.

apoelstra:
> I was able to restore performance by manually implementing `std::hash::Hash` for `FinalType`, using its TMR. Running `cargo test` seems to be as fast as on `master`.

Good catch! We should maybe audit any other derived traits on recursive types.

apoelstra:
ACK 2730fe5918edbe9e5eddd6a3dfdc0d5d71bfef94. (That's the second-to-top commit.) Still have the final commit to go.

Overall this approach looks great. In future, I think it would be helpful to split this into multiple PRs -- the first three commits, which just rename and move some stuff, I was able to easily review in a couple of minutes. Then there is some heavier refactoring, and the unification of the IMR/CMR stuff, and replacement of WMR with CMR ... all of these could feasibly be split up.

Don't be afraid to have multiple PRs open at once which all build on each other :) it can be annoying to be rebasing all the time but it gets quick feedback and makes life easier for reviewers.

chlewe:
Thanks for the feedback. You're right: The last two commits can easily be split into many smaller commits. Ideally, one commit does one thing. The term stuff can also be split from the Merkle root stuff, as long as no one writes consensus-critical programs with the new constructors _(and no one does)_. I use interactive rebases a lot and commits tend to grow in the process due to `git commit --amend`. I am sure there is a way to split commits during rebasing to achieve the opposite effect.

src/extension/bitcoin.rs
chlewe: Example Schnorr1 is no longer in the C implementation of Simplicity. Instead, we have [Schnorr0](https://github.com/ElementsProject/simplicity/blob/master/C/schnorr0.c) and [Schnorr6](https://github.com/ElementsProject/simplicity/blob/master/C/schnorr6.c). Of course, I could copy the CMRs that Rust produces and fix the assertions, but this doesn't seem right. If we replace the current Schnorr1 with Schnorr0 or Schnorr6, then we can use "official" CMRs from C. As for the Fibonacci program, there doesn't seem to be any trace of it in C. I think, [I found it in the Haskell implementation](https://github.com/ElementsProject/simplicity/blob/master/Haskell/Tests/Simplicity/Programs/Example.hs), but there is no hard-coded CMR. Maybe, @sanket1729 has ideas on how to test the Rust CMRs against C.

apoelstra:
@chlewe splitting commits is a bit annoying to do. During a rebase you can do `git reset HEAD^` to "uncommit" the changes from the current commit. Then you can selectively re-commit them using `git add -p` for example, or by just copy/pasting chunks of code into temporary files.

apoelstra (approved):
ACK ae1580e81f587e3b2959c454f0216bfe9ef41cf9

I notice that some intermediate commits don't compile. In the interest of moving forward, and because this is a young project (and big PR), I won't hold up merging, but in future we should spend some more time iterating on stuff like this.

I'm also not 100% clear on how this API is to be used -- at what stage do you have a `TypedNode` vs a `ProgramNode`, etc. It would be helpful if there were some example code.

chlewe:
I will add further documentation to the various node types, because (I think) I have a solid understanding of what the respective one does. There is a path-way where nodes get increasingly enriched, starting at the untyped `Term` and ending at the fully developed `ProgramNode`. I will also add unit tests for sharing.

apoelstra (approved):
ACK 671fac50d0c6c6fc1587c2f28eb2d95bbf92e516

This looks really great. Looking forward to being able to construct SImplicity programs!

src/program.rs
```diff
@@ -114,27 +118,8 @@ impl<Ext: extension::Jet> Program<Ext> {
...
-                ),
-                frame_count_bound: compute_frame_count_bound(&ret, &node.node, index),
-                node: node.node,
-                source_ty: node.source_ty,
-                target_ty: node.target_ty,
-            };
-            ret.push(final_node);
-        }
-        Ok(Program { nodes: ret })
+        let finalized_program = compress_and_finalize(typed_program);
```
sanket1729, line 135: This should return Err if the `typed_program` is uncompressed. Not having maximal sharing is a dos vector in simplicity and would not be allowed by consensus rules. 

We should enforce in type system only fully compressed programs ever constructed in `Program`. 

We should offer programmatic utilities to help users compress a program, but while decoding programs from the wire, we should error if they are not compressed.

sanket1729: We should revisit our type of safety guarantees for data structures to make sure that users cannot create invalid data structures. 

At the very least, we should do this for `Program`. This would involve removing `pub` from fields of `Program`, and only keeping necessary constructors.

sanket1729 (approved):
ACK 37520ee77ed0ce7fa9ef88dd64f78dece16c2d2d. Did not review the tests
- Left a comment about mandatory sharing, that can be addressed in a future PR. No need to redo this to invalidate ACKs.
- I think we should make a small Glossary of all Data structures and stick in README or add it as another doc. It would save a lot of time if we knew `TermDag`, `TypeProgram`, `Program`, `TypedNode`, `UnTypedNode`.

sanket1729:
ACK 671fac50d0c6c6fc1587c2f28eb2d95bbf92e516

PR #22, opened 2022-04-13
Simplicity bindings

chlewe:
Exposes part of [Simplicity's C API](https://github.com/ElementsProject/simplicity/tree/master/C) in Rust. At the moment, one can decode Simplicity DAGs from bytes, and one can access test programs with hardcoded bytes and their correct Merkle roots.

# Todo
- [x] Discussion: use minimal set of necessary C files instead of entire library?
- [x] Discussion: use C test files?
- [x] ~~test CMRs and IMRs against C~~ [postponed to later PR due to outdated jets]
- [x] Make sure that FFI is safe and that pointers are freed


chlewe:
Please check if this is the way we should expose C bindings, @apoelstra, @sanket1729. I mimicked [secp256k1-sys](https://github.com/rust-bitcoin/rust-secp256k1/tree/master/secp256k1-sys) with important differences:
1. There is no renaming / tagging of functions.
2. The crate is called `simplicity_sys`, since hyphens are not allowed in crate identifiers.

simplicity-sys/build.rs
```diff
@@ -0,0 +1,44 @@
...
+        "typeInference.c",
+        "primitive/elements/env.c",
+        "primitive/elements/exec.c",
+        "primitive/elements/jets.c",
+        "primitive/elements/primitive.c",
+    ]
+    .into_iter()
+    .map(|x| simplicity_path.join(x))
+    .collect();
+    let test_files: Vec<_> = vec![
```
apoelstra, line 27: In cdee8cba155909bf562aa287b1a0ef3d5318e471:

I don't think we need this array (or these files, for that matter).

simplicity-sys/src/bitstream.rs
```diff
@@ -0,0 +1,44 @@
...
+/// Corresponds to `bitstream` in `depend/simplicity/bitstream.h`.
+#[repr(C)]
+pub struct Bitstream {
+    file: *mut FILE,
+    available: c_int,
+    byte: c_uchar,
+}
+
+impl From<&[u8]> for Bitstream {
+    fn from(bytes: &[u8]) -> Self {
```
apoelstra, line 28: In 8e45d76bc0e528aed678978f5111b76d2b23063a:

This entire function needs to be marked `unsafe`, or better, you need to tie the lifetime of `bytes` to the `Bitstream` type (perhaps with one of the types in `std::marker`). As written it takes an array `bytes` and then returns an object pointing into `bytes`' data which might outlive `bytes`.

simplicity-sys/src/dag.rs
```diff
@@ -0,0 +1,120 @@
...
+
+/// (First node of) a Simplicity DAG (typed or untyped).
+///
+/// Corresponds to `dag_node` in `depend/simplicity/dag.h`.
+#[repr(C)]
+pub struct DagNode {
+    jet: *const c_void,
+    cmr: [u32; 8],
+    // Assume largest union member
+    union_aux_types: [size_t; 2],
```
apoelstra, line 52: In 8e45d76bc0e528aed678978f5111b76d2b23063a:

This makes me a little nervous. Rust has a `union` keyword. Could you use that?

simplicity-sys/build.rs
apoelstra: Ah -- you have unit tests in -sys that use them. In this case I think you could gate this array with `#[cfg(test)]` so that it's not built when simplicity-sys is used as a dependency.

apoelstra:
Regarding "should we try to minimize the set of C", I vote no. The whole C library is super small; better to use the whole thing to make review easier. Link-time optimization will strip unused symbols from the final binaries.

Done reviewing 8e45d76bc0e528aed678978f5111b76d2b23063a. Overall this looks great!

simplicity-sys/build.rs
chlewe: We could reduce the built files to a minimum. For instance, in order to construct C-DAGs from binary, we definitely need `decodeMallocDag()` from `dag.c`.

simplicity-sys/src/dag.rs
```diff
@@ -0,0 +1,120 @@
...
+        stream: *mut Bitstream,
+    ) -> i32;
+}
+
+impl DagNode {
+    /// Decode a Simplicity DAG from bytes.
+    /// Returns the DAG and its length on success.
+    pub fn decode(bytes: &[u8]) -> Result<(&mut DagNode, usize), Error> {
+        let mut bitstream = Bitstream::from(bytes);
+        let mut dag: *mut DagNode = std::ptr::null_mut();
```
chlewe, line 71: Do we have to explicitly free the DAG in the destructor of `DagNode`? [In the C code, the DAG is freed after use.](https://github.com/ElementsProject/simplicity/blob/master/C/test.c#L152)

simplicity-sys/src/bitstream.rs
```diff
@@ -0,0 +1,49 @@
...
+    byte: c_uchar,
+    _marker: PhantomData<&'a u8>,
+}
+
+impl<'a> From<&'a [u8]> for Bitstream<'a> {
+    fn from(bytes: &'a [u8]) -> Self {
+        unsafe {
+            let mode = "r";
+            // `file` does not need to be freed as
+            // `bytes` will free itself at the end of its lifetime
```
apoelstra, line 34: In 213c69795b5c00d46bd593c3d7af764fa539e973:

This comment doesn't make sense to me. At some point you do need to `fclose` `file`, don't you?

simplicity-sys/src/dag.rs
apoelstra: Yep, you need to free `dag` in the `Drop` impl of `DagNode`. Good catch.

apoelstra:
Briefly reviewed 213c69795b5c00d46bd593c3d7af764fa539e973. (Only changes from the last review were in the final commit; I think we'll need a bit more iteration on this one.)

simplicity-sys/src/bitstream.rs
chlewe: I updated `Bitstring` to `Bitstream<'a>` where `'a` is the lifetime of `&bytes`.

simplicity-sys/src/dag.rs
chlewe: `DagNode` now uses unions, namely `AuxTypes` [[1]](https://github.com/chlewe/rust-simplicity/blob/simplicity-sys/simplicity-sys/src/dag.rs#L46) and `IndicesChildrenWitness` [[2]](https://github.com/chlewe/rust-simplicity/blob/simplicity-sys/simplicity-sys/src/dag.rs#L54).

simplicity-sys/src/bitstream.rs
chlewe: `Bitstream<'a>` lives at most as long as `&'a bytes`. According to man pages, `fmemopen` does not open a file descriptor, nor does a read-only stream have any buffers that need to be flushed. `fclose` simply closes file descriptors and flushes. If I understand correctly, this means that the `FILE` pointer in the stream can safely be dropped without freeing it.

That being said, not freeing a `FILE` pointer is probably bad practice. I will add a destructor.

apoelstra: I'm sure that you need to actually call `fclose` at some point in this source file.

Otherwise, the lifetime stuff looks like it's been fixed.

apoelstra: Actually, on further reading I don't know. I can't find a citation either way. It is surprising that there is no descriptor and that `fileno` will just fail.

simplicity-sys/build.rs
chlewe: The point of this PR is to expose (1) a way to decode DAGs from bytes in C, and (2) test programs with hardcoded bytes and Merkle roots. This is why we need `test_files`. My original plan was to test Rust against C, but this has to be done in a future PR due to outdated jets.

chlewe:
I updated how DAGs are represented, and added `free` and `fclose` for each allocation and opened stream.

Because `Drop` now tries to free a DAG when it goes out of scope, we cannot do pointer addition to index the DAG and return a struct of the same type. The fact that DAGs are not their root node is confusing in the first place, so I decided to hide that from the user completely: A DAG has exactly _one_ CMR, namely its root CMR. In future PRs, we can add the notion of DAG nodes, whose lifetime will depend on that of their parent DAG.

chlewe:
@apoelstra This should be ready to be merged. Sorry for the many force pushes.

simplicity-sys/src/dag.rs
```diff
@@ -0,0 +1,152 @@
...
+}
+
+/// Simplicity DAG (typed or untyped).
+/// Uses the Elements extension by default.
+pub struct Dag {
+    first: *mut PrivateDagNode,
+    len: usize,
+}
+
+impl<'a> fmt::Debug for Dag {
```
apoelstra, line 87: In a0faaf2957e71fc5e72a23d41e916703e739bb53:

nit: this `'a` shouldn't be there. I'm not sure why the compiler doesn't warn about this.

apoelstra:
Could you also rebase onto master? This currently doesn't have #23 which is making it hard for me to test locally.

apoelstra (approved):
ACK 190a669a4a5c328365579d51b9401ab485995914

simplicity-sys/src/test.rs
```diff
@@ -0,0 +1,84 @@
...
+    static hashBlock_cmr: [u32; 8];
+    static hashBlock_imr: [u32; 8];
+
+    static elementsCheckSigHashAllTx1: [u8; 1151];
+    static elementsCheckSigHashAllTx1_cmr: [u32; 8];
+    static elementsCheckSigHashAllTx1_imr: [u32; 8];
+}
+
+/// Simplicity program with bytes, CMR and IMR for testing.
+pub struct TestProgram {
```
sanket1729, line 36: Where are we testing the decoding works?

simplicity-sys/src/dag.rs
```diff
@@ -0,0 +1,152 @@
...
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::test::SCHNORR0;
+
+    #[test]
+    fn decode_cmr() {
+        let bytes = SCHNORR0.bytes;
+        let dag = Dag::decode(bytes).expect("decoding");
+        assert_eq!(SCHNORR0.cmr(), dag.cmr());
```
sanket1729, line 150: We should also test that calculation for cmr from SCHNORR0 from rust-simplicity is same as C simplicity. What we are testing here is that is the value exported from C is correctly calculated from C.

sanket1729 (commented):
high level review ACK. Left a couple questions

simplicity-sys/src/dag.rs
chlewe: I agree; this was my original intention of this PR. Unfortunately, outdated jets etc. make it so the Rust CMR is almost always different than the C CMR _(for the same program)_. Rust's `SCHNORR0` and C's `SCHNORR0` are also very different programs, the later using jets and being much smaller. In a later PR, I will replace `src/test_progs` in Rust with tests that use the FFI once there is a prospect that these tests will succeed.

apoelstra (approved):
ACK 66135ceb7b83d51a6c11a9396bf30b9080a6701c

sanket1729 (approved):
utACK 66135ceb7b83d51a6c11a9396bf30b9080a6701c. The build script seems fragile. We should have some vendoring-like script that updates the underlying library when anything changes.

apoelstra:
Let's merge this for now. We can improve things later to write a revendoring script that will do codegen and maybe symbol renaming, but for now this is something we've wanted for a while.

PR #21, opened 2022-04-04
Elements Miniscript rev 2022-03

chlewe:
Update to latest combination of `rust-bitcoin`, `rust-elements` and `rust-miniscript`.

# Todo
- [x] Update `Cargo.toml` after official release of `rust-bitcoin` v0.28.0, `rust-miniscript` v7.0.0 and `rust-elements` v0.19.0 (?)

PR #23, opened 2022-04-19
Fix features

chlewe:
Remove dependency of (implicit) feature `elements` upon `bitcoin` by using `elements::bitcoin`. Fix GitHub CI to test for successful compilation using each of these features separately.

Proposed in #21.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -46,12 +46,12 @@ jobs:
           toolchain: ${{ matrix.rust }}
           override: true
       - name: Running cargo test
-        env: 
-          FEATUERS: bitcoin elements
+        env:
+          FEATURES: bitcoin elements
         run: |
-          for f in $FEATURES; do [cargo test --no-default-features --features=f]; done
-          cargo test --no-default-features
-          cargo test --all-features
+          for f in $FEATURES; do echo "Features: $f" && cargo test --no-default-features --features="$f"; done
+          echo "No default features" && cargo test --no-default-features
+          echo "All features" && cargo test --all-features
 
   clippy:
     name: Clippy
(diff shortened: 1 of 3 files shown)


sanket1729:
I think there are two different ways of looking at features here:

>  there is no way how dependency / implicit feature elements can enable dependency / implicit dependency bitcoin.


1) Features as dependency tree: I think when you say there is no way to have elements without bitcoin, you mean making sure that the rust-bitcoin crate is not being used as a dependency. 

2) Features are simplicity extensions: In this way of separation, when we are using simplicity with elements extension, we can use both rust-elements and rust-bitcoin. But the jets are available only for elements intropsection. 

I think we should classify our features by the latter method. If possible in rust, we should also mark features as incompatible with each other. That is, you can run in both elements and the bitcoin feature.

apoelstra:
@sanket1729 unfortunately we can't make features incompatibile with each other, so we have made sure that it is possible to have both features enabled.

The problem that this PR addresses is that right now you can't have `elements` *without* `bitcoin`

apoelstra:
@chlewe I would like it to be possible to use `elements` without `bitcoin`. This is what we want for use with elements-miniscript, for example.

Making the dependency explicit fixes the compilation issue but I'd like to remove the dependency. Heads up that you can access the rust-bitcoin crate through rust-elements, as `elements::bitcoin`.

chlewe:
As far as I can see, the namespaces of features and dependencies get properly separated in [Rust 1.60](https://blog.rust-lang.org/2022/04/07/Rust-1.60.0.html#new-syntax-for-cargo-features). In earlier versions, implicit features cannot enable other features / dependencies because implicit features are defined in `[dependencies]` and not in `[features]`. [Incompatible features should be avoided](https://doc.rust-lang.org/cargo/reference/features.html#mutually-exclusive-features).

apoelstra:
@chlewe the two features are not incompatible, are they?

This fix looks great, exactly what I was looking for. If there are actual incompatibilties we should find and fix those but they can wait..

apoelstra (approved):
ACK 0385213c394e39423b8b6e273901f67602820f88

chlewe:
Right now, all features are compatible and no feature depends on another.

PR #24, opened 2022-04-26
Encode / decode

chlewe:
Updates how Simplicity programs, nodes and witness values are decoded from binary. Updates and complete how the above are encoded into binary. Fixes some long-standing todo's in the decoding code. Attempts to simplify some code _(large match cases)_.

### DIFF
--- a/src/encode.rs
+++ b/src/encode.rs
@@ -14,66 +14,39 @@
 
 //! # Encoding
 //!
-//! Functionality to encode Simplicity programs. These programs
-//! are encoded bitwise rather than bytewise, so given a hex dump of a
-//! program it is not generally possible to read it visually the way you
-//! can with Bitcoin Script.
+//! Functionality to encode Simplicity programs.
+//! These programs are encoded bitwise rather than bytewise,
+//! so given a hex dump of a program it is not generally possible
+//! to read it visually the way you can with Bitcoin Script.
 
 use std::{io, mem};
 
-use crate::extension;
-use crate::extension::Jet as ExtNode;
-use crate::Term;
+use crate::extension::Jet;
+use crate::Value;
+use crate::{Term, UntypedProgram};
 
-/// Trait for writing individual bits to some sink
-pub trait BitWrite {
-    /// Write a single bit to the writer
-    fn write_bit(&mut self, b: bool) -> io::Result<()>;
-
-    /// Write out all cached bits, including those written since the last
-    /// byte boundary. (0s will be written after the actual data to pad out
-    /// to the next byte boundary). Then flushes the underlying `io::Write`
-    /// if one exists.
-    fn flush_all(&mut self) -> io::Result<()>;
-
-    /// Total number of bits written to this object (not necessarily the number
-    /// of bits written to the underlying byte-oriented sink, which in general
-    /// will be less)
-    fn n_written(&self) -> usize;
-
-    /// Write several bytes at once, up to 8. If `len` is less than 8, the
-    /// data is read in big-endian order from the least significant bits.
-    /// On success, returns `len`.
-    fn write_u8(&mut self, n: u8, len: usize) -> io::Result<usize> {
-        for i in 0..len {
-            self.write_bit(n & (1 << (len - i - 1)) != 0)?;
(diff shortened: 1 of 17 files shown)


src/encode.rs
```diff
@@ -12,92 +12,66 @@
...
-
-use crate::extension;
-use crate::extension::Jet as ExtNode;
-use crate::{Error, Term};
-
-use crate::core::term::UntypedProgram;
-use crate::merkle::cmr::Cmr;
-
-/// Trait for writing individual bits to some sink
-pub trait BitWrite {
```
chlewe, line 35: Removing `BitWrite` in favour of `BitWriter` is a decision worth discussing. My idea was to mimic the use of `BitIter`, which is also a struct. Any thoughts?

src/encode.rs
```diff
@@ -113,373 +87,203 @@ impl<W: io::Write> BitWrite for BitWriter<W> {
...
-
-    Ok(UntypedProgram(program))
+    Ok(())
 }
 
-/// Encode a natural number according to section 7.2.1 of the Simplicity tech
-/// report. Returns the length of the written number, in bits
-pub fn encode_natural<W: BitWrite>(n: usize, writer: &mut W) -> io::Result<usize> {
+/// Encode a natural number to bits.
+pub fn encode_natural<W: io::Write>(n: usize, w: &mut BitWriter<W>) -> io::Result<()> {
```
chlewe, line 256: Should all functions return the number of written bits, or is it easier to calculate the total number "higher up" using subtraction?

src/encode.rs
apoelstra: Yeah, I think this is a good decision. The trait thing was probably a bit too clever.

apoelstra (approved):
ACK 5583824808f75b26d5990041a6b523619e1e9de4

I had a coulpe tiny nits (e.g. `if` statements that coiuld've been `match`s) but I can't find them now. This all seems good to me. I'm fine with the API changes, including dropping length returns from certain functions (we can always put them back if we find a use for them).

chlewe:
I don't like how the fuzz tests succeed, even though the decoding and encoding of witness data is currently different _(bit length vs value length)_. I need to clarify my understanding of the serialisation and check the code.

apoelstra (approved):
ACK 7853f8ae3619a62852d2e678b0556c3c97e54985

chlewe:
Our fuzz tests only decode and re-encode programs without witness, so my inconsistent code passed. I wrote a fuzz test that decodes and re-encodes 64-bit witnesses, but the API doesn't allow the arbitrary creation of typed programs _(for good reason)_. For now, I added a unit test. Later, we should extend the API with ways to create programs of arbitrary level of enrichment (untyped < typed < finalised) and to enrich / reduce programs along this hierarchy.

chlewe:
At the moment, we use the deprecated jet encoding. This will be fixed in another PR.

apoelstra:
I'm a bit confused -- why can't we decode programs with witnesses?

chlewe:
The code decodes the witness according to the types of a given `TypedProgram`. That program we have to supply, but I don't see any way to construct a struct of such a type using only `pub` functions. We can construct `Program`, but not `TypedProgram`. _(The C code decodes the witness as a bitstring and decodes values later, as witness nodes are updated. I preferred `Vec<Value>` over bitstrings.)_

chlewe:
Our node types are not nestings of each other: `Term` is included in `TypedNode`, but `ProgramNode` does not include `TypedNode`. It's worth thinking about whether we should change that. This would facilitate conversions between kinds of programs. _(In C, a node does everything at once, using union types, but of course, we don't want that in Rust.)_

apoelstra:
Ok, yep, I think I understand. Agreed, we may want to revisit this in the future: I think the public API should probably only allow decoding and encoding of fully typed (and witness-containing) programs. But for now let's get things working :).

apoelstra (approved):
ACK 5330021ab6d0cfc1e01ee59fde2254e001e36c42

apoelstra:
@sanket1729 could you take a quick look at this?

PR #25, opened 2022-06-11
Fixes

chlewe:
Adds fixes that came up during the BTC++ Hackathon on 2022-06-10.


chlewe:
I rewrote the method that converts recursive DAGs into linear programs, in order to understand the conversion better. The new code is more concise, but the resulting linear program may be exponential in the size of the DAG, due to shared `Rc`s. Is this a problem?

src/core/term.rs
```diff
@@ -139,4 +166,70 @@ impl<Witness, App: Application> UntypedProgram<Witness, App> {
     pub fn iter(&self) -> impl Iterator<Item = &Term<Witness, App>> {
         self.0.iter()
     }
+
+    /// Visit the DAG that underlies the program
+    /// and return a stack of node indices that can be popped such that the nodes are in post order.
+    fn get_post_order_stack(&self) -> Vec<usize> {
```
apoelstra, line 172: In 1cef378b9dcf712fa46853b9d96d55e4c865a70a:

This would be a little more idiomatic and efficient as an iterator rather than a method that returns a complete `Vec`. But I like this algorithm.

apoelstra: It would also be simpler to understand -- if we just yielded items on every loop iteration then we wouldn't have a list returned in the reverse order and a `post_order` method being called from a method that needs a pre ordering.

src/core/term.rs
```diff
@@ -139,4 +166,70 @@ impl<Witness, App: Application> UntypedProgram<Witness, App> {
...
+        }
+
+        false
+    }
+
+    /// Return whether the program is in _canonical order_.
+    /// This means that node indices appear in post order,
+    /// i.e., children appear before their parent.
+    /// It also means that if there is a node that cannot be reached from the root,
+    /// then the program is not in canonical order.
```
apoelstra, line 220: In 1cef378b9dcf712fa46853b9d96d55e4c865a70a:

This comment is at least partially incorrect (it's not about children appearing before their parent, which is always trivially true, but about left children appearing before right children). We should also clarify that we **also** check that there are no disconnected nodes, in the last line of the function. As written it looks like the ordering check somehow can detect disconnected nodes by itself.

Curiously the C code seems to have an assertion that `bottom == self.len` and no comment about checking for disconnected nodes ... can you trigger this assertion in the C code?

src/core/term.rs
apoelstra: The same comment applies to 41ae92f368f222738c1b99058f11f723b86e1155 where you implement essentially the same algorithm on `TermDag`.

I also wonder whether we could abstract these into a trait where you have to implement `get_left` and `get_right` and then get the iteration logic for free. In `TermDag` we iterate over `&Self`s and here we iterate over `usize`s but it's easy to make this type an associated type of the trait.

src/core/term_dag.rs
```diff
@@ -116,6 +115,111 @@ impl<Witness, App: Application> TermDag<Witness, App> {
...
+        for dag in to_visit {
+            if self.is_reachable_from(dag) {
+                return true;
+            }
+        }
+
+        false
+    }
+
+    // TODO: Slow for large programs, like HASHBLOCK or SCHNORR
```
apoelstra, line 185: In 41ae92f368f222738c1b99058f11f723b86e1155:

Fine to leave this as a TODO for now. (Though arguably it should be a github issue rather than being a code comment, but that's fine.) A couple solutions, for posterity..
* Memoize -- I think this is probably the best thing. When you convert this to an iterator you can just stick a `HashSet<TermDagWrapper>` where the wrapper type is just something that implements `Eq` and `Hash` based on cmr rather than trying to recurse into the object.
* Stick a `RefCell<bool>` into the `TermDag`s themselves and mark each one `true` when it's first visited...though I can't figure out how to make sure they're all set to `false` at the start of iteration so maybe this is a no-go
* The above, but with pointer tagging rather than an explicit `RefCell<bool>` :). I'm pretty sure this is just impossible in Rust.

src/core/term_dag.rs
```diff
@@ -116,6 +115,111 @@ impl<Witness, App: Application> TermDag<Witness, App> {
...
+    // Memoize reachable nodes? Save unique pairs of parents to leftmost child (requires additional iteration).
+    /// Return whether the DAG root `self` is reachable from the DAG root `other`.
+    ///
+    /// Performs DFS internally.
+    fn is_reachable_from(&self, other: &Self) -> bool {
+        let mut to_visit = vec![other];
+
+        while let Some(node) = to_visit.pop() {
+            if std::ptr::eq(node, self) {
```
apoelstra, line 194: In 41ae92f368f222738c1b99058f11f723b86e1155:

This is a neat trick, but I worry about using pointer identity in Rust. I **believe** this code is correct because nothing will move once it's in an `Rc` [but strictly speaking this is not guaranteed and we'd need to think about pinning rules etc to be sure](https://stackoverflow.com/questions/72148631/how-can-i-hash-by-a-raw-pointer). Are we able to use CMR here instead? I guess not, we'd need to use the IMR which we don't yet have.

src/core/term_dag.rs
apoelstra: Actually, on further thought I think we need to memoize based on pointer identity (as you're doing now) rather than CMR.

apoelstra:
Done reviewing 7603f0f95b5b4c751d4ef29da72bf1bbd2b23340. This looks great! I left a few comments.

apoelstra:
Heads up that the fuzztests do not compile in every commit.

src/core/term_dag.rs
chlewe: I think, there is no way around pointer equality. `TermDag`s have no other kind of identity _(as of now, and we probably don't want to add CMRs to the enum)_.  I tried to define a notion of paths from root, but this approach suffers from the fact that the same shared node can simply have more than one path, depending on where you come from. @sanket1729 's [code also uses pointer equality to find shared DAGs.](https://github.com/ElementsProject/rust-simplicity/blob/master/src/core/term_dag.rs#L169)

src/core/term.rs
chlewe: That's a great idea; I will write an iterator.

src/core/term_dag.rs
apoelstra: Yes, on further reflection I think we should stick with pointer equality here.

And even if we did want to add a CMR this is actually not a sufficient notion of "identity" for this purpose..

src/core/term_dag.rs
apoelstra: Ok. Godspeed.

src/core/term.rs
chlewe: I updated the wording to clarify.

src/core/term.rs
chlewe: We now have the trait `DagLike` that comes with a default implementation of a post-order iterator.

chlewe:
At least on my machine, the fuzz tests (now) seem to compile for every commit. Here is what I ran: `git rebase HASH -x 'cd fuzz;HFUZZ_BUILD_ARGS="--features honggfuzz_fuzz" cargo +1.56 hfuzz build`.

src/core/iter.rs
```diff
@@ -0,0 +1,98 @@
...
+/// A DAG-like structure with a root,
+/// left children and right children per node,
+/// without directed cycles.
+///
+/// We use this trait to implement general DAG operations, like node traversal.
+pub trait DagLike: Sized {
+    /// Node in the DAG
+    type Node: Copy;
+    /// Additional data for internal computations
+    type Extra;
```
apoelstra, line 29: In 925b8cbdb0be693e1fa8d93005eb87d2a9c54846:

Would prefer this be called `Aux` rather than `Extra`

src/core/iter.rs
```diff
@@ -0,0 +1,98 @@
...
+        visited
+    }
+
+    /// Return an iterator that visits the DAG in post order.
+    ///
+    /// Post order means, left children appear before right ones, and children appear before their parent.
+    /// Shared nodes appear only once at their leftmost position.
+    fn iter_post_order(&self) -> Rev<IntoIter<Self::Node>> {
+        self.get_post_order_stack().into_iter().rev()
+    }
```
apoelstra, line 97: In 925b8cbdb0be693e1fa8d93005eb87d2a9c54846:

Why is this split into two functions? I would create a custom iterator type which maintains the stack and just returns elements as they are requested. Also, importing `vec::IntoIter` is confusing because `IntoIter` is already a trait name which is included in the std prelude.

src/core/iter.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// Node in the DAG
+    type Node: Copy;
+    /// Additional data for internal computations
+    type Extra;
+
+    /// Return the DAG root, if the DAG is nonempty.
+    fn root(&self) -> Option<Self::Node>;
+
+    /// Return the left child of the given `node`.
+    fn left_of(&self, node: Self::Node) -> Option<Self::Node>;
```
apoelstra, line 35: In 925b8cbdb0be693e1fa8d93005eb87d2a9c54846:

Would prefer this be called `left_child` and `right_child`

src/core/term.rs
```diff
@@ -1,44 +1,80 @@
...
+/// and terms have no meaning outside a program.
 ///
-/// References to other nodes are relative indices in the context of a program.
-/// For example, node `InjL(2)` at index 7 represents DAG `InjL → x`,
-/// where `x` is the DAG that the node at index 5 represents.
-/// A node has no meaning without a program.
-///
-/// The node representation is later used for executing Simplicity programs.
+/// Nodes refer to other nodes via indices that are relative to the node's own index.
+/// This simulates a DAG-like structure.
```
apoelstra, line 37: In 024f9115bc3f626199ebc464c6cf507bf96449fd:

This is not a simulation, nor is it "DAG-like". It's literally a DAG.

src/core/term.rs
```diff
@@ -1,44 +1,80 @@
...
+/// For example, node `InjL(2)` at index 7 in its program
+/// represents a DAG with root `InjL` whose child is `x`,
+/// where `x` is the DAG that is represented by the node at index 7 - 2 = 5.
```
apoelstra, line 40: If you are going to be this wordy you should say that `x` is the **root** of the DAG ... but you shouldn't be this wordy. Just say "For example, `InjL(2)` at index 7 represents an `InjL` node whose child is the node at index 7 - 2 = 5."

src/core/value.rs
```diff
@@ -1,12 +1,47 @@
...
+// along with this software.
+// If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
+//
+
+//! # Simplicity values
+//!
+//! Simplicity processes data in terms of [`Value`]s,
+//! i.e., inputs, intermediate results and outputs.
+//!
+//! Rust values can be converted to Simplicity values and back.
```
apoelstra, line 21: In 024f9115bc3f626199ebc464c6cf507bf96449fd:

I'd drop this line. It's not true in general.

src/core/iter.rs
chlewe: By two functions, do you mean `into_iter()` and `rev()`? My goal was to avoid a new iterator type if the same functionality can be had using standard functions. Arguably, the return type of `iter_post_order()` turned into something nasty due to that. Is there no way around writing "duplicate" code?

src/core/iter.rs
chlewe: Sure. I had to chose different names for these methods, because they clashed with `TermDag::get_left()` and `TermDag::get_right()`. I feel that the latter are more important and hence should have the "better" names.

chlewe:
Thanks for reading all this stuff. Your suggestions on identifiers and documentation are easy to follow. What is left is the issue of the post-order iterator type, which is merely about aesthetics.

src/core/iter.rs
apoelstra: I mean `get_post_order_stack` and `iter_post_order`. Splitting them up still involves an unnecessary allocation and exposes the user to the reversed logic.

src/core/iter.rs
apoelstra: I don't think it really matters if these names collide, we only call these ones from within this trait's iterator logic, and things will be unambiguous there.

But having said this, why do we have `TermDag::get_left` etc when we already have these methods?

chlewe: The `get_left()` and `get_right()` methods are defined on the `Term`(`Dag`) itself and are there to make pattern matching less verbose. Possibly, one could use these methods for the inductive construction of `TermDag`s, too.

The `left_of()` and `right_of()` methods are defined on `DagIterable` and have a different signature: They take the dag as `self` and then the `node`. These methods are for use inside `DagIterable`. For `Term`, this distinction is important because its children depend on the program.

Therefore, I think that the two sets of methods fulfil different use cases and are both needed.

src/core/iter.rs
chlewe: I realised that having the stack method in a public interface is not ideal, so I moved the logic into a new post-order iterator type. That being said, I don't think there was any unnecessary allocation, unless I am missing something. The stack is necessary and needs to grow to its maximum size until we can pop from it.

src/core/iter.rs
apoelstra: Yeah, I see, the signatures don't match and can't be unified.

src/core/iter.rs
```diff
@@ -0,0 +1,108 @@
...
+
+                if let Some(right) = dag.right_of(node) {
+                    if !dag.is_cross_reference(node, left, Some(right), &to_visit, &extra) {
+                        to_visit.push(right);
+                    }
+                }
+            }
+        }
+
+        PostOrderIter { stack: visited }
```
apoelstra, line 92: In cf8832de5c3d800f6f9a63d9620f537450368908:

This whole thing should be in `PostOrderIter::next`. Then you could maintain `to_visit` in the iterator and not need the `visited` variable at all, saving space as well as a ton of time during iterator construction, and avoid wasting time/space for users who don't want to iterate through the whole thing.

apoelstra:
Reviewed f28dcb4ff71756bde3b353ba9307476fd2bbb64f

src/core/iter.rs
chlewe: I think you misunderstand the algorithm: What `visited` contains is the nodes in the _reverse_ order _(AFAIU, this is a "reverse pre-order", adding to the confusion)_. We visit the parent, its right children and then its left children. This means that in order to get the first node in post-order, we need to visit all nodes. The fastest approach is to store all visited nodes in a stack (`visited`) and to pop from that.

At first glance, it seems wasteful to visit the entire graph in the iterator constructor. However, even for large programs (1k~10k nodes) this is fast thanks to the cross-reference checking. We could make it so that the graph is visited on the first call of `next()`, but I am not sure if this is worth it. In any case, not iterating over the entire graph is _impossible_.

chlewe: To check canonical order, we actually do not need post-order iterators. Reverse-pre-order iterators would work, too _(post order iff successor index is predecessor minus 1)_. In this case, we could write a lazy iterator that only visits as much of the graph as it needs to. Two thoughts regarding this:
(1) To convert `TermDag` to `UntypedProgram`, we need post order. Maybe, we don't want to write two iterators for very similar use cases.
(2) When checking for canonical order, we greedily visit all nodes anyway. A lazy iterator wouldn't add value for this use case.

apoelstra: Right now you're using O(program size) space in this iterator. You should be able to use O(program depth) space.

I see now that this is a little more involved because you'll need to change your algorithm, but I don't see why it should be impossble to do a post-order traversal without first iterating (and storing) the entire program.

chlewe: I suggest that we postpone this optimisation to a future PR. What I used was a standard algorithm that might have sub-optimal space complexity. There might be other algorithms, but I didn't find them over the course of a week. More work seems like diminishing returns to me. We could open an issue.

chlewe: Actually, I take that back. Looking at the [algorithm on Wikipedia](https://en.wikipedia.org/w/index.php?title=Tree_traversal&oldid=1083604462#Post-order_implementation), it seems to do exactly what you wanted. It is O(depth) and lazy, and worth having. I implemented it before, and it seemed slow for some reason, but I will try it again.

chlewe:
I implemented a lazy post-order traversal. It wasn't clear to me that this approach has these upsides, so thanks for pointing that out. While the iterator is lazy, we have to collect all visited nodes to check for duplicates. There is probably no way around this. What I could remove is the complicated cross-reference checking logic, so that's good.

apoelstra (approved):
ACK a4d7997c2045e6305980427031c374b1b30abf99

This looks great! Thanks for doing so many iterations.

PR #26, opened 2022-06-16
Jet reform

chlewe:
This large PR reforms the way jets work in `rust-simplicity`:

1. Shared interface `JetNode` for both primitives and jets (that encode sub-DAGs); to avoid confusion, I will refer to the latter as _(jet) macros_. This resembles `jet_ptr` in C land, which also combines primitives and macros.
2. Terms refer to a jet via a static reference `&'static JetNode`
3. Replace the notion of _extensions_ with that of _applications_
4. A Haskell importer _(not part of this PR)_ can generate the files in `src/jet/init` to keep `rust-simplicity` up to date

# Duplicate jet macros

Because a `JetNode` has a name in context of its application (`JetNode.name: App::JetName where App: Application`), jet macros cannot be shared between applications. I chose to to this in order to exhaustively `match` the jet name during (de)serialization and execution, etc. Because Bitcoin and Elements use Core macros, there are duplicate `JetNode`s in `src/jet/init` that only differ in their Rust type. I think, this trade-off is worth it, since the folder is automatically generated anyway. What are your thoughts?

# Future extensions

Implement jet execution using FFI to C implementation.


apoelstra:
I find the name "jet macro" confusing. I think we should just stick with "jet".

I am also musing about whether we can get rid of this `Type`/`TypeName`, `Jet`/`JetName` distinctions ... I suspect not, because Rust doesn't have laziness so we need to simulate it .. but it's a little ugly.

apoelstra (approved):
ACK be91c0a052ed963099d0f78d3a6f770fdaa8be4a

I observe that the fuzztests don't compile for most of the commits, which I'll let slide because this is such a big PR. I haven't read all of the code in detail, but spent some time poking at the new APIs and looking at the resulting codebase under `git blame`.

Overall this seems like a big improvement. In followup PRs I'd like to

* rename "macros" to no longer conflict with an already-confusing part of the Rust language
* explore whether we can construct `Type`s directly without needing to use `TypeName`s
* replace `exec_jet` impls with ones that call into the C code

apoelstra:
No, sorry, I have a local script (which is too messy to share [though it's based on this thing](https://github.com/apoelstra/rsgit)) (though that one has threading issues so I largely just use a Python version of it) that just tries to build and `cargo test` every commit in every PR.

PR #30, opened 2022-07-14
Convenience

chlewe:
Adds some convenience methods to programs and DAGs, and updates library exports.

Adds support for fail nodes in decoded and constructed programs. We might want to add sanity checking in future versions.


src/core/term_dag.rs
```diff
@@ -164,6 +164,78 @@ impl<Witness, App: Application> TermDag<Witness, App> {
...
+    /// Create a DAG that takes any input and returns bit `1` as constant output.
+    pub fn bit_true() -> Rc<Self> {
+        TermDag::injr(TermDag::unit())
+    }
+
+    /// Create a DAG that takes a bit and an input,
+    /// such that the `left` child is evaluated on the input if the bit is `1` _(if branch)_
+    /// and the `right` child is evaluated on the input otherwise _(else branch)_.
+    ///
+    /// _The input and output types of both children must be the same._
```
apoelstra, line 201: In 32dcdfc863406a71fc46022f5adb43c242f7f973:

All the "must be"s in this commit should have corresponding assertions in the code (or maybe these methods should just return `Result`s).

apoelstra:
Done reviewing 5acddc4f95903b32e71d19c01fdf03a616074cc1

src/core/term_dag.rs
chlewe: Assertions would be helpful, but unfortunately `TermDag` contains no type information. We could convert the children into linear programs and do type checking every time we construct a larger DAG, but this seems like a bit much.

I added the "type hints" to the documentation to help resolve type checking issues. I will update the way I wrote it, but for now it will be limited to documentation only.

src/core/term_dag.rs
```diff
@@ -237,7 +237,7 @@ impl<Witness, App: Application> TermDag<Witness, App> {
     }
 
     /// Return the left child of the given DAG root, if there is such a child.
-    pub fn get_left(&self) -> Option<&Self> {
+    pub fn get_left(&self) -> Option<&Rc<Self>> {
```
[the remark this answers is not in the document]
chlewe: This is indeed curious: In a later commit (outside this PR), I wrote a method that returns the source and target type of a node. The return value of that is cloned, so it makes sense to return a smart pointer with cheap cloning. Similarly, if we ever wanted to clone the left or right child of a DAG, this updated method signature would be useful.

Arguably, cloning `&TermDag<_, _>` is as fast as cloning `&Rc<TermDag<_, _>`, so we can remove this commit.

src/core/term_dag.rs
apoelstra: Ah! I understand. Maybe you could add to the comment something like "if this is not upheld, type inference will fail, possibly in surprising and nonlocal ways".

apoelstra (dismissed):
ACK 093e3f8f94a4715d3940b34689b54daa799597c9 though I'd like to see the "must be" comments updated to explain the consequences of not upholding the conditions.

src/core/term_dag.rs
chlewe: Type checking is a huge source of frustration when writing programs from scratch. I think it would be easy to write a `TypedDag` that includes type information and can check its children as it is being constructed. We could convert that straight to `TypedProgram`, making use of the generic DAG traversal code. It's an idea for a next PR.

chlewe:
I added explanations to the documentation.

apoelstra (approved):
ACK db79c7ea76f49773f25ea8d0093cf6b1442c6f63

PR #31, opened 2022-07-18
Nested programs

chlewe:
This PR updates the hierarchy of program types:

1. `UntypedProgram` is for encoding / decoding and for conversions to / from DAGs.
2. `TypedProgram` holds all metadata for the time of commitment. This includes the CMR.
3. `Program` holds all metadata for the time of redemption, i.e. all metadata.

Node types are nested such that types with more metadata include those with less. This is useful for iterating over terms etc.

Programs can be enriched with more metadata by using a chain of methods, where each link is interchangeable. For instance, the bare program can be decoded from binary whereas the witness is provided as a vector.

Maximal sharing is now checked by default, but programs have to be explicitly compressed.


src/program.rs
```diff
@@ -90,7 +86,24 @@ impl<App: Application> fmt::Display for ProgramNode<App> {
...
+}
+
+impl<App: Application> ProgramNode<App> {
+    /// Return the underlying term of the node.
+    pub fn term(&self) -> &Term<Value, App> {
+        &self.typed.term
+    }
+
+    /// Return the source type of the node.
+    pub fn source_ty(&self) -> &Arc<FinalType> {
```
apoelstra: In 224e081c056fc608d00f7563faec2089320fe028:

It feels a bit unidiomatic to return `&Arc`. Probably better to just return `&FinalType` unless we have a good reason to expose internals.

src/core/typed.rs
```diff
@@ -30,6 +32,37 @@ pub struct TypedNode<Witness, App: Application> {
     pub source_ty: Arc<FinalType>,
     /// Target type of the node
     pub target_ty: Arc<FinalType>,
+    /// Index of the node inside the surrounding program
+    pub index: usize,
```
apoelstra, line 35: In 1587bf728038f6dcf28964d25d9e7be609fd07e7:

It doesn't make sense to me that the index would be in a `TypedNode`. You don't know the indices until after pruning. You say in the commit message that `TypedNode` is "for commitment time" but at commitment time the indices are not known.

The only reason this works is that the current codebase only typechecks programs after they've been linearized, which doesn't seem like a design choice we want to commit to. And the "indices" that are available in an unpruned program are meaningless and (as near as I can tell) never actually used.

src/program.rs
apoelstra: Ah, in c010975498dee1db8d1654ea7b2fd47bf78a653b you actually clone these arcs to use them as hash keys. One alternate might be to just extract the TMRs, but what you've done here is good.

src/program.rs
```diff
@@ -104,9 +102,87 @@ impl<App: Application> Program<App> {
...
+
+        for node in &self.nodes {
+            if let Term::Hidden(h) = node.term() {
+                if seen_hashes.contains(h) {
+                    return false;
+                } else {
+                    seen_hashes.insert(h);
+                }
+            } else {
+                let primary_key = (node.imr, node.source_ty().clone(), node.target_ty().clone());
```
apoelstra: In c010975498dee1db8d1654ea7b2fd47bf78a653b:

I'd prefer you use `Arc::clone` rather than `.clone` to make it explicit that you're just cloning a rc rather than a full structure (and this will also prevent future changes from making these `clone`s become expensive).

The codebase is not very consistent about doing this, but that's because I didn't know that you could call `Arc::clone` back when I first wrote it :)

src/program.rs
```diff
@@ -104,9 +102,87 @@ impl<App: Application> Program<App> {
...
+    }
+
+    /// Compress the program such that it has maximal sharing (see [`Self::has_maximal_sharing`]).
+    pub fn maximize_sharing(self) -> Self {
+        let mut shared_program = Vec::new();
+        let mut shared_idx = 0;
+
+        let mut primary_key_to_shared_idx = HashMap::new();
+        let mut hash_to_shared_idx = HashMap::new();
+        let mut unshared_to_shared_idx = HashMap::new();
```
apoelstra: In c010975498dee1db8d1654ea7b2fd47bf78a653b:

I really think you could avoid a lot of this messiness by not computing indices until the last minute (after sharing, after pruning).

apoelstra: In fact I wonder if indices should only exist during de/serialization?

During serialization you could maintain a hashmap of `(imp, types)` to indices and then you'd also get sharing for free...this seems like a bigger refactor though and probably shouldn't be part of this PR.

apoelstra (dismissed):
ACK 5cd2177455f390d94595d9a6fe476953a8101eda I think all my non-index-related comments are nits, and the index-related ones should be dealt with in a followup

src/program.rs
chlewe: The complexity of sharing comes from the relative indices inside `Term` that need to be updated. If what a node pointed to was removed in favour of an equivalent DAG earlier on, then the node needs to point to this earlier DAG.

src/program.rs
chlewe: This is the same problem as with [get_left() and get_right()](https://github.com/ElementsProject/rust-simplicity/pull/30#discussion_r921191530); both actually stem from the same original commit. I prefer your solution of using the TMR because it seems Simplicity-idiomatic. `Tmr` implements `Copy`, so this should be fast.

src/core/typed.rs
chlewe: We will need to come back to that once we implement pruning: Both sharing and pruning may change indices. However, pruning involves a pseudo run of the Bit Machine in order to determine used branches, and the Bit Machine uses `.index` during its execution. This means that we might need `.index` in `TypedNode` after all.

chlewe: Pruning may also change node types, which influences the IMR and static bounds, and it may lead to more sharing. We might want to prune before sharing, if the larger program is not too slow to do so.

chlewe:
I made use of the TMR during sharing, added a decode method for `TypedProgram<Value, App>` and made the decode methods call each other to prevent code duplication.

src/core/typed.rs
apoelstra: Interesting observation that the Bit Machine uses `.index`. Probably it could use `get_left` and `get_right` instead?

Also good observation that pruning may change sharing. This strengthens my hope that we can move sharing to the de/serialization layer and forget about indices in the "real" code.

apoelstra:
@chlewe the commits don't all compile now.

chlewe:
I cannot reproduce the error, using `git rebase 78f6236 -x` and `cargo` with all feature combinations plus the fuzz tests. Could you tell me which combination in particular fails? Maybe, I am missing something.

chlewe:
The issue arises because [honggfuzz depends on crate `arbitrary`](https://crates.io/crates/honggfuzz/0.5.54/dependencies), which recently made use of const generics. As far as I can see, there is nothing that we can do. I recommend that you execute the fuzzer with `cargo +1.56`, while our crate is built with `cargo +1.41.0`.

apoelstra:
`cargo test` fails on 550e0910d1534ae6111e7879b3c2a354f6766d33 with rustc 1.57.0.

chlewe:
I was able to fix the problem. `cargo build` worked on the commit, but `cargo test` failed. Strange, but I fixed my testing script accordingly and this should not happen again.

apoelstra (approved):
ACK 6e6479ff946f4915a8e2c4984a227a34563bbff9

PR #32, opened 2022-07-25
Iterators

chlewe:
This PR makes the encoding use iterators over terms and values (instead of slices) for more flexibility. Encoding methods were added for all program types, which use the type system to infer whether there is a witness. Finally, fuzz tests were added to check roundtrips of (program +) witness de/encoding, and bugs were fixed.

### DIFF
--- a/fuzz/fuzz_targets/rtt_dag.rs
+++ b/fuzz/fuzz_targets/rtt_dag.rs
@@ -15,14 +15,15 @@
 extern crate simplicity;
 
 use simplicity::bititer::BitIter;
-use simplicity::core::TermDag;
-use simplicity::decode;
+use simplicity::core::{TermDag, UntypedProgram};
 use simplicity::jet::application::Core;
 
 fn do_test(data: &[u8]) {
     let mut iter = BitIter::new(data.iter().cloned());
 
-    if let Ok(program) = decode::decode_program_no_witness::<_, Core>(&mut iter) {
+    if let Ok(program) = UntypedProgram::<_, Core>::decode(&mut iter) {
+        // println!("{:?}", program);
+
         let dag = TermDag::from_linear(&program);
         let program_from_dag = dag.to_linear();
         assert_eq!(program, program_from_dag);
--- a/src/core/iter.rs
+++ b/src/core/iter.rs
@@ -40,6 +40,7 @@ pub trait DagIterable: Sized {
 /// Iterates over a DAG in _post order_.
 /// That means, left children appear before right ones, and children appear before their parent.
 /// Shared nodes appear only once at their leftmost position.
+#[derive(Clone, Debug)]
 pub struct PostOrderIter<'a, D: DagIterable> {
     dag: &'a D,
     stack: Vec<D::Node>,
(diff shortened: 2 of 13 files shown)


src/encode.rs
```diff
@@ -133,60 +137,72 @@ pub fn encode_program_no_witness<W: io::Write, App: Application>(
...
         w.write_bit(false)?;
     } else {
         w.write_bit(true)?;
-
-        let mut payload = Vec::new();
-        let mut payload_w = BitWriter::from(&mut payload);
         let mut bit_len = 0;
+        let witness_clone = witness.clone();
 
         for value in witness {
```
apoelstra: In 74ff5ee08b4de0fa948b2bf4b73522fb0162cc28:

I think it'd be clearer to just do `for value in witness.clone() {' here rather than having a separate cloned variable that you keep in scope the whole time for use later.

fuzz/fuzz_targets/rtt_dag.rs
```diff
@@ -15,14 +15,15 @@
...
-use simplicity::decode;
+use simplicity::core::{TermDag, UntypedProgram};
 use simplicity::jet::application::Core;
 
 fn do_test(data: &[u8]) {
     let mut iter = BitIter::new(data.iter().cloned());
 
-    if let Ok(program) = decode::decode_program_no_witness::<_, Core>(&mut iter) {
+    if let Ok(program) = UntypedProgram::<_, Core>::decode(&mut iter) {
+        // println!("{:?}", program);
```
apoelstra, line 25: In d13f0e4aaa49162d46622b3c51b99198cfdc3530:

Not sure if you meant to commit this println (or the other one below). It's not a problem, just making you aware.

src/core/iter.rs
```diff
@@ -94,3 +96,20 @@ impl<'a, D: DagIterable> Iterator for PostOrderIter<'a, D> {
         }
     }
 }
+
+/// Convert the given iterator over [`Term`]s into an iterator over the contained `Witness` values.
```
apoelstra, line 100: In 17ace32625de33b048b864573ba2929fddb861bf:

The iterator is over `Witness`es not `Term`s.

apoelstra (approved):
ACK 1539206a718006fc66ad265e032e547b5a930108

src/core/iter.rs
chlewe: Doesn't the text say that an iterator over `Term` is converted into an iterator over `Witness`? Maybe, I should reword it.

fuzz/fuzz_targets/rtt_dag.rs
chlewe: This is how I debug: I uncomment a print statement of a decoded byte string that fails later down the line. Some fuzz tests already contained these comments, so I added the remaining ones.

src/core/iter.rs
apoelstra: Oh, yes, the text is correct, I just misread the function signature. Never mind.

fuzz/fuzz_targets/rtt_dag.rs
apoelstra: Sounds good!

chlewe:
I cloned the iterators like you suggested. Thanks for pointing this out.

PR #33, opened 2022-07-26
Policy

chlewe:
This PR brings the Simplicity policy up to speed: The compiler was updated and tested _(`pk` and `older` type-checking only; rest execution tests)_. To this end, a SIGHASH_ALL jet was added as well as support for x-only public keys. Finally, a Miniscript can now be directly converted into a Simplicity policy.

It might seem like this PR has a lot of edits, but this is due to the movement of code inside files or from file to file. The effective edits are much fewer.


apoelstra (approved):
ACK 8568effd01bb357d33a41e1f92f9d75a0117c0fe

apoelstra:
This is really cool!

PR #34, opened 2022-07-30
Types

chlewe:
This PR refactors the type system: Types are kept separate from type inference, identifiers were changed, and documentation was added. The big addition is that variables have `String`s as identifiers and variable types implement `Display` using that. As these types are variable, we cannot use the optimisation of storing formatted strings in the struct. Instead, we create a large string and replace occurrences of powers of two using relatively few allocations. This formatting can be used in a later PR to create helpful type errors.

### DIFF
--- a/src/analysis.rs
+++ b/src/analysis.rs
@@ -28,7 +28,7 @@ pub(crate) fn compute_extra_cells_bound<App: Application>(
             program[node.index - i].extra_cells_bound
         }
         Term::Comp(i, j) => {
-            program[node.index - i].target_ty().bit_width()
+            program[node.index - i].target_ty().bit_width
                 + cmp::max(
                     program[node.index - i].extra_cells_bound,
                     program[node.index - j].extra_cells_bound,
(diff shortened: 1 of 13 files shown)


chlewe:
@apoelstra I managed to get a SIGSEGV for fuzz target `rtt_program_witness`, which does type checking. (All our unit tests run fine.) It must be the `RefCell`s. I suspect that my rewriting of `find_root()` or my compression of `finalize()` included an error that led to this. Could you take a look?

apoelstra:
Yes, it is a stack overflow. You either need to put depth limits on things or (better) get rid of all the recursive algorithms.

apoelstra (approved):
ACK f5d720193211e619bcd7413a9aca010fde8e06d7

apoelstra:
I'm going to go ahead and ACK this (once my local scripts finish). In general we need to deal with recursion throughout the library. For things like stringification we ought to be caching this data (and perhaps computing it in constructors, which would let us avoid search-and-replacing complete strings to compactify type names).

PR #36, opened 2022-08-30
Sharing

chlewe:
This PR adapts the encoding of programs to the new heap structures from #35. Importantly, programs without witness data can now be encoded and compressed via sharing. This works by populating each witness node with a `Unit` value during encoding, and requires that witness nodes are made fresh during decoding _(i.e., each witness node has at most one parent)_. Fuzz and unit tests that were disabled in #35 are adapted and re-enabled. An implementation of `Display` is added to programs for easier debugging.


chlewe:
One open question is the handling of errors during `CommitNode::encode()`: The program might not type-check. but the method returns `io::Result`, so we cannot pass `crate::Error`. Once we have local type checking, only the occurs check can fail at that point, which is unlikely. Still, we might not want to panic on some programs.

src/encode.rs
```diff
@@ -111,29 +116,115 @@ impl<W: io::Write> BitWriter<W> {
...
 }
 
-/*
-/// Encode an untyped Simplicity program to bits.
+/// Encode a Simplicity program to bits, without witness data.
 ///
 /// Returns the number of written bits.
-pub fn encode_program_no_witness<'a, W: io::Write, Witness, App: Application, I>(
-    program: I,
+pub fn encode_program_no_witness<W: io::Write, Witness, App: Application>(
```
apoelstra: In 0d6eb34178b483026316d1af8085f3bf90095230:

I think it'd be simpler and more efficient to inline `compute_maximal_sharing`. Then you'd only need to iterate over the program once: for each node, if it exists in `node_to_index` map then you can skip over it; otherwise serialize it in the next available position in the vector. (You are guaranteed that its children will already exist in `node_to_index` so they can be referenced.)

In bbad34e554b0950ea05d6c69a8567bc7447822b2 I think there's a similar argument for inlnining `check_maximal_sharing` into the decoding logic, though it's less strong.

src/core/commit.rs
```diff
@@ -443,8 +445,29 @@ impl<Witness, App: Application> CommitNode<Witness, App> {
...
+    /// # Usage
+    ///
+    /// Use this method only if the serialization **does not** include the witness data.
+    /// This means, the program simply has no witness during commitment,
+    /// or the witness is provided by other means.
+    ///
+    /// If the serialization contains the witness data, then use [`Node::decode()`].
     pub fn decode<I: Iterator<Item = u8>>(bits: &mut BitIter<I>) -> Result<Rc<Self>, Error> {
-        decode::decode_program_no_witness(bits)
+        decode::decode_program_no_witness(bits, true)
```
apoelstra, line 481: In 6076bdbfe1e6624aa1c09ed9021950fe1952694c:

This boolean argument shouldn't be exposed outside the `decode` module. It's fine to have one function so you can reuse a bunch of the code but there should be two separate functions exposed from `decode::`.

src/decode.rs
```diff
@@ -150,19 +182,42 @@ fn decode_node<I: Iterator<Item = u8>, App: Application>(
...
+/// then a witness node is inserted at the parent index,
+/// the parent index is incremented by one
+/// and the fresh witness is returned as child.
 fn get_child_from_index<App: Application>(
     index: usize,
-    index_to_node: &HashMap<usize, Rc<CommitNode<(), App>>>,
+    parent_index: &mut usize,
+    index_to_node: &mut HashMap<usize, Rc<CommitNode<(), App>>>,
+    maybe_used_witness_indices: &mut Option<HashSet<usize>>,
 ) -> Rc<CommitNode<(), App>> {
```
apoelstra, line 199: In 6076bdbfe1e6624aa1c09ed9021950fe1952694c:

Is this function needed, or could you just detect witness nodes and always return a fresh `Witness` node without bothering to keep track of which indices had been used?

I think for `CommitNode`s we just want to blindly construct fresh `Witness` nodes (which should be basically free) and for `RedeemNode`s we want to enforce the ordinary sharing logic.

I also still think this logic would be easier to follow if sharing-related stuff were inlined into decoding.

src/decode.rs
```diff
@@ -297,7 +300,7 @@ pub fn decode_value<I: Iterator<Item = bool>>(ty: &Type, iter: &mut I) -> Result
 }
 
 /// Decode a 256-bit hash from bits.
-fn decode_hash<I: Iterator<Item = u8>>(iter: &mut BitIter<I>) -> Result<[u8; 32], Error> {
+pub fn decode_hash<I: Iterator<Item = u8>>(iter: &mut BitIter<I>) -> Result<[u8; 32], Error> {
```
apoelstra, line 303: In c26c76b26623cab3f0d645be451331223a046a11:

It just occurred to me now that we could add a `fuzz` module which was gated on `#[cfg(fuzzing)]` which would expose this stuff. And then "only exposed for fuzzing" APIs could actually be exposed only for fuzzing.

apoelstra (approved):
ACK 937190f57047159346b708327b37f46b3a081ea6

apoelstra:
Some of my comments are things that I'd like to get fixed, but I'll leave it up to you whether you want to edit this PR or just do a followup one.

src/decode.rs
chlewe: This is a really cool idea. I want to keep the API as small as possible, but fuzzers still need access.

chlewe:
Any opinions on `CommitNode::encode()` (see above)?

apoelstra:
@chlewe do we care about typecheckability at encoding time? If a user manages to create a bad program why not just let them serialize a bad program?

chlewe:
Types are important to collapse (share) witness nodes (of the same type). One way out would be to assign dummy types if type checking fails, but this seems weird. With local type checking, users will be basically unable to create these problematic programs, so this problem will resolve itself.

apoelstra:
Yeah, I think with local type-checking we should be able to make it literally impossible to produce programs where typechecking fails (since the node struct itself has a non-optional cached type). I'm sure users could find a way to abuse the API to get broken programs but they'd have to at least put *something* in there.

If it turns out this is impossible and encoding might fail, then it should just return a `Result` and we should wrap the IO errors in a richer type.

src/encode.rs
chlewe: We must iterate at least two times, because the length of the program is unknown until sharing is complete. This length is the first part of the serialisation, so we must write it before writing any nodes. I do not know of any way to defer part of the serialisation without saving everything else in an inefficient vector. We have the same problem when encoding the witness, where we also iterate twice.

chlewe: Also, the `Node` references that are returned by the post-order iterator are unique. Cloned `Rc`s are already filtered out by the iterator.

src/decode.rs
chlewe: Bookkeeping is needed because nodes reference their children via indices (as in the serialisation), but at the same time the program grows and the indices (as in the deserialised program) change. At the end, we compare the order of nodes as we deserialised them with the post-order of the program's iterator. In my original code, I tried to solve both problems with the same index map, which made it complicated. To make things clearer, I rewrote the code to keep track of "indices" (as in the serialisation) and of "new indices" (as in the deserialised program).

chlewe:
I made the `decode` and `encode` modules private by default unless during fuzzing. `BitWriter` moved into its own module. The decoding logic was simplified, but the sharing check cannot be inlined because it does not apply to `CommitNode`. _(We can check sharing for all programs once we have local types.)_

src/encode.rs
apoelstra: Ok, this sounds reasonable.

apoelstra (approved):
ACK acea6e8fe01b2433f8a9f80075684e152821ae58

PR #37, opened 2022-08-31
Fix fuzzing

chlewe:
Drops honggfuzz dependency `arbitrary` to meet MSRV. `arbitrary` 1.1.3 introduced `array_from_fn`, which is stable from Rust 1.63.0.

See [https://github.com/lightningdevkit/rust-lightning/pull/1687](https://github.com/lightningdevkit/rust-lightning/pull/1687).


apoelstra (approved):
ACK 7372460a7642197b587e6e7c05b4248feda6ae0f

PR #38, opened 2022-09-06
Policy hotfix

chlewe:
This PR enables the (de)serialisation of all compiled Simplicity policies. SIGHASH_ALL replaces INPUTS_HASH, because the addition of the former required a corresponding serialisation. In the future, when the Rust jets are brought up to date, all jets will have some serialization, but the concrete bitstrings will change.


apoelstra (approved):
ACK 6d7be58cc68b5bccf417d512ab73cbb1b4e35dbb

PR #40, opened 2022-09-12
Fix canonical order

chlewe:
The indices strike back: Inserting fresh witnesses breaks our algorithm for checking canonical order. We must check canonical order of the program as it is serialised (without fresh witnesses). This PR adds a light-weight DAG structure to implement this.

We failed to recognise this error because the de/encoding of `CommitNode` is not thoroughly tested. This will change once we have local type checking to check sharing of these programs and write a fuzz test.

### DIFF
--- a/src/core/iter.rs
+++ b/src/core/iter.rs
@@ -216,3 +216,42 @@ impl<I: Iterator<Item = Value>> WitnessIterator for I {
         Ok(())
     }
 }
+
+/// Sequence of nodes with (absolute) left or right indices in post-order that form a DAG.
+pub(crate) type IndexDag = Vec<(Option<usize>, Option<usize>)>;
+
+/// Wrapper for (absolute) indices into [`IndexDag`] that implements [`DagIterable`].
+#[derive(Copy, Clone)]
+pub(crate) struct IndexNode<'a>(pub usize, pub &'a IndexDag);
+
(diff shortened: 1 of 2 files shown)


apoelstra (dismissed):
ACK e9fd5a23c27556684118dd9c53d00e3ef21bffae

apoelstra:
concept ACK, but I can't help the feeling that we've overcomplicated the decoding procerss. Maybe we want a new dag structure which is only used during decoding, in which nodes are labelled with their indices and witnesses are non-expanded. Then we can quickly do the ordering check (by copying the algorithm in `verifyCanonicalOrder` in the C code) before converting to a `CommitNode` (and doing the witness expansion during that step).

Anyway this is good for now.

chlewe:
There were two paths: Either check canonical order of the program as it is serialised (as opposed to how it is deserialised), or update the bookkeeping so that the deserialised program is in canonical order if the serialised one is. I went with the second option, which turned out complicated. My new approach uses the first option, which is shorter and more future-proof _(in case do more changes than just inserting fresh witnesses)_.

apoelstra:
Amazing! This looks 100% better.

apoelstra (approved):
ACK 9d642e0284539e864a72727870f2334e3fbd5e7d

PR #41, opened 2022-09-12
Optimise policy

chlewe:
This PR slightly reduces the size of compiled policies with trivial clauses in conjunctions (Miniscript fragment `t:X = and(X, 1)`). It also fixes the execution and printing of hidden and fail nodes, and adds `assert`.

### DIFF
--- a/src/bit_machine/exec.rs
+++ b/src/bit_machine/exec.rs
@@ -404,13 +404,7 @@ impl BitMachine {
                     }
                 }
                 RedeemNodeInner::Witness(value) => self.write_value(value),
-                RedeemNodeInner::Hidden(h) => {
-                    // TODO: Convert into crate::Error
-                    panic!(
-                        "Hit hidden node {:?} at iteration {}: {}",
-                        ip, iterations, h
-                    )
-                }
+                RedeemNodeInner::Hidden(..) => return Err(ExecutionError::ReachedPrunedBranch),
                 RedeemNodeInner::Jet(j) => App::exec_jet(j, self, env)
                     .map_err(|x| ExecutionError::AppError(Box::new(x)))?,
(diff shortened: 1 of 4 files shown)


apoelstra:
Looks like you handle `t` correctly -- by basically just ignoring the `t:` wrapper -- but why do you bother turning `l:` and `u:` into anything complicated? They have no semantic effect, just throw them away.

chlewe:
The handling of `l:` and `u:` is about avoiding `fail` nodes in the program. Come to think of it, I manually pruned the program. Let's remove these two optimisations and implement pruning instead.

chlewe:
The fail and hidden nodes look huge in the Graphviz program graph. This is what gave me the idea for this PR. Would it be an option to not print the hashes in `fmt::Display`?

apoelstra:
There is no need for `fail` or `hidden` or anything for `l:` or `u:`. Just ignore these wrappers entirely.

chlewe:
We are talking about slightly different things, but what you say is clear: Ignore identity wrappers and convert what is inside. This is actually what the code on `master` does already. The wrappers `l:`, `u:` and `t:` are syntactic sugar that doesn't exist in the Miniscript structures. What we actually have are `and` and `or` with `true` or `false` inside, for which I added special handling. The handling of `or+false` is basically manual pruning, so I removed it. The handling of `and+true` skips the `and` and converts what is inside.

apoelstra:
I am aware that `l:`, `u:` and `t:` are just syntactic sugar. It looks like the current version of this PR doesn't deal with `l` or `u` at all, which is fine, but it would be good to eliminate those too at some point.

apoelstra (approved):
ACK 4b4fba5ae2569477b92ad3501e3551e7e1ccd436

PR #42, opened 2022-09-21
Local types

chlewe:
This PR makes type checking local, i.e., types are checked as programs are constructed. Each method returns a `Result` type that needs to be handled by the user. If there is an error, then it is shown immediately when and where it is made. A `Context` keeps the state for constructing a program and is passed as argument to each construction call.

### DIFF
--- a/src/core/context.rs
+++ b/src/core/context.rs
@@ -1,0 +1,26 @@
+use crate::core::types::VariableFactory;
+use crate::jet::Application;
+use std::marker::PhantomData;
+
+/// Context for constructing a Simplicity program
+#[allow(dead_code)]
+pub struct Context<App: Application> {
+    pub(crate) naming: VariableFactory,
+    _application: PhantomData<App>,
+}
+
+impl<App: Application> Context<App> {
+    /// Create a new context.
+    pub fn new() -> Self {
+        Self {
+            naming: VariableFactory::new(),
+            _application: PhantomData,
+        }
+    }
+}
+
+impl<App: Application> Default for Context<App> {
+    fn default() -> Self {
+        Self::new()
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -63,6 +63,10 @@ pub enum Error {
     NonCaseHiddenChild,
     /// 'case' nodes may have at most one hidden child
     CaseMultipleHiddenChildren,
+    /// Right child of left assertion must be hidden
+    RightChildNotHidden,
+    /// Left child of right assertion must be hidden
+    LeftChildNotHidden,
     /// Bitstream ended early   
     EndOfStream,
     /// Program must not be empty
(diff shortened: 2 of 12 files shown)


chlewe:
Maybe, we want to change the name of `NodeHelper` to something else.

chlewe:
I also realised that the occurs check can fail in [pretty minimal scenarios](https://github.com/chlewe/rust-simplicity/blob/local-types/src/core/allocator.rs#L675).

apoelstra:
Where is `NodeHelper` actually used? It seems the last commit adds the type but doesn't use it.

apoelstra:
Done reviewing 15ee9847a744629654a4e834590747a69a05d55d. Overall looks good. As discussed out of band:

* `NodeHelper` seems unnecessary, and if it is, it does not fit into this PR
* `NodeAllocator` should probably be renamed `Context` and have the ability for users to swap out the algorithm for allocating variable names, etc. But this can wait for a future PR since it's mostly a cosmetic change. It should also be passed as an argument to e.g. `CommitNode::unit` etc rather than having `unit` be a method on it.

Otherwise this looks great!

apoelstra (approved):
ACK 15ee9847a744629654a4e834590747a69a05d55d

chlewe:
I renamed the `NodeAllocator` to `Context`, kept all DAG creation methods in `CommitNode`, and made the methods return `Result` types that produce user-friendly errors upon unwrapping. `NodeHelper` is completely removed.

src/core/context.rs
```diff
@@ -0,0 +1,26 @@
+use crate::core::types::VariableFactory;
+use crate::jet::Application;
+use std::marker::PhantomData;
+
+/// Context for constructing a Simplicity program
+#[allow(dead_code)]
+pub struct Context<App: Application> {
+    pub(crate) naming: VariableFactory,
```
apoelstra, line 8: In bdfd1769a4f02a3ba4377868c70d0248beb699a7:

I know this is only a `pub(crate)` but ideally it should be private and there should just be an accessor method for allocating a new variable.

In future I'd like to replace this with a boxed closure actually and support users replacing it with something different; so we don't want to expose too much of the implementation in the API.

src/lib.rs
```diff
@@ -46,15 +46,22 @@ mod util;
...
-    TypeCheck(&'static str),
+    Unification(&'static str),
     /// A type is recursive (i.e., occurs within itself), violating the "occurs check"
     OccursCheck,
+    /// A DAG cannot be created because the children of the root cannot be unified
+    TypeCheck(
+        [&'static str; 2],
+        Option<UnificationArrow>,
+        Option<UnificationArrow>,
+    ),
```
apoelstra, line 63: In 5f4ba25c37dc8e04c43c0b2fb3d412c3922511e1:

These should all be named fields, and the two strings should not be bundled into one array.

apoelstra (approved):
ACK 8585d71713090c0f0506e62021e66d952d5d1663

chlewe:
I added field names to `Error::TypeError`. Thanks for the suggestion.

apoelstra (approved):
ACK 53dcc907f1bcecc974a17d0de1dff850c2e80014

PR #44, opened 2022-10-21
Fully automatic jets

uncomputable:
Jets are fully defined by their init files so that minimal changes have to be made to the remaining Rust code when these files are changed.

The number of edits is due to mechanical changes.

# Notes
- Added Type names for `char` and `short`
- `Application` no longer made sense because all its logic went into `init`, so the `Jet` trait replaces it
-  Jets return Boolean value (pass / fail) instead of custom errors. Program evaluation must be independent of evaluation order and does not allow for complex errors

# Future
Replace function pointers to Rust code with function pointers to C code via FFI.


src/jet/mod.rs
```diff
@@ -56,14 +57,30 @@ pub trait Application: Clone + 'static {
...
+        let mut code = 0u64;
+
+        for len in 1..=64 {
+            let bit = iter.read_bits_be(1).ok_or(Error::EndOfStream)?;
+            code <<= 1;
+            code += bit;
+
+            if let Some(jet) = Self::Jet::decode(code, len) {
+                return Ok(jet);
+            }
```
apoelstra, line 66: In a98a69d931fddae67da8cf8d588efeb8d713c9db:

I'm really skeptical of this algorithm. It's a clever way to generically decode from a prefix code without having a bunch of complicated branches, but in the case that we encounter a bad jet, it will try to just read the entire remainder of the program searching for some valid prefix.

In fact it won't be able to do this because the `code <<= 1` line above will overflow and cause a panic, but that's even worse :).

I get that the goal is to avoid having the Haskell code produce a prefix-code-decoder algorithm since that'd be very branch heavy and difficult to do without knowing Haskell. But I think we can avoid this by having some sort of `PrefixCode` structure which internally holds a search tree, and having the haskell-generated code just add codewords to this with simple one-liners like `code.add_jet(u64, usize, Jet { ... })` or something.

Hopefully we can produce such a tree at compile-time, but if not, we could hold it in the Context object... which may even give us some form of runtime switching between different Simplicity extensions.

I also think you should pull `(u64, usize)` into a `PrefixCodeWord` type because it's difficult to read code that's passing around bare tuples like this.

uncomputable: The decoding code was at best a bad compromise. Part of this PR is finding ways to encode behaviour in static files. Let's find an encoding that let's us decode jets without having to brute force.

uncomputable: I don't see why `code <<= 1` should overflow. It doesn't on my end if I disable the inner if statement.

apoelstra: @uncomputable are you running in debug mode? If you try to read more than 64 bits into a `u64` it will overflow.

uncomputable:
I rebased and added `PrefixCodeWord` as well as a maximum encoding length.

If computers were bit-based, I would implement a trie of bit strings, but they are not. As far as I can see, `u64` operations are as fast as `bool` operations. Fewer branches are also more readable. We might be able to get slightly more performance, but would the effort be worth that? We do almost no operations as is. Do we have a policy towards non-64-bit architectures?

src/jet/mod.rs
apoelstra: Just closing out this discussion -- I talked to @uncomputable on the phone who politely explained that left-shifting by a single bit is not going to overflow no matter how many times in a row you do it :)

uncomputable:
As discussed off band, I reconstructed this PR from scratch for a less messy history. Decoding now uses a search tree that is concisely defined by a Rust macro. See updated PR description.

src/bit_machine/exec.rs
```diff
@@ -410,8 +410,9 @@ impl BitMachine {
                 }
                 RedeemNodeInner::Witness(value) => self.write_value(value),
                 RedeemNodeInner::Hidden(..) => return Err(ExecutionError::ReachedPrunedBranch),
-                RedeemNodeInner::Jet(j) => App::exec_jet(j, self, env)
-                    .map_err(|x| ExecutionError::AppError(Box::new(x)))?,
+                RedeemNodeInner::Jet(j) => {
+                    App::exec_jet(j, self, env).map_err(ExecutionError::from)?
```
apoelstra: In 6ec8e71cee1906a73fb3dc5c2601e15cd6dc256c:

This call to `from` is unnecessary. `?` will call it for you.

src/bit_machine/exec.rs
```diff
@@ -446,29 +447,29 @@ impl BitMachine {
...
-pub enum ExecutionError<'a> {
+pub enum ExecutionError {
     /// Reached a fail node
     ReachedFailNode,
     /// Reached a pruned branch
     ReachedPrunedBranch,
-    /// Application-related error
-    AppError(Box<dyn AppError + 'a>),
+    /// Jet failed during execution
+    JetFailed,
```
apoelstra, line 454: In 6ec8e71cee1906a73fb3dc5c2601e15cd6dc256c:

Why don't we wrap the `JetFailed` struct here?

src/analysis.rs
```diff
@@ -27,7 +27,7 @@ pub(crate) const IO_EXTRA_FRAMES: usize = 2;
 /// Nodes with left children require their finalized left child,
 /// while nodes with right children require their finalized right child.
 /// Witness nodes require their node type.
-pub(crate) fn compute_bounds<App: Application>(
+pub(crate) fn compute_bounds<App: Jet>(
```
apoelstra, line 30: In d689e20c5cfb81f6a06a15eaef3ea32aa62efe3a:

We continue to have `App: Jet` in many places. I think it may be clearer to have `J: Jet` instead. First because it gets rid of the confusing name "App" and secondly because there's a convention that (usually) generic types have 1- or 2-character names.

apoelstra:
Done reviewing 87ce697edebe8c66ec62b8eaa51fcd96d70cec61. This looks great! It's clear(ish) how this code can be autogenerated from Haskell, and it's much cleaner than before.

src/bit_machine/exec.rs
uncomputable: I wasn't sure if wrapping an empty struct makes sense, but we can do it.

src/analysis.rs
uncomputable: Sure, I will push another commit. It's funny how we return to `Jet` which existed in this repository before I killed it.

src/bit_machine/exec.rs
apoelstra: It's zero-cost, avoids some slightly ugly code, and it'll work if the struct ever stops being empty

uncomputable: Sounds reasonable. It's in the latest commits.

apoelstra (approved):
ACK c1189805e4c562cffc0cb255cecc44a3ede05f8e

Issue #43, opened 2022-10-13
Notes from Tabconf


sanket1729:
- We should have a finalized `FinalizedCommitNode` type
- finalize() API that checks CommitNode is () -> () without any Witness
- `RedeemNode` API should accept `FinalizedCommitNode` to `Witness`. 

- `Value` should save the `Type` information
- Debug uncommented assertions in the code
- Add a comment explaining why we `Clone` in encoding `Witness`	

- Given two finalized nodes, there are two ways to compose them:
	 - _Un_finalize both of them and refinalize them again. This is the trivially correct way to do it but it may change the types of witness nodes.
	 - "Unify" the nodes. This _should?_ work if there are no cross-references. Either it fits or does not fit under unification constraints. 
	 - Can we say under what conditions the second approach is the correct way to do things? Should we refuse to unify if the second approach is not the same as the first?

apoelstra:
cc @roconnor-blockstream the last bullet point in particular we had some doubts about and would've liked to have had you in the room. (We can also have a phone call to try to elaborate if it's not clear from this issue what we're thinking.)

uncomputable:
I think we should split this into separate issues and those this one.

uncomputable:
Moved to #105 #106 #107 #108

PR #45, opened 2022-10-27
Fix bounds

uncomputable:
Asserts that cell and frame bounds are kept, clarifies what these bounds mean, and fixes a bug in how many frames are allocated.

### DIFF
--- a/src/analysis.rs
+++ b/src/analysis.rs
@@ -32,7 +32,7 @@ pub(crate) fn compute_bounds<App: Application>(
 ) -> NodeBounds {
     NodeBounds {
         extra_cells: compute_extra_cells_bound(untyped_node, left.clone(), right.clone(), ty),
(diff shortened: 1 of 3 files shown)


uncomputable:
Addresses point 5 of #43 .

sanket1729 (approved):
ACK cb6942d83b682e5f769a266c3f1ffea2fe58a1de. I guess the +2 is for +1 each for input and output?

uncomputable:
Exactly. I added a constant to make the code more readable.

sanket1729 (approved):
reACK a72caa970ec96215e6c3160300a7b3a5b220ab07. A comment also would have been fine :)

uncomputable:
The Rust compiler should be able to resolve these constants. There is probably a design guide somewhere that says that magic numbers should be constants.

PR #46, opened 2022-11-24
Turing fixes

uncomputable:
These are fixes from the [turing branch](https://github.com/uncomputable/rust-simplicity/commits/turing) that [simple-turing](https://github.com/uncomputable/simple-turing) builds upon.

Making some internals public makes it easier for other crates to use these types. Returning hashes of pruned branches or fail nodes makes debugging much easier. The type inference is fixed and the Miniscript policy compiler is slightly optimised.


apoelstra (approved):
ACK 907ac3b6c41edebbbf9d7211664a8b7656c8e072

PR #48, opened 2022-12-14
Add basic vendoring scripts

sanket1729:
Copied and updated basic vendor scripts from the rust-secp256k1 repos. This does not no any symbol renaming, but is still handy and maintains which versions of simplicity we are using.

PR #49, opened 2022-12-14
PR 2 of 3: Add bindings in simplicity-sys crate.  

sanket1729:
Based on #48


sanket1729:
Rebased on top of merged #48 . Add a commit that fixes the filenames

uncomputable:
Not sure if something went wrong during rebasing, but the first commits 81dfee2 ... 9f55381 have the same hashes as existing commits on `master`.

sanket1729:
@uncomputable, those were merged as a part of the first PR. This PR now only consists of three commits.

simplicity-sys/depend/wrapper.c
```diff
@@ -0,0 +1,38 @@
...
+{
+    *frame = initWriteFrame(n, from);
+}
+
+bool c_readBit(frameItem *frame)
+{
+    return readBit(frame);
+}
+
+void c_writeBit(frameItem *frame, bool bit)
```
apoelstra, line 23: In 7cea6a1dcd9432114558a55758fc985b3d45413c:

I could go either way on this -- the only difference between the wrapped function and the unwrapped one is the presence of `static`. Arguably we should change upstream to expose these, or use a patch-applying trick like we do in rust-secp. (But on the other hand, applying patches is ugly and sometimes breaks, so..).

At the very least we should add a comment explaining that we're doing this since the internal function is `static`. On the others we should mention that we're also doing it to avoid passing C structures by value, since we're unsure about the FFI story there.

simplicity-sys/src/c_jets/c_env.rs
```diff
@@ -0,0 +1,233 @@
...
+///     4. We only create allocations for data fields that are stored differently from the
+/// consensus serialization format.
+#[derive(Debug)]
+pub struct CRawOutputData {
+    pub asset: Vec<c_uchar>,
+    pub value: Vec<c_uchar>,
+    pub nonce: Vec<c_uchar>,
+    pub surjection_proof: Vec<c_uchar>,
+    pub range_proof: Vec<c_uchar>,
+}
```
apoelstra, line 23: In 59c2942084530ec7c2d06b9147dccd3156aa51c3:

This structure (and others) shouldn't have the `C` prefix since it's not actually a FFI type. I think we should only use `C` for things with `repr(C)` on them.

simplicity-sys/src/c_jets/frame_ffi.rs
```diff
@@ -0,0 +1,27 @@
...
+    pub edge: *const c_uchar,
+    pub len: size_t,
+}
+
+extern "C" {
+    pub static c_sizeof_frameItem: size_t;
+    pub static c_sizeof_UWORD: size_t;
+
+    pub(crate) fn c_initWriteFrame(frame: &mut CFrameItem, n: size_t, from: *const c_uchar);
+    pub(crate) fn c_initReadFrame(frame: &mut CFrameItem, n: size_t, from: *const c_uchar);
```
apoelstra, line 21: In 59c2942084530ec7c2d06b9147dccd3156aa51c3:

I'm pretty suspicious about using `c_uchar` instead of `u16` here. The alignment requirements are different.

simplicity-sys/src/c_jets/c_frame.rs
```diff
@@ -0,0 +1,54 @@
...
+            len: 0,
+        }
+    }
+
+    /// Initialize a new read frame.
+    /// 'n' is the number of cells for the read frame.
+    /// 'from' is a pointer to the beginning of the new slice for the array of u8 to hold the frame's cells.
+    ///
+    /// Note: The C implementation uses array of UWORD for `from`. We use a slice of u8 instead.
+    pub fn new_read(n: usize, from: *const c_uchar) -> Self {
```
apoelstra, line 23: In 59c2942084530ec7c2d06b9147dccd3156aa51c3:

This is definitley not a safe function if we take a `*const c_uchar` because it is possible to make such a pointer which is unaligned when reinterpreted as a `*UWORD`.

It is also definitely not safe because it calls into `initReadFrame` which has as a precondition that `from` points to an array of `n` `UWORD`s, while this API allows it to be null or dangling or to point to uninitialized memory.

simplicity-sys/src/c_jets/c_frame.rs
```diff
@@ -0,0 +1,54 @@
...
+            let mut frame = CFrameItem::new_unchecked();
+            frame_ffi::c_initReadFrame(&mut frame, n, from);
+            frame
+        }
+    }
+
+    /// Initialize a new write frame.
+    /// 'n' is the number of cells for the write frame.
+    /// 'from' is a pointer to the one-past-the-end of the new slice for the array of UWORDS to hold the frame's cells.
+    pub fn new_write(n: usize, from: *mut u8) -> Self {
```
apoelstra, line 35: Ditto here. This isn't a safe API because if you violate the preconditions you will get UB.

simplicity-sys/depend/wrapper.c
sanket1729: The inlining was also an issue here. Inline functions don't have symbols to refer to.

simplicity-sys/src/c_jets/frame_ffi.rs
sanket1729: `c_uchar` instead of `u8` you mean?

sanket1729: The caller has to take care about calling with the correct alignment of `c_uchar` which we do in `c_frame.rs`. Refer to the functions `round_u_word` and `ffi_bytes_size`

simplicity-sys/depend/wrapper.c
apoelstra: Makes sense. I think we should have a block comment in this file explaining this.

simplicity-sys/src/c_jets/frame_ffi.rs
apoelstra: No, `c_uchar` and `u8` are the same, that would be fine. The issue is that the C code is expecting a pointer to u16 while we're giving it a pointer to u8.

It looks like `round_u_word` is only ever called from `ffi_bytes_size`, which is never called.

simplicity-sys/depend/wrapper.c
apoelstra: The new commit looks fine, though you wrote `weadBit` rather than `writeBit`.

simplicity-sys/src/c_jets/frame_ffi.rs
sanket1729: See example usage: https://github.com/sanket1729/rust-simplicity/blob/911b6072f90341936479948c08e22c7cd20aa70f/src/jet/mod.rs#L126-L132

Is alignment a concern here? We are passing a pointer to `u8 * sizeof(UWORD)` instead of `UWORD`?

sanket1729: Perhaps, all this should be a reference to `usize`? I think on all systems, but 8 bit systems, `uint_fast_16` should map to usize?

apoelstra: >These functions are not called in this PR. We are not using jets here.

If you are required to call a separate alignment function to avoid UB then these functions can't be marked `safe`.

>Perhaps, all this should be a reference to usize? I think on all systems, but 8 bit systems, uint_fast_16 should map to usize?

I think this would be okay, though we should have tests that assert this.

simplicity-sys/depend/env.c
```diff
@@ -0,0 +1,68 @@
+#include <stdlib.h>
+#include "simplicity/elements/env.h"
+#include "simplicity/primitive/elements/primitive.h"
+
+const size_t c_sizeof_rawBuffer = sizeof(rawBuffer);
+const size_t c_sizeof_rawOutput = sizeof(rawOutput);
+const size_t c_sizeof_rawInput = sizeof(rawInput);
+const size_t c_sizeof_rawTransaction = sizeof(rawTransaction);
+const size_t c_sizeof_rawTapEnv = sizeof(rawTapEnv);
+const size_t c_sizeof_txEnv = sizeof(txEnv);
```
apoelstra, line 10: Since we are using C11 we can also use `alignof` and assert the alignment.

sanket1729: Done

sanket1729:
Force pushed with
1) Using assertions to check the alignment is working as intended. 
2) Marking `new_read` and `new_write` as unsafe
3) Using a pointer to `usize` instead of `u8`. C code has a pointer to `UWORD`, but that should map to usize in on all 32/64 bit machines. This is also checked in tests and will be checked before running any jet

sanket1729:
Also pushed #50 that shows how this code is used. Refer to the file here: https://github.com/ElementsProject/rust-simplicity/blob/8eb9d1af989b709bd173a70fb92d5e547a89d5f4/src/jet/mod.rs#L132-L171

apoelstra:
This looks great now!

apoelstra (approved):
ACK 987ade1e243c10af510c4aa604879cbc871da996

PR #50, opened 2022-12-14
PR 3 of 3: Use bindings created for jet execution.

sanket1729:
Based on #48 and #49


sanket1729:
Rebased after #49

apoelstra:
We never actually re-enable the `policy` module -- is that intentional?

apoelstra:
Ah, yes, I see, there's a TODO for doing Bitcoin jets that still needs to be addressed before we can re-enable the module.

apoelstra:
Overall this looks great. There are a bunch of `unimplemented!`s and `FIXME`s and `TODO`s left which we will need to address in the new year. I'm not sure what the best strategy for testing this all is -- because we call into C code we can't use Miri. I suppose fuzzing+valgrind may be our best bet.

At some point we can use mutagen to assess our test coverage.

apoelstra (approved):
ACK bbbb7caa3a09143a97e76ac1b04ee47d438280a9

sanket1729:
@apoelstra, yep. The policy compiler code is really old and based on bitcoin jets that I assumed would be created in 2020. Time to revisit this in 2023. 

Similar to how we exported the jets, we should also export the test cases from `schnorr6`, `schnorr0`, `hashblock`, `sighashall`, and others to make sure that the programs are also executed correctly by rust-simplicity.

PR #55, opened 2023-01-17
Add AMR

### LINKED ISSUE #54: Should we re-add `amr` to test that the type inference is consistent?
To recall `amr`, this is Hash(imr, source_type, target_type). This is not only useful in directly checking sharing, but also useful for cross testing that our type inference is consistent with C implementation. 

Now that we have a good FFI, we could also be using `test_program` for testing execution, cmr, amr and imr computations. 

```C
static void test_program(
char* name, const unsigned char* program, 
size_t program_len,
bool expectedResult,
const uint32_t* expectedCMR,
const uint32_t* expectedIMR,
const uint32_t* expectedAMR
) {
```

uncomputable:
Resolves #54

### DIFF
--- a/src/core/commit.rs
+++ b/src/core/commit.rs
@@ -20,6 +20,7 @@ use crate::core::redeem::{NodeType, RedeemNodeInner};
 use crate::core::{Context, RedeemNode, Value};
 use crate::inference::UnificationArrow;
 use crate::jet::Jet;
+use crate::merkle::amr::Amr;
 use crate::merkle::cmr::Cmr;
 use crate::merkle::imr::Imr;
 use crate::merkle::{cmr, imr};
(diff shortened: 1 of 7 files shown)


sanket1729:
Can you directly pick up the code from https://github.com/ElementsProject/rust-simplicity/pull/19/commits/860d13c11850c42f55abc3d29ad2c2532bcc262a?

uncomputable:
Rebased on `master`. The FFI tests are removed now, so we cannot test it, but before, the AMR of Rust and C matched. We can think about bring back the FFI tests (in another PR) if we feel like the library is stable enough.

apoelstra (approved):
ACK 296e7f1e6cd94cba9dffd9bf55da3b70f4d47ab0

PR #56, opened 2023-01-17
Disable broken tests

uncomputable:
Some tests broke because of missing `c_jet_ptr`s. We didn't catch this because they were ignored.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,6 +53,26 @@ jobs:
           echo "No default features" && cargo test --no-default-features
           echo "All features" && cargo test --all-features
 
+  ignored:
+    name: Ignored Tests
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


sanket1729:
Can you also add a  commit that tests the ignored tests in CI?

uncomputable:
Done. Not sure why CI doesn't run as usual

sanket1729:
My guess is that CI file indentation might be wrong

uncomputable:
I used the same identifier for multiple jobs. It is fixed now.

sanket1729 (approved):
ACK 6f7aca2f616c2661c2f2ffbeab194d412d848a8e

sanket1729:
FTR, these tests were broken even before we have `c_jet_ptr`. Hopefully, we don't get this in this situation now that we check this in CI

PR #57, opened 2023-01-17
move encode/decode outside the autogen file

sanket1729:
Simple move only change. This should allow us to test things without waiting for `encode`, and `decode` implementation from Haskell.


apoelstra (approved):
ACK d17ef4a3c1cfbf5cedc1ae51e73e70027603e748

PR #60, opened 2023-01-18
Add scripts to update autogenerated jets code

sanket1729:
For now, the warning for checking both simplicity repos have the same
git rev is disabled. It will be re-enabled as soon the PR for generating C
jets in simplicity is merged

PR #61, opened 2023-01-23
Add scripts to update autogenerated jets code

sanket1729:
Supercedes #60

For now, the warning for checking both simplicity repos have the same
git rev is disabled. It will be re-enabled as soon the PR for generating C
jets in simplicity is merged


update_jets.sh
```diff
@@ -0,0 +1,43 @@
...
+SRC_DIR=$1
+REV=$2
+ORIGDIR=$(pwd)
+
+cd "$SRC_DIR" || exit 1
+if [ -n "$REV" ]; then
+    git checkout "$REV"
+fi
+HEAD1=$(git rev-parse HEAD)
+cd $ORIGDIR
```
apoelstra, line 20: In 8a8be85d857af7a55d2c8076f96eaee5957c524b:

should add quotes around `$ORIGDIR`

update_jets.sh
```diff
@@ -0,0 +1,43 @@
+#!/bin/bash
+set -e
+
+
+if [ -z "$1" ]; then
+  echo "\$1 parameter must be the simplicity directory"
+  echo "\$2 parameter (optional) can be the revision to check out"
```
apoelstra, line 7: In 8a8be85d857af7a55d2c8076f96eaee5957c524b:

IMO we shouldn't support specifying a revision. It's surprising that a script called `update-jets.sh` would cd into the source directory and then change files there.

Also, we should have the jet script default to `$(git rev-parse --show-toplevel)/simplicity-sys/depend/simplicity/` if it is not supplied, since going forward that's likely to be the common case.

update_jets.sh
apoelstra: There are several other cases of missing quotes -- can you run `shellcheck` locally on this file to get a full list? I'm happy to comment them all but it might get annoying :)

update_jets.sh
apoelstra: Ah, I see, the bundled version of simplicity does not include any Haskell code. That's a little unfortunate because it makes it hard to ensure you're running `update-jets.sh` with the same version of the Simplicity code as is bundled. I guess, it's fine for now.

apoelstra: Alternate strategy: if $2 is not provided, read the commit ID from `"$(git rev-parse --show-toplevel)"/simplicity-sys/depend/simplicity-HEAD-revision.txt`. In fact, even if $2 **is** provided, warn if the provided value differs from this value.

Then in the source directory, rather than doing `git checkout $REV`, we should check that the output of `git rev-parse HEAD` matches `$REV`, and output an error if they don't match.

apoelstra:
Can you indicate in your commit message for 555975f21113d31f751720cf11601849076a213a which Simplicity commit you used to generate the jet code?

Also, if you manually added `#[rustfmt::skip]` stuff and ran rustfmt, that should happen in a second commit.

update_jets.sh
sanket1729: Agreed

sanket1729:
Force pushed as per new suggestions. Currently, not all commits compile because there the auto-generated code misses an import statement. But the rest should be good to review.

apoelstra:
NIce!

a0c346dab3d87135b1c941c785d1b8b1828b9e03 looks great. Haven't test-run the script yet and haven't run my local CI on every commit. Will wait for the missing import statement in https://github.com/ElementsProject/simplicity/pull/106 to be resolved.

roconnor-blockstream:
Let me know if things like command line arguments would be helpful for you folks.

sanket1729:
> Let me know if things like command line arguments would be helpful for you folks.
`cabal build` followed by `cabal exec GenRustJets` is good enough for us. 

This is ready for review now.

sanket1729:
This is ready for review now.

apoelstra:
Can you update the PR to use simplicity master 28dbacdedc06d444e8a35c8b53a9be332126e166 ?

sanket1729:
There is some more done for the second PR because 28dbacdedc06d444e8a35c8b53a9be332126e166 (latest master as of Feb 14th) includes word jets. 

The updates FFI static test vectors for `schnorr0` and `schnorr6` use these fixed test vectors.

update_jets.sh
```diff
@@ -0,0 +1,41 @@
...
+cd "$C_DIR" || exit 1
+if [ -n "$REV" ]; then
+    if [ "$REV" != "$VENDORED_HEAD" ]; then
+        echo "WARNING: Simplicity GenRustJets and simplicity depend do not have the same git rev"
+        echo "$REV"
+        echo "$VENDORED_HEAD"
+    fi
+    git checkout "$REV"
+else
+    git checkout "$VENDORED_HEAD"
```
apoelstra, line 27: In 28ee0c22fb42b5eb7c26371e67846aca4280be79:

I still don't like that we're running `git checkout` which is potentially destructive. But offhand I don't know what the "correct" solution is (probably something obscure like `git read-tree --prefix`) and anyway I'd like to get this PR in. So we can address it in a followup.

apoelstra (approved):
ACK 3acf97461e87d9f88bcd0f41ebca1da69c8c765b

PR #62, opened 2023-01-23
Add tests from FFI

sanket1729:
Fixes several bugs with tmr/imr/cmr respectively. Based on #61 . All the tests based on FFI are still disabled. We will re-enable them in the next PR.

### DIFF
--- a/src/merkle/cmr.rs
+++ b/src/merkle/cmr.rs
@@ -51,7 +51,7 @@ impl CommitMerkleRoot for Cmr {
             CommitNodeInner::Witness => Cmr::tag_iv(b"Simplicity-Draft\x1fCommitment\x1fwitness"),
             CommitNodeInner::Fail(_, _) => Cmr::tag_iv(b"Simplicity-Draft\x1fCommitment\x1ffail"),
             CommitNodeInner::Hidden(h) => *h,
-            CommitNodeInner::Jet(jet) => jet.cmr(),
+            CommitNodeInner::Jet(j) => Cmr::tag_iv(b"Simplicity-Draft\x1fJet").update_1(j.cmr()),
         }
     }
 }
--- a/src/test_progs/mod.rs
+++ b/src/test_progs/mod.rs
@@ -13,90 +13,34 @@
 // If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
 //
 
-/// Test programs for simplicity.
-pub(crate) mod hashblock;
-pub(crate) mod schnorr0;
-pub(crate) mod schnorr6;
-pub(crate) mod sighash_all;
-
 #[cfg(test)]
+#[cfg(feature = "test-utils")]
 mod tests {
-    use super::hashblock::{HASHBLOCK, HASHBLOCK_CMR};
+
+    use simplicity_sys::c_jets::exec_ffi;
+
     use crate::bititer::BitIter;
-    use crate::core::{CommitNode, RedeemNode, Value};
-    use crate::exec::BitMachine;
-    use crate::jet::Core;
+    use crate::jet::Elements;
     use crate::merkle::common::MerkleRoot;
-    use crate::test_progs::schnorr0::{SCHNORR0, SCHNORR0_CMR};
-    use crate::test_progs::schnorr6::{SCHNORR6, SCHNORR6_CMR};
+    use crate::RedeemNode;
 
-    // TODO: check IMR
-    fn check_merkle_roots(bytes: &[u8], cmr: [u8; 32]) {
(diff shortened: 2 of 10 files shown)


src/merkle/cmr.rs
```diff
@@ -51,7 +51,7 @@ impl CommitMerkleRoot for Cmr {
             CommitNodeInner::Witness => Cmr::tag_iv(b"Simplicity-Draft\x1fCommitment\x1fwitness"),
             CommitNodeInner::Fail(_, _) => Cmr::tag_iv(b"Simplicity-Draft\x1fCommitment\x1ffail"),
             CommitNodeInner::Hidden(h) => *h,
-            CommitNodeInner::Jet(jet) => jet.cmr(),
+            CommitNodeInner::Jet(j) => Cmr::tag_iv(b"Simplicity-Draft\x1fJet").update_1(j.cmr()),
```
uncomputable, line 54: That line depends on the value of `jet.cmr()`. Originally, I intended this to include any IV, for efficiency reasons. Does the C code not include the IV?

src/test_progs/mod.rs
```diff
@@ -13,90 +13,34 @@
...
-        exec_prog(&SCHNORR6);
-    }
-
-    /*
-    TODO: Add c_jet_ptr for used jets and re-enable tests
-
-    #[test]
-    #[ignore] // too expensive to run. Run with -- --ignored and --release to run this
-    fn schnorr0() {
-        exec_prog(&SCHNORR0);
```
uncomputable, line 99: We still want to execute these programs on the Rust Bit Machine, right?

sanket1729: These tests are massively outdated. Would be better just rewrite these using new values.

src/merkle/cmr.rs
sanket1729: We don't have any static initializing, we are not gaining any efficiency. I did this because that's how the cmr_iv is used in imr calculation. Specifically, 
```rust
impl CommitMerkleRoot for Imr {
    fn get_iv<J: Jet>(node: &CommitNodeInner<J>) -> Self {
        match node {
            CommitNodeInner::Disconnect(_, _) => {
                Imr::tag_iv(b"Simplicity-Draft\x1fIdentity\x1fdisconnect")
            }
            CommitNodeInner::Witness => Imr::tag_iv(b"Simplicity-Draft\x1fIdentity\x1fwitness"),
            _ => Cmr::get_iv(node).into_inner().into(),
        }
    }
}
```

This would also require a jet branch which felt unclean to me. That's why included it here.

uncomputable: I'm not sure if the Rust compiler is smart enough to convert [our current code](https://github.com/sanket1729/rust-simplicity/blob/update_tests_ffi/src/jet/init/core.rs#L445) into static return values. But, couldn't we change the code to a match statement that returns `Cmr(Midstate([...]))` in each branch? That is a value that is `const`, and the compiler should be smart enough to save this as a static in the program.

uncomputable:
Let's rebase this one.

sanket1729:
@apoelstra, this is ready for review. We only add the tests in the PR, but do not actually test them. These would fail because the FFI test cases use word jets which are not currently supported.

apoelstra:
Looks good to me. I'm a little annoyed about needing to create a hashmap and iterate over the whole thing rather than caching IMRs. But in a future PR we can maybe revisit this. I suspect that we could combine the IMR calculation into one of the construction passes (maybe typechecking) so that we can compute this without needing to mutate anything or use hashmaps.

The FFI stuff also looks a little bit fragile but I don't see any way around it. I think it should cause test failures if this stuff gets out of sync. In future we could maybe change the Haskell GenTests, which produces the schnorrN.c files to also produce Rust bindings. But for now there's no need, there are only a couple test vectors and they don't change much.

apoelstra (approved):
ACK b97f0ed82d9758cd43e5f4ef4c6f600fdf9ff918

sanket1729:
> Looks good to me. I'm a little annoyed about needing to create a hashmap and iterate over the whole thing rather than caching IMRs. But in a future PR we can maybe revisit this. I suspect that we could combine the IMR calculation into one of the construction passes (maybe typechecking) so that we can compute this without needing to mutate anything or use hashmaps.

I don't think this is possible. The first must be done completely before proceeding to the second pass. The first pass values in imr depend on first pass imr values in the first imr pass of children. So, we cannot cannot throw away first pass values. FWIW, the C implementation also allocates vector for first pass and uses the stored values to compute the second pass. This has to be hashmap here because of rust-simplicity design decisions.

PR #64, opened 2023-03-08
Add benchmarks for simplicity jets

sanket1729:
The current commited benchmarks are good, but are not robust. They are only run on one sample. I will update the benchmarks file after they are run overnight. 

- There are instructions in `jet_benches/README.md` that mention how to run benchmarks. It's basically just `cargo-criterion` followed by `cargo run`. 
- Worst case reasoning for each jet: https://gist.github.com/sanket1729/4f038a0761feb369231f71f0c3b9e332 . For each jet, we run 25 samples multiple times and choose the worst case amongst those. 
- Sample output. This is stored in data directory. https://gist.github.com/sanket1729/0bf92ab9b2d17895d4afdfe3a85bdf70


sanket1729:
Unforetunately, this breaks MSRV. There are no optional dev-dependancies. There should be an easy way to fix this. I will investigate this later. But that should not stop review here.

src/core/types.rs
```diff
@@ -100,6 +100,26 @@ impl Type {
             two_1, two_2, two_4, two_8, two_16, two_32, two_64, two_128, two_256,
         ]
     }
+
+    /// Return an array `pow2s` of types such that `pow2s[i] = 2^i` holds for 0 ≤ `i` < n.
+    pub fn powers_of_two_vec(n: usize) -> Vec<Arc<Type>> {
+        assert!(n < 16);
```
apoelstra, line 106: In 4838302e05e8b1b0b4ef41228cdc96cedb57e1d8:

We could probably drop this assertion. But if we keep it, we should put it in the doccomment.

src/core/value.rs
```diff
@@ -127,6 +131,36 @@ impl Value {
         )
     }
 
+    /// The notation X^<2 is notation for the type (S X)
+    /// The notation X^<(2*n) is notation for the type S (X^n) * X^<n
+    /// Cannot represent >= 2**16 bytes 0 <= n < 16.
+    ///
+    /// # Panics:
+    ///
+    /// Panics if the length of the slice is >= 2^(n + 1) bytes
```
apoelstra, line 146: In 336740d2067f77f868d02823e799a448f811c139:

This doccomment doesn't say anything about the function that it's documenting, and it doesn't even use the notation that it defines!

sanket1729: That notation is well defined in C simplicity and simplicity in general. But I agree, I should use the notation here instead of copy-pasting from C implementation.

src/core/value.rs
```diff
@@ -127,6 +131,36 @@ impl Value {
...
+        let mut res = None;
+        while n > 0 {
+            let v = if v.len() >= (1 << (n + 1)) {
+                let ty = &types[n];
+                let val = decode::decode_value(ty, &mut iter)?;
+                Value::SumR(Box::new(val))
+            } else {
+                Value::SumL(Box::new(Value::Unit))
+            };
+            match res {
```
apoelstra, line 161: In 336740d2067f77f868d02823e799a448f811c139:

Should pull out the `res =`, as
```
res = match {
    Some(prod) => Some(Value::Prod(Box::new(prod), Box::new(v)),
    None => Some(v)
};
```
or even
```
res = Some(res.map_or_else(
    || Some(v),
    |prod| Value::Prod(Box::new(prod), Box::new(v))),
));
```

sanket1729:
@apoelstra, my solution right now is to make a application separate crate inside this repo that has it's own dependencies. Adding dev-dependancies here while making MSRV happy is harder to do now and maintain in future.

src/core/value.rs
apoelstra: Actually, on further reading, I think it would be easier to just early-return when `n == 0`, then `res` doesn't need to be an option anymore and the final `unwrap_or` can be dropped.

src/core/value.rs
```diff
@@ -116,6 +116,12 @@ impl Value {
         Value::Prod(Box::new(Value::u32(w0)), Box::new(Value::u32(w1)))
     }
 
+    pub fn u128(n: u128) -> Value {
+        let w0 = (n >> 64) as u64;
+        let w1 = (n & 0xffff_ffff_ffff_ffff) as u64;
```
apoelstra, line 121: In c4e2f196c437980d3b5947839dbd20a8850bfa9a:

This masking isn't necessary. It doesn't really matter, but it does add a bit of review burden to have to count the `f`s.

apoelstra:
Done reviewing up to c4e2f196c437980d3b5947839dbd20a8850bfa9a (everything but the final two commits, which have the actual content :))

src/core/value.rs
sanket1729: Initializing `res` would essentially be unrolling the first iteration of the loop. Not really sure if that would be better.

apoelstra: Ah, yeah, I see. Agreed, probably not better to unroll it.

sanket1729:
Added fixup commits addressing the comments. Will squash them before merging. Also made the benchmarks into separate repo to not add any additional dev-dependancies.

sanket1729:
Updated the description too.

jets_bench/benches/elements/main.rs
```diff
@@ -0,0 +1,643 @@
...
+                    || {
+                        let mut buffer = JetBuffer::new(&src_ty, &tgt_ty, params);
+                        let (src, dst) = buffer.write(&src_ty, params, &mut rng);
+                        (dst, src, &env, buffer)
+                    },
+                    |(mut dst, src, env, _buffer)| jet.c_jet_ptr()(&mut dst, src, env.c_tx_env()),
+                    BatchSize::SmallInput,
+                )
+            });
+        }
```
uncomputable, line 373: This creates a sub-benchmark for every separate input we test. Is that what we want? An obvious upside is that we can use `bench_with_input` (which is canonical, but I haven't yet understood when it should be used). A downside is that there is a 3 second warmup for every sub-benchmark (which might improve accuracy, but is it really needed?) and the resulting criterion data is slightly harder to process in Python (maybe cargo-criterion can compute percentiles). Depending on what we want, 50 samples might be too few for a meaningful distribution.

The alternative would be to write a custom time loop which is called multiple times by criterion, and which makes sure that the same input is used for each sample. Obvious downside is that this is not canonical criterion use.

uncomputable: There was another push. I see now that we do a two-step approach: First we run sub-benchmarks for `n` inputs that we think are worst-case. Then, we run the `main.rs` script to find the worst among them.

sanket1729: I think even 10 samples should be good enough. The distributions are selected from white-listing what we consider to be the worst case input. I would even argue that even 1/2 is enough, but that might be off because some of the jets do actually randomly select envs.
 
Regarding a custom timing loop, I am not against it. This is just meant as a starter that we have something benchmarks ongoing while we discuss specifics. My intuition is that it would not matter, so feel free to have a go at this with a custom loop. I think would make things harder to review because we are calculating our own times, starting and stoping clocks, etc. 

Regarding warm up time etc, time per loop etc. All of this is configurable. Ultimately, we are doing the same calculation in the custom time loop vs here. Just that the default parameters here make it seem like taking longer(which might be good). Something like benchmark `measurement_time ` 100ms with 500 num samples would be same as one would do in custom loop.

uncomputable:
We definitely should update `libsimplicity`. Maybe as a follow-up PR. The field elements should be normalized via `fe_normalize_var`.

sanket1729 (commented):
some self review

jets_bench/benches/elements/main.rs
sanket1729: > There was another push. I see now that we do a two-step approach: First we run sub-benchmarks for n inputs that we think are worst-case. Then, we run the main.rs script to find the worst among them.

Yep, this was what I always intended to do. This is done in last commit. (modulo fixup commits)

uncomputable:
[There was a recent change](https://github.com/BlockstreamResearch/simplicity/commit/d5d9730276e84e45eed897268d759fddba4133b3). This is in `read_fe` which is called by practically every EC jet.

sanket1729:
Ah, I thought you were referring the way Fe points were sampled.

sanket1729:
I think it should be done in this PR to get correct benchmarks. We already have automated scripts for doing these

sanket1729:
@uncomputable, pushed an update that uses latest simplicity. Thanks for pointing this out, I will re-run my benchmarks again.

jets_bench/benches/elements/buffer.rs
```diff
@@ -0,0 +1,91 @@
...
+            unsafe { c_readBit(&mut src_read_frame) };
+        }
+
+        // Destination frame:
+        // Write dummy bits for alignment and return write frame
+        let mut dst_write_frame =
+            unsafe { CFrameItem::new_write(self.dst_bit_width, self.dst_ptr_end) };
+        for _ in 0..params.tgt_align {
+            unsafe { c_writeBit(&mut dst_write_frame, false) };
+        }
```
apoelstra, line 87: In 5b5e6d2c98c317bba9ec82b98dd29c70122c4cf6:

I'd put basically this whole function in a big `unsafe` block rather than wrapping every individual line (or even parts of lines).

jets_bench/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+#[derive(Clone, PartialEq, Deserialize, Debug)]
+pub struct ConfidenceInterval {
+    pub confidence_level: f64,
+    pub lower_bound: f64,
+    pub upper_bound: f64,
+}
+
+fn main() {
```
apoelstra, line 50: In 4ff2efd471d0417419a3279b664f85d2079b1249:

This can wait for a future PR but we should definitely change this to use `anyhow` and return sensible errors rather than unwrapping everywhere.

apoelstra:
utACK 5ca5f17abd355affecf28ee91e1614ec4629c794

Need the fixups to be squashed, since in particular without the feature gates on `pub use bitcoin` etc early commits don't compile.

sanket1729:
Rebased and ready for review. As discussed on call, we should merge this as is without the exact benchmarks from it. In particular, there are two changes to be done before we run the benchmarks. 

- Update secp 0.3.0
- More fine-tuning for benchmarks so that this does not take super long time to run. My idea here is to have more samples for secp jets and less for other jets.

apoelstra (dismissed):
ACK 09d2ece2a12b301619681cf2a8a7f6d2c41a6e74

uncomputable:
The self-review comments still seem unaddressed. I would also remove the secp update from this PR since we decided to make a separate PR. The rest looks good.

jets_bench/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub median_abs_dev: Estimate,
+    pub slope: Option<Estimate>,
+    pub std_dev: Estimate,
+}
+
+#[derive(Clone, PartialEq, Deserialize, Debug)]
+pub struct Estimate {
+    /// The confidence interval for this estimate
+    pub confidence_interval: ConfidenceInterval,
+    ///
```
uncomputable, line 37: Missing doc?

jets_bench/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+                })
+                .collect::<Vec<_>>();
+            // We expect only one file
+            assert_eq!(measurement_files.len(), 1);
+            let file = measurement_files.pop().unwrap();
+
+            let file = std::fs::File::open(file).unwrap();
+            let data: MeasurementData = serde_cbor::from_reader(file).unwrap();
+
+            // Amongst
```
uncomputable, line 96: Confusing comment

uncomputable:
ACK https://github.com/BlockstreamResearch/rust-simplicity/commit/09d2ece2a12b301619681cf2a8a7f6d2c41a6e74

sanket1729:
rebased and pushed

apoelstra (approved):
ACK 9fba126950b3a1a079855b33f78af49c07d47197

PR #66, opened 2023-03-10
Policy Satisfier

uncomputable:
Extends #65 .

Generate Simplicity witness data from a policy, environment and sets of keys and preimages.


src/jet/elements/environment.rs
```diff
@@ -116,6 +116,31 @@ impl ElementsEnv {
     }
 }
 
+impl Default for ElementsEnv {
```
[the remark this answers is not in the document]
sanket1729: Perhaps, what we need is 
```rust
#[cfg(test)]
impl ElementsEnv {

    pub fn dummy_env() -> Self {
    }
}
```

src/jet/elements/environment.rs
```diff
@@ -53,21 +53,21 @@ pub struct ElementsUtxo {
...
+    pub(crate) script_cmr: Cmr,
     /// Control block used to spend this leaf script
-    pub(super) control_block: ControlBlock,
+    pub(crate) control_block: ControlBlock,
     /// Optional Annex.
-    pub(super) annex: Option<Vec<u8>>,
+    pub(crate) annex: Option<Vec<u8>>,
     /// Genesis block hash
-    pub(super) genesis_hash: BlockHash,
+    pub(crate) genesis_hash: BlockHash,
```
sanket1729, line 70: I think these should be pub(super). And have pub accessor methods to access these elsewhere in the codebase. I don't think others should be setting anything in ElementsEnv. Once, it is created it should not mutated.

simplicity-sys/Cargo.toml
```diff
@@ -10,6 +10,7 @@ cc = "1.0"
 
 [dependencies]
 libc = "0.2"
+bitcoin_hashes = "0.11"
```
sanket1729, line 13: No need to use seperate dependancy. I think we can get it from `elements::bitcoin::hashes`

simplicity-sys/src/c_jets/c_env.rs
```diff
@@ -199,6 +201,12 @@ extern "C" {
     pub fn c_free_transaction(tx: *mut CTransaction);
     pub fn c_free_tapEnv(env: *mut CTapEnv);
 }
+impl CElementsTxEnv {
+    pub fn sighash_all(&self) -> sha256::Hash {
```
sanket1729, line 205: nit: Not to be done in this PR. But ideally, we would want this to new `hash_newtype` like `SimplicitySigHash`

src/policy/satisfy.rs
```diff
@@ -0,0 +1,82 @@
...
+use elements::schnorr::{KeyPair, XOnlyPublicKey};
+use elements::secp256k1_zkp::Message;
+use std::collections::HashMap;
+
+#[allow(unused)]
+struct Parameters {
+    pub preimages: HashMap<sha256::Hash, [u8; 32]>,
+    pub keys: HashMap<XOnlyPublicKey, KeyPair>,
+    pub env: ElementsEnv,
+}
```
sanket1729, line 14: In rust-miniscript, we have a satisfier trait in rust-miniscript to provide abstraction over these values. We have a logic where we compare which satisfaction cost is cheaper when given satisfactions. The API then tries to find the witness with minimum cost. 

This I think is searching if there exists a solution and greedily picking the first one possible. Fine for now, but definitely something we should improve.

uncomputable: I will try to use the `Satisfier` trait from rust-miniscript. The signatures for Elements transactions confused me, but a custom implementation might work. The taproot stuff will need some working out since we are using a different leaf version (I think).

src/jet/elements/environment.rs
uncomputable: I see. That is the difference between member visibility and getters. Got it.

simplicity-sys/Cargo.toml
uncomputable: `simplicity-sys` is a separate crate with separate dependencies. Importantly, the crate cannot use (rust-)`simplicity` or its dependencies. I can add an `elements` dependency, but it this better than `bitcoin_hashes`, which is already a dependency of `simplicity`?

uncomputable: We have to keep `simplicity` and `simplicity-sys` in sync, dependency-wise, but I think we can manage.

src/policy/satisfy.rs
uncomputable: I don't think I can use `miniscript::Satisfier` directly, since it expects a Bitcoin transaction. I implemented a satisfaction cost that is used to decide on the best disjunctive branches. Thresholds are currently not supported because the underlying Simplicity expression (that I wrote) is pretty horrible (it expects witnesses for unused branches; I will fix that in a future PR).

apoelstra: Can you use `CovSatisfier` from `elements-miniscirpt`?

apoelstra: Oops, that's a struct. I meant to say `Satisfier`: https://docs.rs/elements-miniscript/latest/elements_miniscript/miniscript/satisfy/trait.Satisfier.html

uncomputable:
Implemented `elements_miniscript::Satisfier` for a custom type called `PolicySatisfier`. This is used in the `satisfy` method to generate efficient Simplicity witnesses, weighted by their length in bits.

simplicity-sys/Cargo.toml
sanket1729: Yep, this is needed. Sorry about that.

src/jet/elements/environment.rs
```diff
@@ -19,7 +19,7 @@ use bitcoin_hashes::{sha256, Hash, HashEngine};
 use byteorder::{BigEndian, LittleEndian, WriteBytesExt};
 use elements::confidential::{Asset, Nonce, Value};
 use elements::taproot::ControlBlock;
-use elements::{confidential, AssetIssuance, BlockHash};
+use elements::{confidential, AssetIssuance, BlockHash, Transaction};
```
apoelstra, line 22: In 158b54dfe7526b3f3c51cbe9d988e46c05c4aa95:

I really find it hard to read `Transaction` without either a `bitcoin::` or `elements::` prefix. Ditto for `BlockHash` though it's less of a problem since the two kinds of blockhashes are basically the same thing.

src/jet/elements/environment.rs
```diff
@@ -100,6 +100,16 @@ impl ElementsEnv {
         &self.c_tx_env
     }
 
+    /// Returns the transaction of this environment
+    pub fn tx(&self) -> &Arc<Transaction> {
+        &self.tx
+    }
```
apoelstra, line 106: In 8365c621bbc93d1c82f38ce53c1f9073d90ac339:

You should return a `&Transaction` rather than leaking the internal detail that you have an `Arc`. You could add a separate accessor to get a clone of the `Arc` if you want.

TBH I'm not sure why we have an `Arc` here at all. Why can't the `ElementsEnv` directly own its `Transaction`? Or use a normal reference? You could maybe parameterize it over `Deref<Target = Transaction>` like we do with `SighashCache`.

src/policy/satisfy.rs
```diff
@@ -0,0 +1,143 @@
...
+    fn check_after(&self, n: LockTime) -> bool {
+        let m = LockTime::from(self.env.tx().lock_time);
+        match m.partial_cmp(&n) {
+            Some(Ordering::Less) | Some(Ordering::Equal) => true,
+            _ => false,
+        }
+    }
+}
+
+pub fn satisfy(policy: &Policy<XOnlyPublicKey>, satisfier: &PolicySatisfier) -> Option<Vec<Value>> {
```
apoelstra, line 57: In 243e7d6920adbcb1235f1b23de36abb792022614:

I think this architecture where there's just a `satisfy` function that takes a `PolicySatisfier` is good for now. In a later PR we definitely want to abstract over different kinds of satisfiers (in fact, we should be able to just take an arbitrary `elements_miniscript::Satisifier` so we don't even need our own trait).

src/jet/elements/environment.rs
uncomputable: Ok, this is a convention I didn't know about. Good to know and I will make sure to follow it from now.

src/jet/elements/environment.rs
uncomputable: I don't see any uses of `tx` in the library that would not be possible with a reference. Maybe the `Arc` is a "historic artifact". @sanket1729 Any ideas?

src/policy/satisfy.rs
uncomputable: I rebased and implemented `satisfy` for arbitrary `Satisfier<XOnlyPublicKey>`. In a later PR, we should transform this into `Satisfier<Pk>`, but that requires `Policy` to be slightly rewritten in terms of `elements_miniscript`.

src/jet/elements/environment.rs
apoelstra: It's mostly just a convention for Elements software, which always needs to deal with both kinds of objects.

src/jet/elements/environment.rs
sanket1729: There is a big comment here on why it was needed: https://github.com/BlockstreamResearch/rust-simplicity/blob/a57e6a4f32f0ef04839329621bed861ce09bf076/src/jet/elements/environment.rs#L45-L53

sanket1729: But we don't have extension trait anymore. I think can be cleaned up to use &'a or AsRef.

uncomputable:
Is there something left to change, or is this ready to merge?

apoelstra:
Sorry @uncomputable, I lost this tab in a browser crash last week. I will test and ACK it today. It looks good to me. In a later PR we should be able to easily generalize it from x-only pubkeys.

apoelstra (approved):
ACK 7090586

PR #69, opened 2023-04-19
policy: do not parse invalid thresholds

apoelstra:
Right now we are willing to parse policies like `thresh(0)` which are not valid. If you pass these to the compiler it will panic, meaning that a user cannot simply "parse then compile" without their code panicking.

This fixes that bug (3 lines) and adds a bunch of extra tests (the rest).


sanket1729 (approved):
ACK 0fc625f4e3e236a7262178ffa08cd293df5ed09d

PR #70, opened 2023-04-28
API improvements

uncomputable:
Implements useful traits and exports useful structs.

### DIFF
--- a/src/jet/elements/environment.rs
+++ b/src/jet/elements/environment.rs
@@ -28,6 +28,7 @@ use super::c_env;
 /// An Elements UTXO
 // This is not a complete TxOut as it does not contain the nonce that
 // is sent to the recipient.
+#[derive(Debug, Clone, Eq, PartialEq)]
 pub struct ElementsUtxo {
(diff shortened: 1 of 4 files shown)


src/jet/elements/environment.rs
```diff
@@ -28,6 +28,7 @@ use super::c_env;
 /// An Elements UTXO
 // This is not a complete TxOut as it does not contain the nonce that
 // is sent to the recipient.
+#[derive(Debug, Clone, Eq, PartialEq)]
 pub struct ElementsUtxo {
     /// The 'scriptpubkey' (hash of Simplicity program)
     pub script_pubkey: elements::Script,
```
apoelstra, line 34: In d7cbe839764e1cf4059ef72d34d9d51f8c5ee21d:

Not gonna hold up this PR, but we should make this generic over `E: AsRef<elements::Script>` so that people can put references in here. Then we could also #derive `Copy` which would be nice.

apoelstra (approved):
ACK 32e748cc409f6cfdea565d4fc05b0f0b19fb3ea4

PR #71, opened 2023-04-28
Policy improvements

uncomputable:
Makes simplicity::Policy ready for tappy. Depends on #72

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -16,7 +16,6 @@
 
 extern crate bitcoin_hashes;
 extern crate byteorder;
-extern crate miniscript;
 
 #[cfg(feature = "serde")]
 pub use actual_serde as serde;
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -16,7 +16,7 @@
 
 extern crate bitcoin_hashes;
 extern crate byteorder;
-extern crate miniscript;
+extern crate elements_miniscript as miniscript;
```
apoelstra, line 19: In 79387fd1541cf6d067641cace4288a31149bbb52:

If we are going to replace `miniscript` with `elements_miniscript` and keep the name, why did the previous commit in this PR rename `miniscirpt` to `elements_miniscirpt` in a bunch of places?

Also all these `extern crate`s should be pub IMO.

apoelstra:
82c346808a2282f44bd0d647c2fe820c9f1645b5 looks good except the nit about miniscript

src/lib.rs
uncomputable: I changed `elements_miniscript` to `(crate::)miniscript` and replaced `extern crate` with `pub use`.

uncomputable:
Ready for another review

apoelstra (approved):
ACK 9fd00a79c533a8d25c1a60ac6cc47b8c1830ab64

PR #74, opened 2023-05-01
Some fixes for Elements compatibility

apoelstra:
Fixes two bugs discovered when using `testmempoolaccept` to test Simplicity spends.


sanket1729:
Ah, fuzz failure !

apoelstra:
Oh, nice, the problem was that I changed the encoder without changing the decoder.

uncomputable:
ACK d0d93ae

uncomputable (approved):

PR #75, opened 2023-05-10
Fix compiler

uncomputable:
Correctly implements and tests timelocks. Also tests key locks.

### DIFF
--- a/src/jet/elements/environment.rs
+++ b/src/jet/elements/environment.rs
@@ -141,6 +141,11 @@ impl ElementsEnv {
 impl ElementsEnv {
     /// Return a dummy Elements environment
     pub fn dummy() -> Self {
+        Self::dummy_with(elements::PackedLockTime::ZERO, elements::Sequence::MAX)
+    }
+
+    /// Return a dummy Elements environment with given locktime
+    pub fn dummy_with(lock_time: elements::PackedLockTime, sequence: elements::Sequence) -> Self {
         let ctrl_blk: [u8; 33] = [
             0xc0, 0xeb, 0x04, 0xb6, 0x8e, 0x9a, 0x26, 0xd1, 0x16, 0x04, 0x6c, 0x76, 0xe8, 0xff,
             0x47, 0x33, 0x2f, 0xb7, 0x1d, 0xda, 0x90, 0xff, 0x4b, 0xef, 0x53, 0x70, 0xf2, 0x52,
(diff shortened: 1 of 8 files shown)


src/jet/elements/environment.rs
```diff
@@ -141,6 +141,11 @@ impl ElementsEnv {
 impl ElementsEnv {
     /// Return a dummy Elements environment
     pub fn dummy() -> Self {
+        Self::dummy_with(0, elements::Sequence::MAX)
+    }
+
+    /// Return a dummy Elements environment with given locktime
+    pub fn dummy_with(locktime: u32, sequence: elements::Sequence) -> Self {
```
apoelstra, line 148: In 3adb9a528224e6fe5d1a88ca3c1d93f541acbd4e:

Why not just take a `PackedLockTime` directly instead of a `u32`?

src/policy/embed.rs
```diff
@@ -119,7 +119,7 @@ impl<'a, Pk: MiniscriptKey, Ctx: ScriptContext> TryFrom<&'a Miniscript<Pk, Ctx>>
                 Err(Error::ParseError("Public key hashes are not supported"))
             }
             Terminal::After(n) => Ok(Policy::After(n.0)),
-            Terminal::Older(n) => Ok(Policy::Older(n.0)),
+            Terminal::Older(n) => Ok(Policy::Older(n.0 as u16)),
```
apoelstra, line 122: In d975984dd4b6db1fc27ae55c1a175c4d6964167b:

Why is this cast safe?

src/policy/satisfy.rs
```diff
@@ -71,7 +71,7 @@ impl<Pk: MiniscriptKey + ToPublicKey> Policy<Pk> {
                 }
             }
             Policy::Older(n) => {
-                if satisfier.check_older(Sequence(*n)) {
+                if satisfier.check_older(Sequence(*n as u32)) {
```
apoelstra, line 74: In d975984dd4b6db1fc27ae55c1a175c4d6964167b:

If this is going from a u16 to a u32 you should use `.into()` rather than casting.

apoelstra:
d975984dd4b6db1fc27ae55c1a175c4d6964167b does not compile.

src/policy/embed.rs
uncomputable: I added a conversion similar to [to_relative_lock_time from rust-bitcoin](https://docs.rs/bitcoin/latest/bitcoin/blockdata/transaction/struct.Sequence.html#method.to_relative_lock_time). Unfortunately this function does not exist in rust-elements.

uncomputable:
Fixed the above issues

src/lib.rs
```diff
@@ -100,6 +100,9 @@ pub enum Error {
     SharingNotMaximal,
     /// Miniscript Error
     MiniscriptError(miniscript::Error),
+    /// Policy-related error
+    #[cfg(feature = "elements")]
+    Policy(policy::Error),
```
apoelstra, line 105: In 9e6381d84fd06d5d25d043597c4a1e83253d00b3:

You cannot feature-gate enum variants unless the enum is marked `#[non_exhaustive]`. Otherwise enabling the feature will break compilation for downstream users.

uncomputable: Having a separate policy error would be useful. Maybe I can put it in a non-feature-gated module?

apoelstra: I think it's fine to just mark the error enum as `#[non_exhaustive]`.

uncomputable: One rarely matches over all error variants, so this is probably fine

apoelstra (approved):
ACK 16e496b023c742d4a89ecb63d8ffff8362c1263d

PR #76, opened 2023-05-11
Fix Satisfier

uncomputable:
Depends on #75.

Implements pruning in the policy satisfier. This is required to make the produced satisfactions actually satisfy the Simplicity programs. Also tests that satisfactions satisfy.


apoelstra:
58638e52c6df59c6be09714fef0ec27daa167a30 looks good. The CI failure is real -- you'll need to adjust your code to work with 1.41.0. (We should bump the MSRV on this crate to 1.48.0 BTW, though I'm not sure if that will solve this particular issue.)

In general I have a couple comments:
* Does the compiler actually completely prune everything? It seems like, in principle, our "primitives" (sigchecks, timelocks, etc) might have some internal unused branches. I think when we start extending this to covenants we may run into this. This is fine, `finalize` will prune the rest, but I wonder if we should refer to the compiler's work as "pre-pruning" or something.
* I don't like the name `Helper`; in rust-miniscript we call a similar struct `CompilerExtData` and we should maybe copy the name to emphasize the similarity.
* Then we should call `empty_helper` `compile_no_witness` or something.
* Our costing model seems wrong ... right now we are basically only pricing witness data, but strictly speaking we should try to cost out the size of the program as well. In the presence of sharing this seems really hard, but let's keep it in mind.
* We should probably `finalize` before returning from `satisfy` and return a `RedeemNode`. I think it's confusing for the user that you have to do your own finalization.

None of these need to be addressed in this PR, though I suppose you could do the struct rename.

apoelstra:
Actually, sharing or no sharing, we will have a well-defined cost model in Elements which should be easy to compute :). But I think right now that model is a MVP and we haven't implemented it here. So that'll be easy to add once we have the cost estimation code in place.

uncomputable:
Really good ideas; I will integrate them into this PR in the next days.

The satisfier only prunes what is easy to prune (i.e. anything that doesn't require running the Bit Machine). The name is up for debate. "pre-pruning", "semi-pruning", "fast pruning", ... When we add pruning to `finalize()` we could remove this step.

IDK the full details of our cost model, but AFAIU the tx + witness size is one cost and the execution time another. Sharing makes the tx smaller, but execution time may stay constant because the same node is executed twice on different inputs. Which cost should we minimise with the satisfier?

apoelstra:
@uncomputable there's deliberately only one cost, otherwise this exact optimization problem would be NP-complete :) rather than being a trivial "take the minimum of the available options".

The execution time gets converted to transaction weight according to some formula and added to the actual transaction weight.

uncomputable:
I renamed `Helper`, changed the satisfaction output to `RedeemNode` and fixed CI.

The cost model is still open.

uncomputable:
The word "pruned" is actually not used in the code, besides a private member inside `SatisfierExtData`. Since we will add full pruning later, I think we can keep the name for now.

uncomputable:
Added a rudimentary cost measure and refactored the satisfier. @apoelstra Ready for another review

uncomputable:
Recreating the program during satisfaction creates an overhead. The satisfier tests are about as slow as the compiler now tests, probably due to heap usage (before they were instant).

src/policy/satisfy.rs
```diff
@@ -227,112 +232,170 @@ mod tests {
...
 
-        let witness = policy.satisfy(&satisfier).expect("sat");
+        let (witness, commit) = policy.satisfy(&satisfier).expect("satisfiable");
         assert_eq!(1, witness.len());
-        let signature = witness[0].try_to_bytes().expect("bytes");
-        assert_eq!(64, signature.len());
+
+        let sighash = satisfier.env.c_tx_env().sighash_all();
+        let message = secp256k1_zkp::Message::from(sighash);
+        let signature_bytes = witness[0].try_to_bytes().expect("to bytes");
```
apoelstra, line 292: In ec2271e5eabfe3055b25652472041e15f06fb82b:

This `try_to_bytes` followed by `from_slice` thing is ok for now (especially since the existing code does it) but we really need a dedicated method to extract 32-byte arrays from `Value`s.

src/core/commit.rs
```diff
@@ -203,6 +203,14 @@ impl<J: Jet> CommitNode<J> {
         self.arrow.target_ty()
     }
 
+    pub fn iter(&self) -> PostOrderIter<RefWrapper<J>> {
+        RefWrapper(self).iter_post_order()
+    }
+
+    pub fn size(&self) -> usize {
+        self.iter().count()
```
apoelstra, line 211: In b3f7a5e090101ed405fcbc53de1021ccaed18bf1:

Again, pre-sharing or post-sharing? Probably we should add a FIXME comment saying to something more efficient than counting every single node like this.

Also the function should be called `len` rather than `size`, no?

uncomputable: A graph doesn't have a length, but since we are calling these graphs "programs" a length is probably more intuitive.

src/policy/satisfy.rs
uncomputable: Created #80

src/core/commit.rs
```diff
@@ -203,6 +203,14 @@ impl<J: Jet> CommitNode<J> {
         self.arrow.target_ty()
     }
 
+    pub fn iter(&self) -> PostOrderIter<RefWrapper<J>> {
```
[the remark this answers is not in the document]
uncomputable: `RefWrapper` is defined in the same module. I can write `self::RefWrapper`. Not sure how I feel about importing the module in itself.

uncomputable:
Addressed the above issues

src/core/commit.rs
apoelstra: Iterators, hashmaps, btrees, etc., don't have lengths etiher, but in Rust it is conventional to use `len` to refer to the size operator, and `size` only to refer to encoding sizes.

src/core/commit.rs
```diff
@@ -203,6 +203,18 @@ impl<J: Jet> CommitNode<J> {
...
     }
 
+    /// Return an iterator over the unshared nodes of the program
+    pub fn iter(&self) -> PostOrderIter<self::RefWrapper<J>> {
+        RefWrapper(self).iter_post_order()
+    }
+
+    // FIXME: Compute length without iterating over entire DAG?
+    /// Return the number of unshared nodes in the program
+    #[allow(clippy::len_without_is_empty)]
```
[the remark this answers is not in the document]
apoelstra: In 22527ac I mean.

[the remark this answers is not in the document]
uncomputable: Programs are never empty. If you call `is_empty` on a node, that is your witness that the DAG is not empty.

apoelstra (approved):
ACK 93af555ffe43a35cca9e5f281c283bb2c53d4023

PR #77, opened 2023-05-14
Several improvements to ergronomics; eliminate tons of error paths

apoelstra:
Right now our type inference logic is far too eager to use the `?` operator even for error paths that are logically impossible. This PR refactors a bunch of things so that the constructors for most of the `CommitNode` nodes are infallible, `scribe` is infallible, etc., which eliminates a lot of downstream error-checking code.

### DIFF
--- a/src/inference.rs
+++ b/src/inference.rs
@@ -316,7 +316,7 @@ impl UnificationArrow {
                     CommitNodeInner::AssertL(..) => {
                         bind(
                             &find_root(lchild.arrow.source.clone()),
-                            VariableType::Product(a, c.clone()),
+                            VariableType::Product(a, c),
                             "Case: Left source = A × C",
                         )?;
                         lchild.arrow.target.clone()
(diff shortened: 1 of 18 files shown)


src/inference.rs
```diff
@@ -162,200 +172,230 @@ impl<J: Jet> TryFrom<&CommitNode<J>> for NodeType {
...
-                            &find_root(right_arrow.source.clone()),
+                            &find_root(lchild.arrow.source.clone()),
+                            VariableType::Product(a, c.clone()),
+                            "Case: Left source = A × C",
+                        )?;
+                        bind(
+                            &find_root(rchild.arrow.source.clone()),
                             VariableType::Product(b, c),
                             "Case: Right source = B × C",
                         )?;
```
uncomputable, line 342: We are calling `find_root` twice here because of the changes in `bind`. Originally `bind` didn't find the root because some arguments were guaranteed to be roots. This is no longer the case in this PR because these cases are handled differently.

uncomputable:
I like these changes. Especially since the `CommitNode` methods that could fail already had this written in their documentation. We missed to implement them accordingly.

src/inference.rs
apoelstra: Ah, good catch! Yep, I will remove all the `find_root` calls here.

apoelstra:
Addressed your comments in a new commit.

uncomputable:
#75 caused slight conflicts in `src/policy/compiler.rs`

apoelstra:
Rebased and fixed rustfmt. I have not tested every commit after the rebase yet.

apoelstra:
Ok, at least checked that `cargo test --all --all-features` passes on every commit. I think this is good to go.

uncomputable:
ACK 635d05a

uncomputable (approved):

PR #81, opened 2023-05-23
Satisfy threshold policy

uncomputable:
Adds satisfier for threshold policy. This is a little involved because of optimization.

First the `k` cheapest satisfiable children are determined. Those are sorted by index. Finally the satisfier goes through all children in index order and uses the satisfiable ones.

### DIFF
--- a/src/policy/satisfy.rs
+++ b/src/policy/satisfy.rs
@@ -492,4 +492,74 @@ mod tests {
         ]);
         assert!(policy3.satisfy(&satisfier).is_none());
     }
+
+    #[test]
+    fn satisfy_thresh() {
+        let satisfier = get_satisfier();
+        let images: Vec<_> = satisfier.preimages.keys().map(|x| *x).collect();
+        let preimages: Vec<_> = images
+            .iter()
+            .map(|x| satisfier.preimages.get(x).unwrap())
+            .collect();
+
+        let assert_branches = |policy: &Policy<bitcoin::XOnlyPublicKey>, bits: &[bool]| {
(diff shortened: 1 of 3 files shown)


src/policy/satisfy.rs
```diff
@@ -196,7 +217,94 @@ impl<Pk: MiniscriptKey + PublicKey32 + ToPublicKey> Policy<Pk> {
...
+                }
+
+                if satisfiable_children.len() < *k {
+                    return None;
+                }
+
+                // 1) Sort by witness cost
+                // 2) Select k best satisfactions
+                // 3) Sort by index in decreasing order
+                satisfiable_children.sort();
```
apoelstra, line 241: In 027ab50b45d65b555086036e6ccd588679013976:

Nit: using `sort_by_key` would be faster and avoid the need to implement `Ord`/`PartialOrd` on `SatisfierExtData`.

apoelstra:
Looks good to me. I did not review the new cost constants or the exact constructions in detail -- I think we will need to wait until we have the infrastructure to do automated testing against the C implementation before this is practical.

apoelstra (dismissed):
ACK caf8cd675049f25fa1556c7eedc53ee359bb572d

apoelstra:
I'm fine merging as-is; but if you want to fix the `sort` thing I'll let you

uncomputable:
I used `sort_by_key` and removed the `Ord` implementation. There might be other ways to order `SatisfierExtData` that I don't want to preclude.

apoelstra (approved):
ACK 0306408e4bdd9f05e4da0b8f86afe9b7a83bb158

PR #82, opened 2023-05-23
Error variants

uncomputable:
Updates error messages and variants.


uncomputable:
Did I miss anything that could be refactored? The type errors are a mess right now, but they will need their own PR.

apoelstra (approved):
ACK 76a2c683e5f7e2fde59950224e9d4d62b29bb470

apoelstra:
> Did I miss anything that could be refactored? The type errors are a mess right now, but they will need their own PR.

Probably, but that's fine :) let's just keep refactoring as we use the library and then we can zoom out and plan a better API once we have a bit of experience.

PR #88, opened 2023-05-29
add const word support

apoelstra:
This is a complete implementation though not the most efficient one; in particular the CMR calculation is a bit ugly. Probably we want to attach some more data to the new `Word` variant (e.g. its width and/or its TMR) to reduce the amount of redundant computation. But this will do for now and lets us move forward with testing against the C lib and working on other projects.


src/core/commit.rs
```diff
@@ -497,6 +502,19 @@ impl<J: Jet> CommitNode<J> {
         })
     }
 
+    /// Create a DAG that computes some black-box function that is associated with the given `jet`.
+    ///
+    /// _Overall type: A → B where jet: A → B_
```
uncomputable, line 507: This documentation is a copy-and-paste error.

src/decode.rs
```diff
@@ -170,7 +170,9 @@ fn decode_node<I: Iterator<Item = u8>, J: Jet>(
                     None => return Err(Error::EndOfStream),
                     Some(true) => Ok(CommitNode::jet(context, J::decode(bits)?)),
                     Some(false) => {
-                        return Err(Error::ParseError("we don't yet support const words"))
+                        let depth = decode_natural(bits, Some(32))?;
```
uncomputable, line 173: AFAIU, `decode_natural` [fails if the input is longer than 32 bit](https://github.com/BlockstreamResearch/rust-simplicity/blob/master/src/decode.rs#L330). Therefore `bound = Some(32)` and `bound = None` are equivalent. We should use the latter.

src/merkle/cmr.rs
```diff
@@ -54,11 +55,39 @@ impl CommitMerkleRoot for Cmr {
...
+        let imr_iv = Cmr::tag_iv(b"Simplicity-Draft\x1fIdentity");
+        let imr_pass1 = imr_iv.update_1(scribe.cmr());
+        // 2. Add TMRs to get the pass-two IMR
+        let types = crate::core::types::Type::powers_of_two_vec(15);
+        let imr_pass2 = imr_pass1.update(
+            types[0].tmr.0.into_inner().into(),
+            types[w].tmr.0.into_inner().into(),
+        );
+        // 3. Convert to a jet CMR
+        Cmr::tag_iv(b"Simplicity-Draft\x1fJet").update_1(imr_pass2)
```
uncomputable, line 88: That is a _big_ fixme. Do you have a link to this inline algorithm? We should open an issue for FFI bindings.

uncomputable:
The rest looks good 4a071ec

src/merkle/cmr.rs
apoelstra: Sure: https://github.com/BlockstreamResearch/simplicity/blob/master/C/dag.c#L80-L136

It's not too bad, I think we could basically translate the C code directly into Rust and it'd work. Though we'd still have to figure out how to cache the TMRs.

apoelstra: Part of why I did it this way was to double-check that Russell's inline algorithm was actually equivalent to the dumb "just take the IMR of a scribe" version. Committing this will mean we have a copy of that code which is easy to check against.

src/decode.rs
apoelstra: Sure.

apoelstra:
Addressed both nits. Added another `FIXME` to think about what to do when the user tries to create a constant word jet for a value which isn't a power-of-2 type. I think it's ok to open an issue and fix that in a later PR.

src/decode.rs
apoelstra: Lol, you had me going. We fail if the *number* is longer than 32 bits. But here we actually need to bound the number at 32. The fuzzer found that if I remove this `Some(32)` then it's possible to try to read a depth of 157 (or any crazy value) which triggers a panic down the line.

uncomputable: Oh. I will update the documentation of `decode_natural`. It is confusing what the bound is.

uncomputable:
It should be as easy as adding another error variant.

uncomputable (approved):
ACK 4a72f33

src/decode.rs
apoelstra: The offending program was `"5d5d325e3332"` ... let's try to remember to add a unit test with this!

PR #91, opened 2023-05-31
Rewrite type system to unify `Type` and `RcVar`

apoelstra:
This cleans up notation/terminology and eliminates a ton of duplicated code. Also makes some small steps toward const-ifying some of the functionality.

Unfortunately this rewrites all of the type inference logic and touches nearly the entire library.


src/types/mod.rs
```diff
@@ -0,0 +1,811 @@
...
+//!
+//! During type inference, types may also be *free*, which means that they may
+//! turn out to be any type. Additional constraints are provided by the program
+//! itself, and type inference proceeds by using the union-bound algorithm to
+//! unify all the constraints on each type.
+//!
+//! If unification succeeds, types may then be finalized, which sets all free
+//! types to unit and does one additional check (the "occurs check") which
+//! makes sure there are no infinitely-sized types, which are illegal in
+//! Simplicity.
```
uncomputable, line 35: This is confusing. I think the narrative should be that "types" may be variable and incomplete. Finalized types are types without free subtypes.

Types may be free, bound to some structure (unit, sum of two types, product of two types), or equal to some other type. If unification succeeds, types may be finalized. This sets all free types to unit and does one additional check (the "occurs check") ... .

src/types/mod.rs
```diff
@@ -0,0 +1,811 @@
...
+//!   * `Type`, which holds a mutex holding a `TypeInner`. This is the main
+//!     type representing a Simplicity type. This is the only type that users
+//!     are likely to encounter; the others are mainly only used during type
+//!     inference.
+//!   * `Constraint` defines whether a given type is free, bound to a
+//!     particular `Bound`, or equal to another variable.
+//!   * `Bound` defines the actual structure of a type: whether it is a
+//!     unit, sum or product.
+//!   * `Final`, which for a finalized type, provides a simple mutex-free
+//!     structure containing all of the type information and its TMR.
```
uncomputable, line 47: I would like to eradicate the word "variable" and replace it with "type". Something like this: 

- `Type` holds a mutex to a `TypeInner` which in turn contains a `Constraint`. This is the main representation of a Simplicity type, and the only struct that users are likely going to encounter. `Type`s may be variable and incomplete.
- `Constraint` defines whether a `Type` is free, bound to a particular `Bound` or equal to another `Type`.
- `Bound` defines the structure of a `Type`: whether it is a unit, or the sum or product of two `Types`.
- `Final` is a mutex-free structure that can be obtained from a finalized `Type` and includes all its information and its TMR.

src/types/mod.rs
```diff
@@ -0,0 +1,811 @@
...
+pub enum ConcreteBound {
+    /// The unit type
+    Unit,
+    /// A sum of two other types
+    Sum(Arc<Final>, Arc<Final>),
+    /// A product of two other types
+    Product(Arc<Final>, Arc<Final>),
+}
+
+/// Whether this type constraint indicates a finalized type
```
uncomputable, line 337: Rogue doc

src/types/mod.rs
```diff
@@ -0,0 +1,811 @@
...
+        /// traversing the type tree. If it has, it means we have a type containing
+        /// itself, which is illegal. This is referred to as the "occurs check".
+        occurs_check: bool,
+    },
+    /// Variable is equal to another variable, which is referred to as the "parent"
+    /// of this variable.
+    EqualTo { parent: Type },
+}
+
+/// An actual type
```
uncomputable, line 315: "The structure that a `Type` is bound to"

src/types/mod.rs
```diff
@@ -0,0 +1,811 @@
...
+}
+
+impl<'g> ops::Deref for ReadOnlyMutexGuard<'g> {
+    type Target = TypeInner;
+    fn deref(&self) -> &Self::Target {
+        &self.inner
+    }
+}
+
+/// A constraint on a (partially) inferred type
```
uncomputable, line 294: "Whether a `Type` is free, bound to some `Bound` or equal to another `Type`"

src/types/mod.rs
```diff
@@ -0,0 +1,811 @@
...
+            {
+                let mut x_lock = x_lock.into_mutable(); // ok since x is `EqualTo` but only `Bound` constraints are finalized
+                x_lock.constraint = Constraint::EqualTo {
+                    parent: grandparent.shallow_clone(),
+                };
+                drop(x_lock);
+                x = grandparent.shallow_clone();
+            } else {
+                // If there is no grandparent, return the parent.
+                drop(parent_lock);
```
uncomputable, line 423: Are these `drop`s necessary? We are not recursing. Maybe I am missing something

src/types/mod.rs
apoelstra: I like this narrative, but the problem is that we may have types with no free subtypes which are nontheless "incomplete". This happens if the subtypes get unified with something else so their free variables go away... but there is no way for the parent to be informed of this, so the parent continues to be "non-final" in the sense of having no cached TMR etc.

So I unfortunately need this awkward notion of "finalization" which is often basically a no-op, and which doesn't correspond simply to "has no free children".

apoelstra: Let me revisit this. I think maybe I can hide these details from the user and present a simple "finalized means no free children" model.

src/types/mod.rs
apoelstra: Yes, for borrowck reasons. (Try removing the `drop` line and you'll get a compiler error.) I will add a comment about this.

uncomputable: I guess you can't return an owned value or overwrite a value if there is a lock on it. I haven't worked with mutexes in a while

src/types/mod.rs
```diff
@@ -0,0 +1,811 @@
...
+        {
+            *concrete_data = Some(Arc::clone(&data));
+            Ok(data)
+        } else {
+            unreachable!("we just set the constraint to Bound and now it's not Bound")
+        }
+    }
+
+    /// Return a vector containing the types 2^(2^i) for i from 0 to n-1.
+    pub fn powers_of_2(n: usize) -> Vec<Self> {
```
uncomputable, line 703: `powers_of_two()` would be more consistent?

src/types/mod.rs
```diff
@@ -0,0 +1,811 @@
...
+        let mut ret = Vec::with_capacity(n);
+
+        let unit = Type::unit();
+        //ret.push(unit.shallow_clone());
```
uncomputable, line 707: Obsolete code?

uncomputable:
Checked that the new type inference matches the original one

src/core/context.rs
```diff
@@ -1,52 +1,65 @@
...
 use crate::jet::Jet;
+use crate::types;
 
 /// Context for constructing a Simplicity program
 #[allow(dead_code)]
 pub struct Context<J: Jet> {
-    pub(crate) naming: VariableFactory,
-    pow2: Vec<RcVar>,
+    pub(crate) naming: types::variable::Factory,
+    pow22: Vec<types::Type>,
```
uncomputable, line 11: Should be `pow2`. Remnant of when there used to exist `pow2: Vec<RcVar>`.

src/core/context.rs
```diff
@@ -1,52 +1,65 @@
...
+    unit_ty: types::Type,
```
uncomputable, line 12: Maybe add a comment that this is required so that the unit that the `pow2` types refer to is not dropped.

apoelstra: It's not required, it's purely an efficiency thing so that we don't construct a gazillion `Unit` types and compute its TMR over and over.

src/core/context.rs
apoelstra: I can rename to `pow2` (it was originally `pow22` to avoid conflicting with the existing `pow2` field, which later gets deleted). But I think `pow22` is actually a bit better, because these are types of the form `2^(2^n)`.

I will add some more comments to these types because this has caused me a lot of confusion.

src/types/arrow.rs
```diff
@@ -0,0 +1,288 @@
...
+//! This module defines the specific arrows associated with each kind of node.
+
+use std::fmt;
+use std::sync::Arc;
+
+use crate::core::Value;
+use crate::types::{Bound, Error, Final, Type};
+use crate::{jet::Jet, Context};
+
+/// A structure representing an expression's type, which maps a source type to a target type
```
uncomputable, line 29: "Type of an expression, which maps a source type to a target type"

This is slightly confusing for the user because `Type`s don't contain arrows. These kinds of types here are a generalization. We should explain this in the module docs.

src/types/arrow.rs
```diff
@@ -0,0 +1,288 @@
...
+    pub target: Type,
+}
+
+impl fmt::Display for Arrow {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{} → {}", self.source, self.target)
+    }
+}
+
+/// A structure representing an expression's type, which maps a source type to a target type
```
uncomputable, line 44: "Finalized type of an expression, which maps a source `Type` to a target `Type`"

src/lib.rs
```diff
@@ -58,6 +58,8 @@ use std::fmt;
 /// Error type for simplicity
 #[non_exhaustive]
 pub enum Error {
+    /// Type-checking error
+    TypeInference(crate::types::Error),
     /// A type cannot be unified with another type
```
uncomputable, line 63: These can be removed

src/lib.rs
```diff
@@ -58,6 +58,8 @@ use std::fmt;
 /// Error type for simplicity
 #[non_exhaustive]
 pub enum Error {
+    /// Type-checking error
+    TypeInference(crate::types::Error),
```
uncomputable, line 62: I would call it simply `Type`

src/lib.rs
```diff
@@ -99,6 +101,7 @@ pub enum Error {
 impl fmt::Debug for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
+            Error::TypeInference(ref e) => write!(f, "typeck: {}", e),
```
uncomputable, line 104: Maybe just `fmt::Display::fmt(e, f)` for a general type error

uncomputable:
Checked that the arrow code matches the original type inference. Looks good

uncomputable:
I made it until ac11bdc. The rest is for tomorrow

apoelstra:
Force-pushed to add two new `f`ixup commits which change the comments and some names in types.rs and arrow.rs.

Your bookmark commit is now called 2483fbbcf4b4a13717acfd6b5260e716c0c7e58a.

Will hold off on any other changes until after your review is complete, since I don't think we'll need to iterate much on them, and I don't want to step on your toes.

src/core/commit.rs
```diff
@@ -220,28 +211,6 @@ impl<J: Jet> CommitNode<J> {
...
-                inner,
-                arrow,
-            })),
-            Err(Error::Unification(unification_hint)) => Err(Error::TypeCheck {
-                unification_hint,
-                root_hint: hint,
-            }),
-            Err(error) => Err(error),
-        }
-    }
```
uncomputable, line 243: We remove this because the hints from the type inference already explain the error, plus we want to third parties like your decoder to display helpful error messages to the user?

src/types/arrow.rs
```diff
@@ -60,9 +66,11 @@ impl Arrow {
...
     /// Create a unification arrow for a fresh `iden` combinator
     pub fn for_iden<J: Jet>(context: &mut Context<J>) -> Self {
-        // We cheat and make the source and target of iden literally be the same
-        // variable, rather than unifying them. This is slightly more efficient
-        // and shouldn't result in any user-visible consequences.
+        // Throughout this module, when two types are the same, we reuse a
+        // pointer to them rather than creating distinct types and unifying
+        // them. This theoretically could lead to more confusing errors for
+        // the user during type inference, but in practice type inference
+        // is completely opaque and there's no harm in making it moreso.
```
uncomputable, line 88: We have to experiment and see if this becomes an issue for users

src/core/context.rs
uncomputable: Ok, that makes sense. To avoid future confusion, let's add a comment

uncomputable (changes requested):
Reviewed 3f9884a. Looks good, including the documentation. There are some potential sources of confusion, such as the difference between "type" and `Type`, but those can be addressed in a future PR. Let's work on the remaining comments and squash.

src/core/commit.rs
apoelstra: Correct. Adding a second fixed hint provides pretty little value in my opinion.

What we may do in the future is change the `hint` field to an owned `String` or even a closure that can take extra auxiliary information.... but for now my decoder is able to get a fair bit of mileage out of the single fixed hint and a line/column.

Anyway improving the errors is for a later PR.

src/types/arrow.rs
apoelstra: Agreed. But I'm pretty confident it won't be an issue ... the internal state of type inference is pretty useless no matter how much we try to avoid shortcuts.

src/types/mod.rs
uncomputable: I like your trichotomy of incomplete types, complete types and finalized types. Types may be incomplete. Complete types have no free variables. Finalized types have cached information.

uncomputable:
Squash and we are ready to merge 60b7686.

PR #93, opened 2023-06-02
Followups to #91

apoelstra:
Cleans up the merkle types and adds a bit more functionality; re-exports some types from the crate root; updates the type inference block comment.


src/types/mod.rs
```diff
@@ -29,20 +29,27 @@
...
+//! free types to the unit type.
 //!
 //! In addition to requiring that type inference succeeds (i.e. no constraints
 //! are in conflict with each other), there is one additional check, the "occurs
 //! check", which ensures that there are no infinitely-sized types. Such types
-//! occur when a type has itself as a child, and are illegal in Simplicity.
+//! occur when a type has itself as a child, and are illegal in Simplicity. This
+//! check is explicitly done by `finalize`, to avoid infinitely looping, but any
+//! final type is guaranteed to pass the occurs-check, whether it was finalized
+//! explicitly or during unification.
```
uncomputable, line 52: Are final types ever constructed "explicitly"? What you write makes sense, but I still find it confusing.

"Any finalized type passes the occurs-check."

uncomputable:
ACK 0fa3017. Feel free to address my nit or to leave it be.

src/types/mod.rs
apoelstra: By "explicitly" I mean, by calling `.finalize()`.

I reworded this a dozen times and never quite became happy with it. I'm open to suggestions.

PR #95, opened 2023-06-02
Word compiler

uncomputable:
Replaces scribe with word jets in the policy compiler.


apoelstra (approved):
ACK 151796a30a107c975627565f3fceae1e1a3e6c26

PR #97, opened 2023-06-02
decode: fix issue when program is a single witness node

apoelstra:
Also add a detailed unit test that our witness de-sharing works, because currently there are no tests that detect it. (The `fresh_witness` flag is used in exactly one place, and you can comment it out without anything breaking.)


uncomputable (approved):
ACK 860a3ff.

PR #98, opened 2023-06-03
imr: fix IMR computation for constant words

apoelstra:
The IMR of a constant word is the same as its CMR; our code was using the CMR *initialization vector*, since for other "same as CMR" cases (jets, hidden, etc) means "same as CMR IV". For words, not so.


uncomputable (approved):
ACK acf07c35bf966fcf9c26f359c2d8dfdb4fb1ae9f

PR #99, opened 2023-06-03
Move witness unsharing from deserialization to finalization

apoelstra:
I think it makes more logical sense for witness nodes to be unshared during finalization (i.e. when witnesses are being attached to the program) rather than during deserialization (when unsharing is not always appropriate, leading to us having two `decode_program` methods, one which does unsharing and one which doesn't).

Better to deserialize a program without modifying it, and then only when attaching witnesses, make the appropriate decisions about how to do so. It is also dramatically simpler to do unsharing during finalization than it is to do it during deserialization.


apoelstra:
Will leave this as a draft for a bit; there are other bugs I want to fix first.

apoelstra:
This now builds on #98 and #100, and should be good to go once those two are in.

apoelstra:
After this PR, and #104, I will change this so that `RedeemNode::finalize` has a choice of two iterators (a fully-shared one used when deserializing consensus programs, and a fully-unshared one used when constructing witnesses).

At that point there won't be any user-visible notion of "unsharing" at all.

uncomputable:
I will review this tomorrow. Looks good at first glance and short.

apoelstra:
Rebased.

uncomputable:
One inconsistency is that `RedeemNode::decode()` enforces `1 → 1` but `CommitNode::finalize()` doesn't (see #105). We should address this in a follow-up PR.

uncomputable:
ACK 226c2a8b240fe6fa9b18e525b9c65877dd3ab172.

uncomputable (approved):

PR #100, opened 2023-06-03
encode: share witness data when it is repeated

apoelstra:
There are three bugs in our `RedeemNode` encoding logic:
* We forgot to flush the `BitWriter` after writing, meaning that sometimes trailing bits would not get written.
* Flushing a `BitWriter` always writes a byte to the underlying writer even if there is nothing to write.
* We used sharing logic when encoding program nodes but not witnesses. In the case of repeated witnesses, this means that we would sometimes encode witnesses that we were not supposed to.

This fixes all three bugs and adds unit tests that both encoding and decoding work correctly.

### DIFF
--- a/src/bitwriter.rs
+++ b/src/bitwriter.rs
@@ -84,9 +84,11 @@ impl<W: io::Write> BitWriter<W> {
     /// Write out all cached bits.
     /// This may write up to two bytes and flushes the underlying [`io::Write`].
     pub fn flush_all(&mut self) -> io::Result<()> {
-        self.w.write_all(&[self.cache])?;
-        self.cache_len = 0;
-        self.cache = 0;
+        if self.cache_len > 0 {
+            self.w.write_all(&[self.cache])?;
+            self.cache_len = 0;
+            self.cache = 0;
+        }
 
         io::Write::flush(&mut self.w)
     }
--- a/src/core/iter.rs
+++ b/src/core/iter.rs
@@ -189,9 +189,17 @@ pub fn into_witness<'a, J: Jet + 'a, I>(iter: I) -> impl Iterator<Item = &'a Val
 where
     I: Iterator<Item = RefWrapper<'a, J>> + Clone,
 {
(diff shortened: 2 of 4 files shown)


uncomputable:
We should also fix `CommitNode::encode()`

apoelstra:
> We should also fix `CommitNode::encode()`

Sure, you mean, to add the `flush_all` call?

apoelstra:
Fixed and pushed. I didn't add a test for this because it seemed like a real chore to come up with a program of a particular offending length, and because I think we'll later refactor/replace BitIter/BitWriter to be safer to use and not require manual flushing like this.

uncomputable:
Come to think of it, if `CommitNode::encode()` uses `RedeemNode::encode()`, then we might not need to flush. False alarm?

apoelstra:
> Come to think of it, if `CommitNode::encode()` uses `RedeemNode::encode()`, then we might not need to flush. False alarm?

It doesn't -- if anything it would be the other way around. But actually it looks like neither calls the other, they both call `encode_program` (and I'm pretty sure we **don't** want to flush in `encode_program` since we don't want to flush in the case that we still have witness data to add).

src/core/commit.rs
```diff
@@ -771,7 +771,9 @@ impl<J: Jet> CommitNode<J> {
     pub fn encode<W: io::Write>(&self, w: &mut BitWriter<W>) -> io::Result<usize> {
         let empty_witness = std::iter::repeat(Value::Unit);
         let program = self.finalize(empty_witness).expect("finalize");
-        program.encode(w)
+        let ret = program.encode(w)?;
```
uncomputable, line 774: What about this line? Sorry for all the confusion

apoelstra: Derp, you are correct. It was a false alarm. I removed the new commit.

uncomputable (approved):
ACK 478182bc846f0cb241501e3502c345f92e0a174a

PR #101, opened 2023-06-04
Fix jets-bench crate and add it to CI

apoelstra:
This crate was broken by API changes in the rest of the repo, but also I'm not sure that the committed version *ever* compiled since there was a variable defined in an `unsafe` block but referred to outside of it.

I have not actually run the benchmarks since they will take a very long time and need exclusive overnight access to my server.


apoelstra:
Rebased on master.

jets-bench/benches/elements/main.rs
```diff
@@ -125,14 +92,28 @@ impl ElementsBenchEnvType {
     }
 }
 
+fn jet_arrow(jet: Elements, pows_of_two: &[Type]) -> (Arc<types::Final>, Arc<types::Final>) {
+    let src_ty = jet
+        .source_ty()
+        .to_type(|n| pows_of_two[n].shallow_clone())
```
sanket1729, line 98: nit: This can use the `to_bit_width` from #110. But this is only used in benchmarks, not necessary to change this.

sanket1729 (approved):
ACK e49479cef467729f9505d84b360b5ca6556de478

uncomputable (approved):
utACK e49479cef467729f9505d84b360b5ca6556de478

PR #103, opened 2023-06-04
sharing: use IMR rather than AMR for sharing identity

apoelstra:
Also greatly simplify the sharing algorithm, which can be made much simpler thanks to the IMR.

I think there are still some issues with the IMR calculation but we can address that in a later PR with more tests.

### DIFF
--- a/src/encode.rs
+++ b/src/encode.rs
@@ -25,6 +25,7 @@ use crate::core::redeem::{RedeemNodeInner, RefWrapper};
 use crate::core::Value;
 use crate::jet::Jet;
 use crate::sharing;
+use crate::Imr;
 use std::collections::HashMap;
 use std::{io, mem};
 
@@ -42,7 +43,7 @@ pub fn encode_program<W: io::Write, J: Jet>(
 
     let mut index = 0;
     for node in program {
-        if node_to_index.get(&node).unwrap() != &index {
+        if node_to_index.get(&node.0.imr).unwrap() != &index {
             continue;
         }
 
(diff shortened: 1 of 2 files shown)


uncomputable:
@roconnor-blockstream There are cases where the IMR of two nodes is the same and the source / target types are different, right? This is why types must be checked during the sharing computation.

apoelstra:
@uncomputable all non-hidden nodes have their source and target types hashed into the IMR. So it's only hidden nodes where we might want to worry about the types differing.

It can happen that you can have two hidden nodes with different types but same hash (and therefore same IMR), though ofc this would indicate that something weird is going on. But my read of the consensus C code is that this is forbidden -- "sharing" means exactly "all IMRs are unique". See https://github.com/BlockstreamResearch/simplicity/blob/7be9a8725135ea81f6454968af574a6ca2390407/C/dag.c#L551-L572

But yes, let's get some clarification from @roconnor-blockstream before merging this.

uncomputable:
I never understood why the IMR computation has a second pass with types hashed into it. That is the source of my confusion.

Hidden nodes are not a problem because they are never executed. The only reason we care about types is that `iden` nodes have to copy a certain number of bits, etc.

apoelstra:
So, from @roconnor-blockstream on IRC:

* `Hidden` nodes do not actually have types so their IMRs are equal to their CMRs (with a different IV)
* Because they have no types, the types that our code assigns to them does not affect their CMRs. *Any hidden nodes with the same content must be shared, regardless of type concerns.*
* Possibly we should drop `Hidden` nodes entirely ... in consensus they are only valid as part of an `AssertL`/`AssertR`, and it's hard for me to think what other "in-construction" use we might have for them.

So anyway this PR is fine as-is, though there is a bug in our IMR computation for `Hidden` ... and I think for `Jet` as well, and possibly others. But we can address that separately.

uncomputable (approved):
ACK c72d46d330e240239006fb60b98b831202e1a8b5

PR #104, opened 2023-06-04
Redo iteration logic

apoelstra:
Draft since we can't completely replace the existing iteration logic without messing with both sharing (which would conflict with #103) and finalization (which would conflict with #99). But this is mostly complete.

Basically, I was able to encapsulate all the weird pointer-comparison stuff into a new `dag` module which exposes a nice API that lets you iterate over commitnodes or redeemnodes, regardless of whether you want normal references or `Rc`s. As a user you never need to use types like `RefWrapper`.

### DIFF
--- a/src/core/commit.rs
+++ b/src/core/commit.rs
@@ -15,10 +15,10 @@
 
 use crate::bititer::BitIter;
 use crate::bitwriter::BitWriter;
-use crate::core::iter::{DagIterable, PostOrderIter, WitnessIterator};
+use crate::core::iter::{DagIterable, WitnessIterator};
 use crate::core::redeem::RedeemNodeInner;
 use crate::core::{Context, RedeemNode, Value};
-use crate::dag::DagLike;
+use crate::dag::{DagLike, PostOrderIter};
 use crate::jet::Jet;
 use crate::merkle::amr::Amr;
 use crate::merkle::cmr::Cmr;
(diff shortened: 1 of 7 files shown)


apoelstra:
After talking to @roconnor-blockstream on IRC, I am going to follow this up with a PR that gets replaces this iterator (which exposes the internal DAG representation) with one ore more of
* An iterator which yields every node in the entire unshared tree
* An iterator (for `RedeemNode`) which yields nodes with full sharing enforced (for use in bit-serialization)
* An iterator (for both `CommitNode` and `RedeemNode`) which yields nodes with full sharing **except** that disconnect and witness nodes are never shared (which may be used for human-readable serialization).

But I still want to get this PR in first, because it pays down a bunch of tech debt related to iteration.

uncomputable:
Hiding our internal sharing makes sense. Do we care about DDoS for the unshared iterator? The worst case is an unshared program of exponential size, right?

uncomputable:
85664961e28207a5e870407dcc676ea45bb293a3: This commit does like three things: Adding `PointerId`, copying `PostOrderIter` and adding the deduped-witness iterator. Split?

apoelstra:
@uncomputable yes, it is (in general) exponential. It may be that we want to check the static bounds against consensus limits during finalization. Or maybe we should require the user provide an upper limit on resources before starting iteration. Worth filing an issue about.

But I think it's fine for now to have the iterator, and we can think about bounding it later.

uncomputable:
Using `RefWrapper` is a canonical solution for adding trait implementations like `Eq`, but your solution is cleaner for our project. The fewer types users have to deal with the better.

uncomputable (approved):
ACK 8664da82fd8abef689451e592d99a94b2715b729

PR #110, opened 2023-06-06
Replace (non-jet) FFI bindings

apoelstra:
The jet FFI bindings are all autogenerated (and actively used) so I am fairly comfortable with them. However, we had a number of bindings to other parts of the C library, which got no use at all, and which were in fact wildly out of date and would've segfaulted had they ever been used. The type bindings were also inconsistently named and located, which would make auditing them and maintaining them quite difficult.

There was the start of some code to test the Schnorr program (and a couple others) from the C library, but these were also out-of-date, having arrays of wrong length and missing the new `ctxPruned` and `ctxUnpruned` programs.

This PR:
* Deletes the old FFI code entirely, since there was not very much of it
* Introduces a new simplicity-sys/src/ffi.rs file with updated FFI bindings; **all FFI bindings are in this file and this file alone** so they'll be a bit easier to maintain.
* Re-implements the `testProgram` function from the C source code, which exercises the bindings by parsing a program and its witnesses, typechecking it, and executing it.
* Adds unit tests which exercise the C test programs, asserting that everything is of the correct length and works correctly.

I have a followup PR which uses the new `test_program` unit test function to fuzz the C and Rust libraries against each other, which I may open as a draft, but it is currently blocked on an upstream bug https://github.com/BlockstreamResearch/simplicity/issues/151, on some bugs in this library's AMR implementation, and on us updating to a new version of bitcoin-hashes which doesn't ruin all the hashes when `cfg(fuzzing)` is set.


simplicity-sys/src/ffi.rs
```diff
@@ -0,0 +1,478 @@
...
+
+/// Used in the C code for various purposes; actually defined as `uint_least32_t`
+/// which seems not to have an equivalent in Rust libc.
+#[allow(non_camel_case_types)]
+pub type ubounded = u32;
+
+/// Used in C code to give the project a Win32 vibe; actually defined as
+/// `uint_least16_t`.
+#[allow(non_camel_case_types)]
+pub type UWORD = u16;
```
sanket1729, line 37: Same here. This should be usize.

sanket1729 (commented):
Some of the work here is repeated in `c_jets/exec_ffi.rs`. 
Will do another round of review later.

simplicity-sys/src/ffi.rs
```diff
@@ -0,0 +1,478 @@
...
+
+use std::fmt;
+
+/// Used with `evalTCOProgram` to enforce consensus limits.
+pub const BUDGET_MAX: ubounded = 4000050;
+
+/// Used in the C code for various purposes; actually defined as `uint_least32_t`
+/// which seems not to have an equivalent in Rust libc.
+#[allow(non_camel_case_types)]
+pub type ubounded = u32;
```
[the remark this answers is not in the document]
apoelstra: Weird, thanks!

simplicity-sys/src/tests.rs
```diff
@@ -0,0 +1,238 @@
...
+        pub static ctx8Pruned_imr: [u32; 8];
+
+        pub static sizeof_ctx8Unpruned: size_t;
+        pub static ctx8Unpruned: [u8; 4809];
+        pub static ctx8Unpruned_amr: [u32; 8];
+        pub static ctx8Unpruned_cmr: [u32; 8];
+        pub static ctx8Unpruned_imr: [u32; 8];
+
+        pub static sizeof_schnorr0: size_t;
+        pub static schnorr0: [u8; 137];
```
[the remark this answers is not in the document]
apoelstra: Oh, neat. Ok, I'll change these to 0s, and drop the `assert`s on the length.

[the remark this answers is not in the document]
apoelstra: I cannot do this; I take these arrays, convert them to slices, then convert those to bitstreams (which uses the slice length). If I set the array length to 0 then I will get 0-length bitstreams and the tests fail.

I could use some unsafe code to read the `sizeof_schnorr0` variable and force the slice length...but it seems much nicer to just hardcode the value and deal with the assertion failures whenever they change.

simplicity-sys/src/ffi.rs
```diff
@@ -0,0 +1,480 @@
...
+use std::fmt;
+
+/// Used with `evalTCOProgram` to enforce consensus limits.
+pub const BUDGET_MAX: ubounded = 4000050;
+
+/// Used in the C code for various purposes; actually defined as `uint_least32_t`
+/// which seems not to have an equivalent in Rust libc.
+// There are sanity checks for both these types in c_jets/frame_ffi.rs
+#[allow(non_camel_case_types)]
+pub type ubounded = usize;
```
sanket1729, line 33: This is correct as `u32`. I only test the `UWORD` as `usize`. Apparently, `uint_least32_t` maps to u32, while `uint_fast32_t` maps to `usize`. 

The tests are locally failing when I do this. This also means that we are not correctly running this in CI. Can you also add the check on `ubounded` in `c_jets::c_env::tests::test_sizes` ?

simplicity-sys/src/ffi.rs
```diff
@@ -0,0 +1,480 @@
...
+        pub fn computeAnnotatedMerkleRoot(
+            analyses: *mut CAnalyses,
+            dag: *const CDagNode,
+            ty: *const CType,
+            len: size_t,
+        );
+
+        /// Verifies that the 'dag' is in canonical order, meaning that nodes
+        /// under the left branches have lower indices than nodes under
+        pub fn verifyCanonicalOrder(dag: *const CDagNode, len: size_t) -> SimplicityErr;
```
sanket1729, line 305: The C comment says that `dag` here might change in some cases and does not mark it as const pointer. We should also make this mut *

simplicity-sys/src/ffi.rs
```diff
@@ -0,0 +1,480 @@
...
+    extern "C" {
+        /// Run the Bit Machine on the well-typed Simplicity expression 'dag[len]'.
+        pub fn evalTCOExpression(
+            anti_dos_checks: c_uchar,
+            output: *mut UWORD,
+            outputSize: ubounded,
+            input: *const UWORD,
+            inputSize: ubounded,
+            dag: *const CDagNode,
+            type_dag: *const CType,
```
sanket1729, line 367: `type_dag` should be *const. For some reason which I did not investigate, it is called into writeValue function that is indirectly called by this.

simplicity-sys/src/ffi.rs
```diff
@@ -0,0 +1,480 @@
...
+            len: size_t,
+            budget: ubounded,
+            env: *const c_void,
```
sanket1729, line 370: This should be `CElementsTxEnv`?

simplicity-sys/src/ffi.rs
```diff
@@ -0,0 +1,480 @@
...
+    ///
+    /// This function directly wraps `evalTCOExpression`; see the documentation for
+    /// that function in the C source code for preconditions.
+    #[allow(non_snake_case)]
+    pub unsafe fn evalTCOProgram(
+        dag: *const CDagNode,
+        type_dag: *const CType,
+        len: size_t,
+        budget: ubounded,
+        env: *const c_void,
```
sanket1729, line 388: Same here. This can be `CElementsTxEnv`

simplicity-sys/src/ffi.rs
```diff
@@ -0,0 +1,480 @@
...
+        pub injr_cnt: size_t,
+        pub take_cnt: size_t,
+        pub drop_cnt: size_t,
+    }
+
+    /// A source and target type (in the C code this is actually an anonymous struct
+    /// where it is used).
+    #[repr(C)]
+    #[derive(Copy, Clone, Eq, PartialEq, Debug)]
+    pub struct CArrow {
```
sanket1729, line 225: This is unused in FFI. But okay to keep here.

simplicity-sys/src/ffi.rs
```diff
@@ -0,0 +1,480 @@
...
+        /// Given a well-formed 'type_dag', compute the bitSizes, skips, and type Merkle roots of all subexpressions.
+        pub fn computeTypeAnalyses(type_dag: *mut CType, len: size_t);
+
+        /// If the Simplicity DAG, 'dag', has a principal type (including constraints
+        /// due to sharing of subexpressions), then allocate a well-formed type DAG
+        /// containing all the types needed for all the subexpressions of 'dag'.
+        ///
+        /// This is actually from typeInference.h but we stick it in the module for
+        /// types.h for simplicity.
+        pub fn mallocTypeInference(
```
sanket1729, line 473: nit: This in is file `typeinfernce.h`

sanket1729 (commented):
checkpoint 9c2c10e23b90111f5c9a76e7534c222a919c9879. Sorry, I think I incorrectly pointed `ubounded` to be `usize`. Finally, the tests fail locally without `ubounded` being `u32`. Meaning that this is not checked in CI

simplicity-sys/src/ffi.rs
apoelstra: The tests were passing because
* We were directly testing `usize` instead of the `UWORD` type alias, so the test would've passed no matter what I'd put for `UWORD`
* We weren't testing `ubounded` at all.

When I address both those issues then I see test failures, and I have to correctly set the types for the tests to pass. So I think we're ok.

apoelstra: I guess we should also add tests for all the new structures. I'll add that in a separate commit.

simplicity-sys/src/ffi.rs
apoelstra: It's actually not used at all. I will remove it.

sanket1729 (approved):
ACK f6cfe65655818153c80bedaf8132f56fefb3fd8c

PR #111, opened 2023-06-07
cmr: replace word CMR algorithm with inline one

### LINKED ISSUE #90: More efficient word jet CMR
The CMR of a word jet is the same as the IMR of the equivalent scribe expression.

Currently we literally construct the scribe tree and compute the IMR. There exists an [inline algorithm](https://github.com/BlockstreamResearch/simplicity/blob/a365c82347a61b5300a9b02d6f6adb4b7bccaeae/C/dag.c#L80-L136) which is more efficient.

- [x] Check if naive construction and inline algorithm are equivalent
- [x] Replace naive construction with inline algorithm

apoelstra:
This ports the CMR algorithm from the C code into Rust. It does indeed make the unit tests run 10x faster.

Also gets rid of some type inefficiency converting jet names to types. There is now no more use of `Type::powers_of_two` in the main crate.

Fixes #90


src/merkle/tmr.rs
```diff
@@ -55,6 +55,101 @@ impl Tmr {
         0x06, 0xb2, 0x2b, 0xbe, 0x08, 0x8b, 0xb1, 0xaf,
     ]));
 
+    /// The TMRs of the types TWO^(2^n) for small values of n
```
sanket1729, line 58: In the future, we can also cross this with by creating bindings with these values as computed in `precomputed.h`.

src/jet/mod.rs
```diff
@@ -124,18 +123,11 @@ pub trait Jet: Copy + Eq + Ord + Hash + std::fmt::Debug + std::fmt::Display {
...
-            .expect_finalized();
-        let target_ty = self
-            .target_ty()
-            .to_type(|n| pows_of_two[n].shallow_clone())
-            .expect_finalized();
-
-        let a_frame_size = round_u_word(src_ty.bit_width());
-        let b_frame_size = round_u_word(target_ty.bit_width());
+        let src_ty_bit_width = self.source_ty().to_bit_width();
+        let target_ty_bit_width = self.target_ty().to_bit_width();
```
sanket1729, line 127: nit: In future, we should also cache the value of bit width. But this would require significant refactors ranging from the autogenerated code.

sanket1729 (approved):
ACK d20bf7743d2bd1f60fc03fb8e987b55bc7562c3d

src/merkle/tmr.rs
```diff
@@ -87,4 +182,13 @@ mod tests {
             Tmr::PRODUCT_IV,
         );
     }
+
+    #[test]
+    fn const_powers_of_2() {
+        let n = Tmr::POWERS_OF_TWO.len();
+        let types = crate::types::Type::powers_of_two(n);
```
uncomputable, line 189: Let's move `Types::powers_of_two` into this test module. It is API bloat at this point.

uncomputable:
Checked that `master` and this PR compute the same IMR.

```
diff --git a/src/policy/compiler.rs b/src/policy/compiler.rs
index 2139f11..ad8a40c 100644
--- a/src/policy/compiler.rs
+++ b/src/policy/compiler.rs
@@ -275,6 +275,9 @@ mod tests {
         let finalized = commit
             .finalize(witness.into_iter(), true)
             .expect("finalize");
+
+        dbg!(finalized.imr);
+
         let mut mac = BitMachine::for_program(&finalized);
 
         let success = match mac.exec(&finalized, env) {
```

src/merkle/tmr.rs
apoelstra: > Let's move `Types::powers_of_two` into this test module. It is API bloat at this point.

I'm unsure if it's purely bloat. It's convenient in a pinch (though I guess it's also very easy to reimplement, and so far every time I've used it I've eventually found a better solution..).

apoelstra:
I'm gonna go ahead and merge. I think all the nits fall under a general "we could be precomputing more and handling precomputed types more efficiently".

PR #114, opened 2023-06-08
dag: extend iterators to handle witness unsharing; remove last uses of `RefWrapper`

### LINKED ISSUE #113: Remove RefWrapper entirely
The last remaining use is in `CommitNode::finalize()`. Finishes what #104 started.

apoelstra:
This is the second part of #104. It extends the iterator abstraction to track indices and child indices, meaning that the iterators can be used directly for sharing checking without any additional hashmaps. It also adds the ability to unshare witnesses and disconnect combinators (and their parents, which the existing code does not do), and uses this during `CommitNode::finalize` to correctly unshare witnesses in a somewhat elegant way.

Fixes #113


apoelstra:
Converting to draft because there's a slightly simpler variant of `CommitNode::finalize` I want to try. But for the most part this is good to go.

src/dag.rs
```diff
@@ -263,47 +263,186 @@ impl<J: jet::Jet> DagLike for Rc<RedeemNode<J>> {
...
 pub struct PostOrderIter<D: DagLike> {
-    stack: Vec<D>,
-    maybe_current: Option<D>,
-    visited: HashSet<PointerId>,
+    /// The index of the next item to be yielded
+    index: usize,
+    /// A stack of elements to be yielded; each element is a node, then its left
+    /// and right children (if they exist and if they have been yielded already)
+    stack: Vec<IterStackItem<D>>,
+    visited: HashMap<PointerId, usize>,
```
uncomputable: AFAIU, this maps a memory address to its shared index. Please add documentation.

src/dag.rs
```diff
@@ -263,47 +263,186 @@ impl<J: jet::Jet> DagLike for Rc<RedeemNode<J>> {
...
+                        Previous::ParentLeft => {
+                            assert!(self.stack[stack_len - 1].processed);
+                            self.stack[stack_len - 1].left_idx = Some(self.index);
+                        }
+                        Previous::ParentRight => {
+                            assert!(self.stack[stack_len - 1].processed);
+                            self.stack[stack_len - 1].right_idx = Some(self.index);
+                        }
+                        Previous::SiblingLeft => {
+                            assert!(self.stack[stack_len - 2].processed);
```
uncomputable, line 606: I know we have a lot of `assert!`s, but what is our current policy? Should `assert!` become `debug_assert!`?

uncomputable:
08016f55cdec3b6c7d4a9108e9f4c83179f1f9e4 could have been split into smaller commits :)

src/dag.rs
```diff
@@ -263,47 +263,186 @@ impl<J: jet::Jet> DagLike for Rc<RedeemNode<J>> {
...
+                        current.left_child(&self.visited),
+                        current.right_child(&self.visited),
+                    ) {
+                        // No children is easy, just mark it processed and iterate
+                        (Child::None, Child::None) => {
+                            self.stack.push(current);
+                        }
+                        // If there is no left child, the only possibility is no
+                        // right child either.
+                        (Child::None, _) => unreachable!(),
```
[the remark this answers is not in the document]
apoelstra: Oh, that's a good idea.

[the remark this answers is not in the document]
apoelstra: In future can you indicate what commit you're reviewing?

apoelstra:
Improved finalization algorithm (it is indeed simpler and has a smaller diff now), and split the "introduce sharing and unsharing iterators" commit into two, one which introduces the iterators and one which deletes the sharing module. Also addressed comment nits on the first commit.

Ready for review.

apoelstra:
Fixed bug that fuzzer found (forgot to finish the witness iterator in `finalize()`); added unit test.

src/core/redeem.rs
```diff
@@ -171,7 +171,14 @@ impl<J: Jet> RedeemNode<J> {
         let witness = WitnessDecoder::new(bits)?;
         let program = commit.finalize(witness, false)?;
 
-        if sharing::check_maximal_sharing((*program).post_order_iter::<FullSharing>()) {
+        let iter = (&*program)
+            .post_order_iter::<InternalSharing>()
+            .map(|node| node.imr);
+        let fully_shared_iter = (&*program)
```
uncomputable: `&* program` can be `program.as_ref()`

uncomputable:
> The old logic did this in an ad-hoc way and didn't actually work -- if a
witness was part of a larger tree that was shared, we wouldn't detect
this and unshare the tree.

(from 2ed17c3bd3bf2cb5b00584cdd9ffb67e5d399523)

We should add a unit test.

src/core/redeem.rs
apoelstra: Pushed fix for this (and one other instance in decode.rs where I did the same thing).

Will add a sharing unit test in a fresh commit.

apoelstra:
Added a commit with a unit test for witness-unsharing when witness nodes are in a shared subexpression. You can cherry-pick this onto master and confirm that it fails there.

src/core/commit.rs
```diff
@@ -601,142 +601,66 @@ impl<J: Jet> CommitNode<J> {
...
-                Some(v)
+        // Iterate over either the internally-shared DAG, or the internally-shared DAG
+        // with witnesses unshared. The borrowck makes this a little irritating to express.
+        let mut share = InternalSharing::default();
+        let mut unshare = dag::UnshareWitnessDisconnect::new(share.clone());
+        let tracker: &mut dyn dag::SharingTracker<&CommitNode<J>> = if unshare_witnesses {
+            &mut unshare
+        } else {
+            &mut share
+        };
```
uncomputable, line 612: If the trait object is annoying, we could write a private method `finalize_helper<S>(&self, mut witness: W)` with `S` being the sharing tracker type, and wrap this in the public `finalize()` method.

uncomputable (approved):
ACK 526d66aee6571d8bc509a55a21ff3ac635a3dc8f. I love how we internalized the sharing logic into the iterators.

src/core/commit.rs
apoelstra: Ah, that's not too bad an idea. Let's keep it in mind the next time we're messing with this code.

PR #118, opened 2023-06-10
decode: change algorithm to avoid using CommitNode::Hidden; other cleanups

apoelstra:
This is the first step of removing `Hidden` nodes.

The next steps are:
1. Refactor: introduce an `encoding` module, pull `encode`, `decode`, `bitwriter` and `bititer` into that.
2. Update `encode_node` to also handle hidden nodes specially.
3. Change `AssertL` and `AssertR` to directly hold CMRs rather than `Hidden` nodes
4. Delete `Hidden` nodes from everywhere except for `DagNode` and `DecodeNode`.


apoelstra:
Initially I had thought that the refactor was necessary because I wanted `DecodeNode` to be private to the `bit_encoding` module but still be usable by both `decode` and `encode`. Now that I've implemented the encoding half, I realize that this isn't actually necessary.

I might do the refactoring anyway because I think it leaves the crate in a better state, but I could go either way. What do you think?

apoelstra:
Tacked a couple more commits on which just add tests (and fix the witness AMR bug since it was interfering with my tests). I can move those to a separate PR if you'd like.

uncomputable:
`InconsistentWitnessLength` should be part of `crate::decode::Error`.

We could add an associated `Error` type to `WitnessIterator` or rewrite the trait entirely. It's hard to reconcile an iterator that never fails (`Iterator<Item = Value>`) with one that may (`WitnessDecoder`).

src/decode.rs
```diff
@@ -377,27 +377,63 @@ mod tests {
     use crate::bitwriter::BitWriter;
     use crate::encode;
     use crate::exec::BitMachine;
+    use crate::jet::Core;
+    use bitcoin_hashes::hex::ToHex;
+
+    fn assert_program_succeeds<J: Jet>(prog_bytes: &[u8]) -> Rc<CommitNode<J>> {
```
uncomputable: `assert_program_serialization_succeeds` or `assert_program_serializable` might be better identifiers.

apoelstra:
> `InconsistentWitnessLength` should be part of `crate::decode::Error`.

Let's file an issue and revisit this.

Right now we're using the same trait to provide a stream of witnesses from the bit-serialization, and to provide a "stream" of witnesses from the satisfier. I think probably these should be separate traits somehow; then we can move the existing `WitnessIterator` into the `decode` module and drop its `Iterator<Item = Value>` impl entirely.

But I'm not sure exactly how to handle witnesses from the satisfier so I think let's avoid dealing with this for now.

uncomputable:
Checkpoint 1be4f2aa146efb304288d31ce8b428f62761551c. Rest comes tomorrow.

src/decode.rs
```diff
@@ -135,51 +182,94 @@ pub fn decode_program_arbitrary_type<I: Iterator<Item = u8>, J: Jet>(
...
-        ) {
-            if let Some(left_index) = data.left_index {
-                if let CommitNodeInner::Hidden(..) = nodes[left_index].inner() {
-                    return Err(Error::HiddenNode);
-                }
-            }
-            if let Some(right_index) = data.right_index {
-                if let CommitNodeInner::Hidden(..) = nodes[right_index].inner() {
-                    return Err(Error::HiddenNode);
+        // Helper function to handle the fact that our list of CommitNodes may be
```
uncomputable, line 213: Outdated doc

src/decode.rs
```diff
@@ -135,51 +182,94 @@ pub fn decode_program_arbitrary_type<I: Iterator<Item = u8>, J: Jet>(
...
+    }
+    use Converted::{Hidden, Node};
+    impl<J: Jet> Converted<J> {
+        fn get(&self) -> Result<Rc<CommitNode<J>>, Error> {
+            match self {
+                Node(rc) => Ok(Rc::clone(rc)),
+                Hidden(_) => Err(Error::HiddenNode),
+            }
+        }
+    }
```
uncomputable, line 203: My IDE doesn't like this, but maybe Clion just get an update.

src/decode.rs
```diff
@@ -408,6 +405,43 @@ pub fn decode_value<I: Iterator<Item = bool>>(
...
+            let right = stack.pop().unwrap();
+            let left = stack.pop().unwrap();
+            stack.push(StackElem {
+                value: Value::Prod(Box::new(left.value), Box::new(right.value)),
+                width: left.width * 2,
+            });
+        }
+    }
+    Ok(stack.pop().unwrap().value)
+}
```
uncomputable, line 443: 795cd3383cda71bed4aae43bbf4101d7540bbd6c: Why should we use this specialized function? We have general functions and caching in the context that should be faster. Let's keep the total number of functions low.

uncomputable:
9a197c2453f42de021b858ae38e8d0df357bc5c2: Is the `iter::equal` check a todo, or is this fixed with this PR? The test seems to pass.

uncomputable (changes requested):
Open comments 9a197c2453f42de021b858ae38e8d0df357bc5c2

src/decode.rs
apoelstra: Because this is doesn't need a context object or precomputed values. This completely removes the `context` parameter from `decode_program`.

Adding a whole extra parameter to all decoding logic to save a single function that users don't need to care about, does not seem like a reasonable tradeoff to me.

apoelstra:
@uncomputable renamed `assert_program_succeeds` and `assert_program_fails`; removed dead/outdated comment.

No other changes.

PR #120, opened 2023-06-13
Remove `Hidden` nodes

apoelstra:
I apologize for putting these all into one PR. I can split it up if you'd like. This PR:

1. Moves and renames some decoding-related types (first several commits)
2. Updates the encoding algorithm to avoid using `CommitNode::Hidden`, instead treating assertions as first-class
3. Eliminates `Hidden` nodes and makes `AssertL`/`AssertR` directly hold CMRs rather than nodes.


apoelstra:
I'm also open to pushback about any of the refactors/renames, though obviously I'll be a little annoyed at having to edit commits :)

uncomputable:
14059e922ce1af8e4bc11b5fe8519826b8aa75a3: Is there a reason why `DecodeNode` lives in `bit_encoding` while `EncodeNode` lives in `bit_encoding::encode`?

apoelstra:
@uncomputable no ... I had originally thought that `DecodeNode` could be used for both purposes, but I was wrong.

Probably I should just remove the "move `DecodeNode`" commit.

src/bit_machine/exec.rs
```diff
@@ -397,16 +397,28 @@ impl BitMachine {
...
+                            call_stack.push(CallStack::Back(1 + padl_a_b));
+                            call_stack.push(CallStack::Goto(left));
+                        }
+                        (RedeemNodeInner::AssertL(_, r_cmr), true) => {
+                            return Err(ExecutionError::ReachedPrunedBranch(*r_cmr))
+                        }
+                        (RedeemNodeInner::AssertR(l_cmr, _), false) => {
+                            return Err(ExecutionError::ReachedPrunedBranch(*l_cmr))
+                        }
+                        _ => unreachable!(),
```
uncomputable, line 421: I feel like we could reduce code duplication somehow, but let's look into that later.

src/merkle/amr.rs
```diff
@@ -141,6 +139,28 @@ impl Amr {
...
+                let l_amr = l_cmr.into();
+                let r_amr = left.as_ref().unwrap().amr;
+                let (sum_a_b, c) = ty.source.split().unwrap();
+                let (a, b) = sum_a_b.split().unwrap();
+                let d = &ty.target;
+                amr_iv
+                    .update(a.tmr().into(), b.tmr().into())
+                    .update(c.tmr().into(), d.tmr().into())
+                    .update(l_amr, r_amr)
+            }
```
uncomputable, line 160: Same as above

src/merkle/imr.rs
```diff
@@ -58,14 +58,6 @@ impl CommitMerkleRoot for Imr {
 }
 
 impl Imr {
-    /// The IV used in the second pass of IMR computation.
-    fn pass_two_iv<J: Jet>(node: &CommitNodeInner<J>) -> Self {
-        match node {
-            CommitNodeInner::Hidden(_) => Imr::tag_iv(b"Simplicity-Draft\x1fHidden"),
-            _ => Imr::tag_iv(b"Simplicity-Draft\x1fIdentity"),
-        }
-    }
```
uncomputable, line 67: 397cf75d4890f31fff27aa93edd3f619c3869b64: It is really surprising that the code works without the pass two IV. Are we running tests against the C code?

uncomputable: `decode::assert_lr` assets the same IMR for `master`. It seems that pass two of the IMR doesn't use the children IMRs, but just the children TMRs. Strange, but good find.

uncomputable (changes requested):
397cf75d4890f31fff27aa93edd3f619c3869b64: Let's remove 14059e922ce1af8e4bc11b5fe8519826b8aa75a3 since it's the easiest change in the world. The rest looks good.

src/bit_machine/exec.rs
apoelstra: I have a later PR which does reduce it somewhat (though at the expense of having a bunch more precomputed fixed vectors)

src/merkle/imr.rs
apoelstra: Heh, yeah, I was surprised too. And yes, I did check it against the C code when I first noticed this.

apoelstra:
Ok @uncomputable removed the offending commit.

uncomputable (approved):
ACK 4959565982f39904ff9d0f04d632965ede52ddf7

PR #122, opened 2023-06-14
A few miscellaneous cleanups/bugfixes

apoelstra:
These commits didn't fit into the other `decode` PRs, but I'd like to get them in if possible before shifting gears to clean up and bugfix the Merkle roots (which are the last thing we need to get the fuzztests going, other than upstream fixes).

The first two commits are just refactors. We continue to move away from the `Iterator<Item=bool>` model to putting things as methods on `BitIter`; while we are at it, we replace the recursive `Value`-decoding algorithm with an iterative one.

The third commit fixes a bug found by the fuzzer, where a program with a complicated sharing pattern causes a node to be yielded twice by the post-order iterator. Fixing this replaces the "repeated children" logic with something a little bit (IMHO) conceptually simpler.

Finally, the last two commits replace some `Hashmap` indices with flat `Vec`s, fixing a bug that was causing us not to recognize improperly-shared programs as being improperly shared.

Again, this PR does multiple things and I'm happy to split it into multiple PRs if you want. But each commit should be independent at least.

### DIFF
--- a/src/bit_encoding/bititer.rs
+++ b/src/bit_encoding/bititer.rs
@@ -21,6 +21,25 @@
 //! `Iterator<Item=bool>`.
 //!
 
+use crate::Cmr;
+
+/// Attempted to read from a bit iterator, but there was no more data
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+pub struct EarlyEndOfStreamError;
+
+/// Two-bit type used during decoding
+///
+/// Use of an enum rather than a numeric primitive type makes it easier to
+/// match on.
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, PartialOrd, Ord)]
+#[allow(non_camel_case_types)]
+pub enum u2 {
+    _0,
+    _1,
+    _2,
+    _3,
+}
+
 /// Bitwise iterator formed from a wrapped bytewise iterator. Bytes are
 /// interpreted big-endian, i.e. MSB is returned first
 #[derive(Debug)]
@@ -91,45 +110,39 @@ impl<I: Iterator<Item = u8>> BitIter<I> {
         Self::from(iter)
     }
 
-    /// Reads up to 64 bits as a big-endian number
-    ///
-    /// `n` must be between 0 and 63 inclusive; the function will panic
-    /// otherwise.
-    ///
-    /// Returns `None` if the underlying byte iterator runs out
-    pub fn read_bits_be(&mut self, mut n: usize) -> Option<u64> {
-        assert!(n < 64);
-        if n == 0 {
-            return Some(0);
-        }
+    /// Reads a single bit from the iterator.
+    pub fn read_bit(&mut self) -> Result<bool, EarlyEndOfStreamError> {
+        self.next().ok_or(EarlyEndOfStreamError)
+    }
 
-        // 0 <= avail_bits <= 8
-        // 0 <= n < 64
-        let avail_bits = 8 - self.read_bits;
-        if avail_bits < n {
-            n -= avail_bits;
-            let mask = if avail_bits < 8 {
-                (1 << avail_bits) - 1
-            } else {
-                0xff
-            };
-            let pre_result = ((self.cached_byte & mask) as u64) << n;
-            self.cached_byte = self.iter.next()?;
-            self.read_bits = 0;
-
-            if let Some(read) = self.read_bits_be(n) {
-                self.total_read += avail_bits;
-                Some(pre_result + read)
-            } else {
-                None
-            }
-        } else {
-            // 0 <= n <= 8
-            self.read_bits += n;
-            self.total_read += n;
-            let mask = if n < 8 { (1 << n) - 1 } else { 0xff };
-            Some(((self.cached_byte >> (avail_bits - n)) & mask) as u64)
+    /// Reads two bits from the iterator.
+    pub fn read_u2(&mut self) -> Result<u2, EarlyEndOfStreamError> {
+        match (self.next(), self.next()) {
+            (Some(false), Some(false)) => Ok(u2::_0),
+            (Some(false), Some(true)) => Ok(u2::_1),
+            (Some(true), Some(false)) => Ok(u2::_2),
+            (Some(true), Some(true)) => Ok(u2::_3),
+            _ => Err(EarlyEndOfStreamError),
         }
+    }
+
+    /// Reads a byte from the iterator.
+    pub fn read_u8(&mut self) -> Result<u8, EarlyEndOfStreamError> {
+        debug_assert!(self.read_bits > 0);
+        let cached = self.cached_byte;
+        self.cached_byte = self.iter.next().ok_or(EarlyEndOfStreamError)?;
+
+        Ok(cached.checked_shl(self.read_bits as u32).unwrap_or(0)
+            + (self.cached_byte >> (8 - self.read_bits)))
+    }
+
+    /// Reads a 256-bit CMR from the iterator.
+    pub fn read_cmr(&mut self) -> Result<Cmr, EarlyEndOfStreamError> {
+        let mut ret = [0; 32];
+        for byte in &mut ret {
+            *byte = self.read_u8()?;
+        }
+        Ok(Cmr::from(ret))
     }
 
     /// Accessor for the number of bits which have been read,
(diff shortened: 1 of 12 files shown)


apoelstra:
Converting to draft because I think there's a more efficient way for me to do the DAG bugfix. But I will explore this tomorrow morning.

src/bit_encoding/decode.rs
```diff
@@ -275,70 +283,66 @@ fn decode_node<I: Iterator<Item = u8>, J: Jet>(
...
+                // Bits 4 and 5: subcode
+                match bits.read_u2()? {
+                    u2::_0 => Ok(DecodeNode::Iden),
+                    u2::_1 => Ok(DecodeNode::Unit),
+                    u2::_2 => Ok(DecodeNode::Fail(bits.read_cmr()?, bits.read_cmr()?)),
+                    u2::_3 => Err(Error::StopCode),
+                }
+            }
+            u2::_3 => {
+                // Bit 4: subcade
```
uncomputable, line 337: 54e6cf2bc70669ef6dda71647e2db015339d2f59 typo: "subcode"

src/bit_encoding/bititer.rs
```diff
@@ -91,45 +110,39 @@ impl<I: Iterator<Item = u8>> BitIter<I> {
...
-                (1 << avail_bits) - 1
-            } else {
-                0xff
-            };
-            let pre_result = ((self.cached_byte & mask) as u64) << n;
-            self.cached_byte = self.iter.next()?;
-            self.read_bits = 0;
+    /// Reads a byte from the iterator.
+    pub fn read_u8(&mut self) -> Result<u8, EarlyEndOfStreamError> {
+        debug_assert!(self.read_bits > 0);
```
uncomputable, line 162: 54e6cf2bc70669ef6dda71647e2db015339d2f59: Isn't this a potential panic if users use the `BitIter`? Why should `self.read_bits` be always greater than zero?

src/bit_encoding/bititer.rs
```diff
@@ -91,45 +110,39 @@ impl<I: Iterator<Item = u8>> BitIter<I> {
...
-            } else {
-                None
-            }
-        } else {
-            // 0 <= n <= 8
-            self.read_bits += n;
-            self.total_read += n;
-            let mask = if n < 8 { (1 << n) - 1 } else { 0xff };
-            Some(((self.cached_byte >> (avail_bits - n)) & mask) as u64)
+        Ok(cached.checked_shl(self.read_bits as u32).unwrap_or(0)
```
uncomputable, line 166: 54e6cf2bc70669ef6dda71647e2db015339d2f59: Are we updating `self.read_bits` anywhere?

src/bit_encoding/bititer.rs
```diff
@@ -103,6 +105,35 @@ impl<I: Iterator<Item = u8>> Iterator for BitIter<I> {
...
+
+        let actual_sl = &sl[start / 8..(end + 7) / 8];
+        let mut iter = actual_sl.iter().copied();
+
+        let read_bits = start % 8;
+        if read_bits == 0 {
+            BitIter {
+                iter,
+                cached_byte: 0,
+                read_bits: 8,
```
uncomputable, line 123: 29a2250ed158a237204f8c06eb9ddf560b989f46: Possibly doc: `// Force iterator to read next byte for next bit`

src/bit_encoding/decode.rs
```diff
@@ -375,7 +375,10 @@ impl<'a, I: Iterator<Item = u8>> WitnessDecoder<'a, I> {
 
 impl<'a, I: Iterator<Item = u8>> WitnessIterator for WitnessDecoder<'a, I> {
     fn next(&mut self, ty: &types::Final) -> Result<Value, crate::Error> {
-        decode_value(ty, self.bits).map_err(From::from)
+        self.bits
+            .read_value(ty)
+            .map_err(Error::from)
+            .map_err(crate::Error::from)
```
uncomputable, line 381: 29a2250ed158a237204f8c06eb9ddf560b989f46: `self.bits.read_value(ty).map_err(crate::Error::from)` works directly because of `From<EarlyEndOfStreamError> for Error` in `lib.rs`

src/bit_encoding/bititer.rs
```diff
@@ -225,6 +225,14 @@ impl<I: Iterator<Item = u8>> BitIter<I> {
         Ok(result_stack.pop().unwrap())
     }
 
+    /// Decode a natural number from bits.
+    ///
+    /// If a bound is specified, then the decoding terminates before trying to
+    /// decode a larger number.
+    pub fn read_natural(&mut self, bound: Option<usize>) -> Result<usize, decode::Error> {
+        decode::decode_natural(self, bound)
+    }
```
uncomputable: 9d942f71cce4c1c817eae8086c6adf6cd04c0ebc: We will replace `decode::decode_natural(self, bound)` with an iterative method that will probably live in `BitIter` in a future PR.

src/dag.rs
```diff
@@ -599,52 +627,54 @@ impl<D: DagLike, S: SharingTracker<D>> Iterator for PostOrderIter<D, S> {
...
-                            .seen_before(&self.stack[stack_len - 1].elem)
-                            .is_some()
-                    {
-                        self.stack[stack_len - 2].right_idx = Some(self.index);
-                        self.stack.pop();
+                    // Having updated the parent indices, so that our iterator is in a
+                    // consistent state, we can safely skip here if the current item is
+                    // already yielded.
+                    if already_yielded {
+                        continue;
```
uncomputable, line 671: 8e485b5ccb5130204c12fbdf5398f7acf986b3f4 We don't need to `self.stack.pop()` like before because the current item has `processed = True` and will be popped and skipped in the next loop iteration?

We should add this to the documentation.

uncomputable (changes requested):
056f4c3cd1db0b373157390a0e0b63565f0f5578 Open comments. Looks mostly good

apoelstra:
> What is our strategy when the fuzzer finds a test vector that fails and we fix it? Should we add a unit test? Both [8e485b5](https://github.com/BlockstreamResearch/rust-simplicity/commit/8e485b5ccb5130204c12fbdf5398f7acf986b3f4) and [4210465](https://github.com/BlockstreamResearch/rust-simplicity/commit/42104652d61f00211da5e3f2251885ba3cc21b02) mention test vectors that are nowhere in the code.

Yep, our strategy is to add unit tests. Thanks for the reminder; I guess I can add those tests to this PR.

src/bit_encoding/bititer.rs
apoelstra: Not in this function, no. `read_bits` is how much of the current byte we've read ... and if we read an entire byte, this number doesn't change.

src/dag.rs
apoelstra: This question only makes sense in the context of the old (wrong) algorithm. We already popped the item at the top of the loop. There's no reason we would pop anything else here, so I don't know what I would add to the documentation.

apoelstra:
Corrected a couple comment issues, and added the unit tests (which previously appeared in commit messages with "blocked on #118 but ofc #118 is now long-since merged ... that's what I get for building PRs from out-of-order commits).

src/bit_encoding/bititer.rs
uncomputable: Ah, that makes a lot of sense. I mixed it up with `total_read`.

src/dag.rs
```diff
@@ -599,52 +627,54 @@ impl<D: DagLike, S: SharingTracker<D>> Iterator for PostOrderIter<D, S> {
...
-                    // In this case we have a repeated node which nonetheless made it
-                    // into the stack, so we pop it here.
-                    if check_sibling
-                        && self
-                            .tracker
-                            .seen_before(&self.stack[stack_len - 1].elem)
-                            .is_some()
-                    {
-                        self.stack[stack_len - 2].right_idx = Some(self.index);
-                        self.stack.pop();
```
uncomputable, line 642: 5364f815565567e73abbda4a4d03aa41944bc84e: Regarding `self.stack.pop()`: This line pops the sibling on `master`, which is removed by the commit.

The current item is already popped at the top of the loop, you're right, but its sibling still sits on the stack unless we pop it. We assert that the sibling is processed, so it is popped and skipped by the next loop iteration.

uncomputable (approved):
ACK e72d614526300c643a90d3aee089799f1ba80cad

src/dag.rs
apoelstra: Right, exactly. It was wrong to treat siblings as being special; we could have repeats of siblings *or* any of their descendants, which requires more general code.

PR #126, opened 2023-06-17
merkle: overhaul module, fix several bugs

apoelstra:
After this PR, and a PR to update our vendored Simplicity version, we should be able to fuzz the C and Rust implementations against each other without finding bugs.

I'll PR that next (vendoring update and fuzztest) and then shift gears to making even more dramatic changes to `RedeemNode` and `CommitNode`.


uncomputable (approved):
ACK 2747d546c2f9e45b8cb78b2d91d416dd009418a5

PR #127, opened 2023-06-18
Update vendored Simplicity and add a fuzztest

apoelstra:
This fuzz test currently just fuzzes the decoding logic between C and Rust, checking that they are consistent in what programs they decode and their CMR/AMR/IMR computations.

To check that the bit machine implementations are also consistent, change `TestUpTo::CheckOneOne` to `TestUpTo::Everything` and check `c_result.eval_result` against the result of the Rust bit machine implementation.

I didn't do this here because I think it's useful to have a test that covers only the smaller scope; long pipelines are harder for fuzzers to work with. I think we should add the longer test as a separate test.

I'm happy to split this PR into two, one that does the update and one that does the fuzztest, if that's easier.


fuzz/fuzz_targets/c_rust_merkle.rs
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    #[test]
+    fn duplicate_crash() {
+        let mut a = Vec::new();
+        extend_vec_from_hex("7b88f500", &mut a);
+        super::do_test(&a);
+    }
+}
```
uncomputable, line 83: d4f0ed44fb0716acac61d54787369c29949caadf: I never understood what this part does. Can we remove it?

uncomputable (approved):
ACK f47bfd862be0d85bcb68b01f9569ec38b7ef9f47

fuzz/fuzz_targets/c_rust_merkle.rs
uncomputable: If this is obsolete code, we can remove all of it in another PR.

PR #128, opened 2023-06-18
Node overhaul 1: introduce new *`Node` types

apoelstra:
This is part of another multi-PR project to replace `RedeemNode` and `CommitNode` with three new node types, `ConstructNode` (roughly the same as our current `CommitNode`), `CommitNode` (which will have complete types and be immediately usable for e.g. address computation), and `RedeemNode` (pretty-much the same as our current `RedeemNode`).

The idea is that when building a node programmatically, you would use `ConstructNode`s, which have no way to be de/serialized and generally have pretty limited functionality. Then you'd convert to `CommitNode`s, which have a canoncial bit-encoding as well as a human-readable encoding, and represent a complete program, up to witnesses and disconnects. Though for now the code continues to treat `Disconnect` as always-populated and basically a version of `Comp` with a weird CMR. We can revisit `disconnect` once we start writing nontrivial code using it.

This PR does a couple cleanups then introduces some new data structures. It does not actually use them anywhere or replace any existing code, which it turns out is fairly invasive and needs to happen across many more commits (and probably multiple PRs). But it does include basic conversions between the new node types so you can see how that'll work.

Happy to shrink or re-arrange this PR if that would make review easier.

### DIFF
--- a/src/dag.rs
+++ b/src/dag.rs
@@ -25,42 +25,23 @@ use crate::jet;
 use crate::node::{self, Node, NodeData};
 use crate::{Imr, Value};
 
-/// Generic container for Simplicity DAGs
+/// Abstract node of a directed acyclic graph.
+///
+/// Tracks the arity (out-degree) of nodes in a Simplictiy program, as well
+/// as whether they represent `witness` or `disconnect` combinators, which
+/// are treated specially when working with the graph structure of
+/// Simplicty programs.
 pub enum Dag<T> {
-    /// Identity
-    Iden,
-    /// Unit constant
-    Unit,
-    /// Left injection of some child
-    InjL(T),
-    /// Right injection of some child
-    InjR(T),
-    /// Take of some child
-    Take(T),
-    /// Drop of some child
-    Drop(T),
-    /// Composition of a left and right child
-    Comp(T, T),
-    /// Case of a left and right child
-    Case(T, T),
-    /// Left-assertion of a child
-    AssertL(T),
-    /// Right-assertion of a child
-    AssertR(T),
-    /// Pair of a left and right child
-    Pair(T, T),
-    /// Disconnect of a left and right child
-    Disconnect(T, T),
-    /// Witness data (missing during commitment, inserted during redemption)
+    /// Combinator with no children
+    Nullary,
+    /// Combinator with one child
+    Unary(T),
+    /// Combinator with two children
+    Binary(T, T),
+    /// Witness node, which gets special treatment
     Witness,
-    /// Universal fail
-    Fail,
-    /// Hidden CMR
-    Hidden,
-    /// Application jet
-    Jet,
-    /// Constant word
-    Word,
+    /// Disconnect node, which gets special treatment
+    Disconnect(T, T),
 }
 
 /// How much sharing/expansion to do when running an iterator over a DAG
@@ -319,46 +300,22 @@ pub trait DagLike: Sized {
     /// Accessor for the left child of the node, if one exists
     fn left_child(&self) -> Option<Self> {
         match self.as_dag_node() {
-            Dag::Iden => None,
-            Dag::Unit => None,
-            Dag::InjL(sub) => Some(sub),
-            Dag::InjR(sub) => Some(sub),
-            Dag::Take(sub) => Some(sub),
-            Dag::Drop(sub) => Some(sub),
-            Dag::Comp(left, _) => Some(left),
-            Dag::Case(left, _) => Some(left),
-            Dag::AssertL(left) => Some(left),
-            Dag::AssertR(right) => Some(right), // note that we treat the child of an assertR as a left child!
-            Dag::Pair(left, _) => Some(left),
-            Dag::Disconnect(left, _) => Some(left),
+            Dag::Nullary => None,
+            Dag::Unary(sub) => Some(sub),
+            Dag::Binary(left, _) => Some(left),
             Dag::Witness => None,
-            Dag::Fail => None,
-            Dag::Hidden => None,
-            Dag::Jet => None,
-            Dag::Word => None,
+            Dag::Disconnect(left, _) => Some(left),
         }
     }
 
     /// Accessor for the right child of the node, if one exists
     fn right_child(&self) -> Option<Self> {
         match self.as_dag_node() {
-            Dag::Iden => None,
-            Dag::Unit => None,
-            Dag::InjL(_) => None,
-            Dag::InjR(_) => None,
-            Dag::Take(_) => None,
-            Dag::Drop(_) => None,
-            Dag::Comp(_, right) => Some(right),
-            Dag::AssertL(_) => None,
-            Dag::AssertR(_) => None, // note that we treat the child of an assertR as a left child!
-            Dag::Case(_, right) => Some(right),
-            Dag::Pair(_, right) => Some(right),
-            Dag::Disconnect(_, right) => Some(right),
+            Dag::Nullary => None,
+            Dag::Unary(_) => None,
+            Dag::Binary(_, right) => Some(right),
             Dag::Witness => None,
-            Dag::Fail => None,
-            Dag::Hidden => None,
-            Dag::Jet => None,
-            Dag::Word => None,
+            Dag::Disconnect(left, _) => Some(left),
         }
     }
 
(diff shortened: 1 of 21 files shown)


apoelstra:
Ok, this should be good to go. I'll stop poking at it.

src/node/inner.rs
```diff
@@ -0,0 +1,230 @@
...
+
+use super::FailEntropy;
+use crate::jet::Jet;
+use crate::{Cmr, Value};
+
+use std::fmt;
+use std::sync::Arc;
+
+/// Underlying combinator of a [`CommitNode`].
+/// May contain references to children, hash payloads or jets.
```
uncomputable, line 23: 37a6c18ec24f06e9f6a2255b5cd529658f473abd: Copy-paste error

src/node/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+
+impl<N: NodeData<J>, J: Jet> hash::Hash for Node<N, J>
+where
+    N::CachedData: hash::Hash,
+{
+    fn hash<H: hash::Hasher>(&self, h: &mut H) {
+        self.cmr.hash(h);
+        self.data.hash(h);
+    }
+}
```
uncomputable, line 292: 37a6c18ec24f06e9f6a2255b5cd529658f473abd: One downside of this generic approach is that things like unification arrows and node bounds influence equality. I feel like we should implement custom `Eq` implementations for `Final` and `{ConstructData,Commit,Redeem}Data`.

src/node/mod.rs
```diff
@@ -12,6 +12,76 @@
...
+//! All nodes represent the root of an expression. Expressions whose source
+//! and target types are both unit are called "programs". Generally speaking,
+//! any nodes that users hold directly will be programs.
+//!
+//! There are three main node types:
+//!
+//! 1. [`ReedemNode`] represents a Simplicity node as it exists on the blockchain.
+//!    Every witness node is populated with a correctly-typed [`Value`], every
+//!    disconnect node has a child, and every non-executed branch of a `case`
+//!    combinator is replaced by a hidden node, and there is nothing you can do
```
uncomputable, line 30: bebad27e1959bbd1f6e7229676035e139ae82fb7: If hidden nodes are a transport-layer thing now, I would avoid mentioning them. Maybe "every case is converted into a left or right assertion if only the left or right branch, respectively, is executed"

src/node/mod.rs
```diff
@@ -12,6 +12,76 @@
...
+//!
+//! 1. [`ConstructNode::finalize_types`] converts a [`ConstructNode`] to a
+//!    [`CommitNode`] by setting any free variables, as well as the source and
+//!    target of the root node, to unit.
+//!
+//!    This conversion requires no input from the user but may fail with a type
+//!    error, in case of infinitely-sized types or in the case that the unit
+//!    bounds cannot be applied.
+//!
+//! 2. [`CommitNode::rtl_finalize`] converts a [`CommitNode`] to a [`RedeemNode`]
```
uncomputable, line 64: bebad27e1959bbd1f6e7229676035e139ae82fb7: `CommitNode::rtl_finalize` does not exist in the final diff. Also, what does `rtl` mean?

src/node/mod.rs
```diff
@@ -12,6 +12,76 @@
...
+//!    by attaching witnesses to each witness node, and deciding whether to hide
+//!    branches for each `case` node. It iterates through the entire program, in
+//!    right-to-left post-order, to ensure that both children of every `case`
```
uncomputable, line 67: bebad27e1959bbd1f6e7229676035e139ae82fb7: Why right-to-left post order? Is the post-order iterator that we have not sufficient? (The method doesn't exist so I cannot read the code)

src/node/mod.rs
```diff
@@ -0,0 +1,226 @@
...
+        Self::comp(context, &pair_left_iden, &case_right_true)
+    }
+}
+
+/// A node in a Simplicity expression.
+///
+/// There are several kinds of nodes...
+///
+/// For equality and hashing purposes, nodes are characterized entirely by their
+/// CMR and cached data. (For `Contruct` nodes this is, ...)
```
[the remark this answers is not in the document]
apoelstra: Yep, I'll take a look to see which ones are missing.

src/node/mod.rs
apoelstra: Yep -- I should actually probably just replace this line with a TODO since we haven't implemented anything that would require full pruning. (If we added a fourth `UnprunedNode`, which you had to run the executor to get a `RedeemNode` from, that would achieve that. Thoughts?)

src/node/mod.rs
apoelstra: Yes, it will come in a later PR. Would you like me to delete this comment for now? `rtl` stands for `right-to-left`.

src/node/mod.rs
apoelstra: Because when iterating over a `comp(witness, case(<left>, <right>))`, you want to iterate over the children first, then the `witness`, and then the `case`, so that you know what you're doing with both children before you have to make decisions about hiding.

The PR that introduces this method will have a ton of comments explaining how this all works. Would you like me to delete these "teaser" comments for now?

src/dag.rs
```diff
@@ -27,40 +27,16 @@ use crate::{Imr, Value};
 
 /// Generic container for Simplicity DAGs
 pub enum Dag<T> {
```
uncomputable, line 29: b4f11b4e205413003b6c58fb595ed57853808534: We should update this description.

All these different graph types can be confusing. `Dag` is a pretty generic term, so it `dag.rs`. Maybe we should rename it to `iter.rs` :) IDK what `Dag` would become. Happy to track this in an issue and make a PR later.

uncomputable:
7cb438bfceea9582a6d0c728b0addb04ea8e9105: We have come a long way since we had relative indices in node types :)

src/dag.rs
apoelstra: Let's file a separate issue. I agree calling the module `iter` would make more sense. `Dag` I'm unsure of, I think the current name is ok. I wouldn't mind `DagNode` though I think that could be confused with `Node`.

Worth noting that users almost never interact with this enum directly.

src/node/mod.rs
apoelstra: I think that:

* `ConstructData` should not implement any equality or hashing traits
* `CommitData` should implement `PartialEq` which compares incomplete IMRs, but not `Eq` or `Hash`
* `RedeemData` should implement all three traits based on IMR

Does this sound reasonable?

src/node/mod.rs
uncomputable: The docs say what you said, but there is still a big question mark inside my head. Naively pruning would execute the program (in post order). You seem to be doing something different. Anyway, I look forward to the followup on this.

apoelstra: We discussed this a bit on IRC, but basically, when your satisfier is deciding what bit to input to a `case` combinator, you want the children of the `case` to have already been evaluated, so that you know which one(s) are available and cheapest. If you iterate in left-to-right order you don't have this, and you have to do some complicated lookahead.

uncomputable:
In a future PR we should add documentation to `NodeData` and its implementations `Construct`, `Commit`, `Redeem`, as well as to the node types `ConstructNode`, `CommitNode`, `RedeemNode`.

src/node/construct.rs
```diff
@@ -48,6 +49,42 @@ impl<J: Jet> ConstructNode<J> {
     pub fn arrow(&self) -> &Arrow {
         &self.data.arrow
     }
+
+    /// Sets the source and target type of the node to unit
+    pub fn type_as_program(&self) -> Result<(), types::Error> {
```
uncomputable, line 54: b3bce4e36dc504598954839cb0c9ab3556a2725e: Confusing method name. Maybe `set_program_type`? At least `set` should be in the name.

src/node/construct.rs
```diff
@@ -48,6 +49,42 @@ impl<J: Jet> ConstructNode<J> {
...
+    /// [`Self::finalize_types_without_fixing`] instead.
+    pub fn finalize_types(&self) -> Result<Arc<CommitNode<J>>, types::Error> {
+        self.type_as_program()?;
+        self.finalize_types_without_fixing()
+    }
+
+    /// Convert a [`ConstructNode`] to a [`CommitNode`] by finalizing all of the types.
+    ///
+    /// Also sets the source and target type of this node to unit.
+    pub fn finalize_types_without_fixing(&self) -> Result<Arc<CommitNode<J>>, types::Error> {
```
uncomputable, line 79: b3bce4e36dc504598954839cb0c9ab3556a2725e: Wrong documentation. This method does not enforce 1 → 1 expressions.

Also, we might want to name this method in accordance with `type_as_program` from above, to show the contrast. Maybe `finalize_types_no_program_type`.

src/node/commit.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    pub fn new_unfinal(arrow: FinalArrow, inner: Inner<&Arc<Self>, J, NoWitness>) -> Self {
+        let first_pass_imr = Self::first_pass_imr(inner);
+        CommitData {
+            first_pass_imr,
+            imr: first_pass_imr.map(|imr| Imr::compute_pass2(imr, &arrow)),
+            arrow,
+            phantom: PhantomData,
+        }
+    }
```
uncomputable, line 129: b3bce4e36dc504598954839cb0c9ab3556a2725e: Kind of confusing how `CommitNode::new_unfinal` gets the `FinalArrow`. I get it, it is used in `RedeemNode::unfinalize`, but the name doesn't make sense in the context of `CommitNode`. Maybe call it `CommitNode::new_final` and make `CommitNode::new` call `CommitNode::new_final`?

src/node/redeem.rs
```diff
@@ -0,0 +1,191 @@
...
+use std::marker::PhantomData;
+use std::sync::Arc;
+
+#[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
+pub enum Redeem {}
+
+impl<J: Jet> NodeData<J> for Redeem {
+    type CachedData = Arc<RedeemData<J>>;
+    type Witness = Arc<Value>;
+    type SharingId = (Cmr, Imr);
```
uncomputable, line 32: 1778550599030115c2d33b7417d46aa9e7e734f2: Let's make this `type SharingId = Imr`

uncomputable (changes requested):
699e5f5cd127bc110abaeda9b8a62917d3534c68 there are unresolved comments

uncomputable:
I really like how `node::Inner` represents expressions, witness data and node metadata in one unified structure. This feels really elegant.

apoelstra:
Addressed comments.

apoelstra:
Rebased.

uncomputable (approved):
ACK 052cc9e0870818507a3626b8fb123673d38951d5

PR #129, opened 2023-06-19
Simplicity descriptors

uncomputable:
This PR adds descriptors of Simplicity outputs. This is the API that wallets will use to interact with rust-simplicity.

Currently the descriptors are very limited. We can only construct tap trees with a single leaf. Future PRs can add support for multiple leaves, non-Simplicity leaves, etc.

### DIFF
--- a/src/policy/descriptor.rs
+++ b/src/policy/descriptor.rs
@@ -1,0 +1,173 @@
+use crate::policy::key::PublicKey32;
+use crate::policy::satisfy::PolicySatisfier;
+use crate::{policy, Cmr, Context, Policy};
+use elements::secp256k1_zkp;
+use elements::taproot::{LeafVersion, TaprootBuilder, TaprootSpendInfo};
+use elements_miniscript::{MiniscriptKey, ToPublicKey};
+use std::fmt;
+use std::str::FromStr;
+
+/// Bytes of x-only public key whose discrete logarithm (secret key) is unknown
+///
+/// Taken from BIP 341
+const UNSPENDABLE_PUBLIC_KEY: [u8; 32] = [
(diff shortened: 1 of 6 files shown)


src/policy/descriptor.rs
```diff
@@ -0,0 +1,173 @@
...
+        // let mut mac = crate::exec::BitMachine::for_program(&program);
+        // let _output = mac
+        //     .exec(&program, &satisfier.env)
+        //     .expect("sanity check");
+
+        let program_and_witness_bytes = program.encode_to_vec();
+        let cmr_bytes = Vec::from(program.cmr.as_ref());
+
+        // FIXME: Should env be public?
+        let control_block = satisfier.env.control_block();
```
apoelstra, line 142: In fe24e0c692af3361be7a2476d34baca666c45903:

Is this `FIXME` asking whether `env` should be a field or an accessor method on `PolicySatisfier`? I think it should be a method.

uncomputable: I made `env` private. The other members `keys` and `preimages` are still open, but we might want some sort of integration with `HashMap<Pk, Sig>` etc. which implement the `Satisfier` trait, so I will leave them for later.

apoelstra: Why make only one member private? Either this a POD type and everything should be public or it is an encapsulated type and none should be.

In any case, how do are you supposed to construct this object now?

apoelstra (dismissed):
ACK dd4423f9a518154575978b01dd045c62d5e9206f

uncomputable:
The satisfier did dynamic signing and was ugly. I pushed a new commit to fix that.

apoelstra:
This no longer compiles without the `elements` feature flag.

uncomputable:
My bad. I wasn't running my local script to verify compilation on all feature combinations.

apoelstra (approved):
ACK 296c801ff8b2cf0aa22a5b1ba1c7be1464c06b39

PR #131, opened 2023-06-21
Make bit_encoding modules private

### LINKED ISSUE #121: Make bit_encoding modules private
Remove `bit_encoding::bititer` and `bit_encoding::bitwriter` from `lib.rs` and make both modules private (see FIXMEs). This requires an update to the generated Rust code, which we can do after #120 is merged.

uncomputable:
Fixes #121. Depends on https://github.com/BlockstreamResearch/simplicity/pull/158.


apoelstra (approved):
ACK fef13c36ff7f7f4367d2d6a3a51bffa6814056bf

PR #132, opened 2023-06-21
Node overhaul 2: transition decoder logic to use the new node types

apoelstra:
We temporarily introduce some conversion functions from the new node types to the old ones. When we do encoding we'll introduce conversions in the opposite direction. Then later we'll delete the entire `core` module along with those functions.

Also includes several commits which are essentially cleanups; only the last two commits are the meat of this PR.


apoelstra:
Actually I'm gonna improve the early context commits and push those as their own PR first.

PR #133, opened 2023-06-21
Remove context object

apoelstra:
We are currently using the context object for two things: type variable names and precomputed types. Replace both of those with simpler solutions. Commit messages provide more detailed justification, or see IRC discussion.


uncomputable (commented):
Checkpoint 9d50eed5c7f17c2e7fb2ada288c2e0ef2e51b252. Looks good, but I am confused about why 5de2e6125c920a972c3bf1a4447d0e35b4c65822 was necessary (see IRC)

apoelstra:
Updated the "split up `Constructible`" commit to make the blanket impls on `Node` clearer.

uncomputable:
We discussed on IRC and came to the conclusion that splitting `Composible<W, J: Jet>` into `CoreComposible`, `WitnessComposible<W>` and `JetComposible<J: Jet>` is the best way because it avoids verbose type annotations:

Methods like `Arrow::for_iden()` which would be implemented in `impl<W, J: Jet> Constructible<W, J> for Arrow` are the same for all `W` and `J`. But if we call this method, the compiler does not know its types. `Arrow` has no generics, so it doesn't help us. Even if we call `Arrow::for_iden()` inside say `impl<'a, J: Jet> Constructible<&'a NoWitness, J> for ConstructData<J>` there is nothing that forces the `W` of `Arrow` and the `W` of `ConstructData<J>` to be the same! They are different type variables! (Same for `J`.) This forces us to write `<Arrow as Constructible<J>>::iden()` which is verbose.

uncomputable (approved):
ACK b2d7314ef136c45cf08d03fb37a19529e3dd8f6f

PR #134, opened 2023-06-21
Node overhaul 2: use new node types for decoding

apoelstra:
We temporarily introduce some conversion functions from the new node types to the old ones. When we do encoding we'll introduce conversions in the opposite direction. Then later we'll delete the entire core module along with those functions.

Replaces #132.

### DIFF
--- a/src/node/redeem.rs
+++ b/src/node/redeem.rs
@@ -13,13 +13,14 @@
 //
 
 use crate::analysis::NodeBounds;
-use crate::dag::MaxSharing;
+use crate::dag::{DagLike, InternalSharing, MaxSharing};
 use crate::jet::Jet;
 use crate::types::{self, arrow::FinalArrow};
-use crate::{Amr, Cmr, FirstPassImr, Imr, Value};
+use crate::{Amr, BitIter, Cmr, Error, FirstPassImr, Imr, Value};
 
 use super::{CommitData, CommitNode, Inner, NoWitness, Node, NodeData};
 
+use std::collections::HashSet;
 use std::marker::PhantomData;
 use std::sync::Arc;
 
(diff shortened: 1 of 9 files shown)


uncomputable:
bdbf029dd1600d72f7cad7db374327147b906bf8: We might want to remove `CommitNode::finalize_no_witnesses()` once we have a general finalization. It's mostly for testing and every useful program will contain witness nodes.

uncomputable:
4e7c921b50677ae6ab388109e5e3e46030c6cd20: Users will want to share sub expressions like hash constructs to build large programs, but they can do so using the human-readable format or similar.

apoelstra:
> [bdbf029](https://github.com/BlockstreamResearch/rust-simplicity/commit/bdbf029dd1600d72f7cad7db374327147b906bf8): We might want to remove `CommitNode::finalize_no_witnesses()` once we have a general finalization. It's mostly for testing and every useful program will contain witness nodes.

Agreed. Later I even add a `SimpleFinalizer` which takes a witness iterator and simulates the old-style finalization code. Then we can pass `iter::empty` to that to get the same behavior as `finalize_no_witnesses` if we really want it.

> [4e7c921](https://github.com/BlockstreamResearch/rust-simplicity/commit/4e7c921b50677ae6ab388109e5e3e46030c6cd20): Users will want to share sub expressions like hash constructs to build large programs, but they can do so using the human-readable format or similar.

Yep.

src/node/redeem.rs
```diff
@@ -188,4 +189,51 @@ impl<J: Jet> RedeemNode<J> {
...
+
+        // 4. Check sharing
+        let mut imrs: HashSet<Imr> = HashSet::new();
+        for data in program.as_ref().post_order_iter::<InternalSharing>() {
+            if !imrs.insert(data.node.imr()) {
+                return Err(Error::SharingNotMaximal);
+            }
+        }
+
+        Ok(program)
```
uncomputable, line 237: Using `is_shared_as` seems more canonical.
```
if program.as_ref().is_shared_as::<MaxSharing<Redeem, J>>() {
    Ok(program)
} else {
    Err(Error::SharingNotMaximal)
}
```

uncomputable (approved):
ACK 4e7c921b50677ae6ab388109e5e3e46030c6cd20. Feel free to address the nit. I am happy to merge tomorrow.

src/node/redeem.rs
apoelstra: We could use `is_shared_as` but that's at least twice as slow since it involves running two iterators at once. In practice it's probably more than twice as slow because it involves feeding IMRs into a hashmap which are 8 times as large as pointer IDs.

I'll add a comment explaining that they're equivalent though.

apoelstra: Oh, maybe it's not twice as slow. It'll be slower but it's probably not much of a cost at all compared to the extra cost from using this iterator at all, versus an in-line algorithm.

At some point we should add some benchmarks and try to optimize this path, since it's used for decoding on-chain programs. But for now I'll leave it be.

uncomputable (approved):
ACK 12bbbe9a939206877c040d3300123fac13e18e0e

PR #136, opened 2023-06-22
dag: fix bug with the right child of disconnect nodes

apoelstra:
This is a 1-line patch with a 55-line test, most of which is chatty comments. Let me know if I should attempt to reduce this.


uncomputable (approved):
ACK 79703afbbf230dc7274921d93345e6a35191b6b8

PR #137, opened 2023-06-22
Node overhaul 3: introduce generic finalizing logic

apoelstra:
This is a big algorithm that doesn't really have any tests as of this PR. When we introduce the policy satisfier it will get some heavy use and we can double-back and fix any bugs that we find (though I've already done a few iterations of this so hopefully it's okay).

Meanwhile, review of this PR should probably just assume that the fine details are okay, and instead review the algorithm itself and surrounding comments.


apoelstra:
Going to close this and take a different approach.

PR #138, opened 2023-06-22
Node overhaul 3: generalize node conversion a bit, add `SimpleFinalizer` for tests

apoelstra:
Introduce a new trait for node conversions, and the ability to prune nodes while converting. Unclear yet whether we will use this for policy satisfaction, but the result seems generally useful, and allows us to port all our existing unit tests to the new node types. Introduces a new `SimpleFinalizer` for this purpose which implements the trait and simple adapts a witness iterator to it.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -97,6 +97,9 @@ pub enum Error {
     Policy(policy::Error),
     /// Program does not have maximal sharing
     SharingNotMaximal,
+    /// Finalization led to the root node being hidden, because not enough witnesses
+    /// were provided (or some branches incomplete branches were kept instead of pruned).
+    IncompleteFinalization,
 }
 
 impl fmt::Display for Error {
@@ -113,6 +116,9 @@ impl fmt::Display for Error {
             #[cfg(feature = "elements")]
             Error::Policy(ref e) => fmt::Display::fmt(e, f),
             Error::SharingNotMaximal => f.write_str("Decoded programs must have maximal sharing"),
+            Error::IncompleteFinalization => {
+                f.write_str("finalization did not complete (missing witnesses)")
+            }
         }
     }
 }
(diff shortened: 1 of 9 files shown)


uncomputable:
If we already plan to abandon right-to-left traversal, then I wouldn't merge this. Let's take a step back and restart.

apoelstra:
@uncomputable why? RTL is a totally minor thing.

src/lib.rs
```diff
@@ -97,6 +97,9 @@ pub enum Error {
     Policy(policy::Error),
     /// Program does not have maximal sharing
     SharingNotMaximal,
+    /// Finalization led to the root node being hidden, because not enough witnesses
+    /// were provided (or some branches incomplete branches were kept instead of pruned).
+    IncompleteFinalization,
```
uncomputable, line 102: 86c4195171d58817d5da5aa60a1db9da93267e07: This error variant is not being used and it is confusing. Why would we ever hide the root node?

uncomputable (commented):
Checkpoint 86c4195171d58817d5da5aa60a1db9da93267e07. There is one open question

apoelstra:
@uncomputable removed the error variant. It will reappear in the next PR (with a better doccomment).

uncomputable (approved):
ACK 884b76d74911b2113801f10530e76332981faefd

src/node/mod.rs
```diff
@@ -517,31 +519,24 @@ impl<N: NodeData<J>, J: Jet> Node<N, J> {
...
-        FD: FnMut(
-            &PostOrderIterItem<&Self>,
-            Inner<&Arc<Node<M, J>>, J, M::Witness>,
-        ) -> Result<M::CachedData, E>,
-        FW: FnMut(&PostOrderIterItem<&Self>, &N::Witness) -> Result<M::Witness, E>,
+        C: Converter<N, M, J>,
     {
         let mut converted: Vec<Arc<Node<M, J>>> = vec![];
-        for data in self.post_order_iter::<S>() {
+        for data in self.rtl_post_order_iter::<S>() {
```
apoelstra, line 536: Lol, oops, I didn't mean to use the `rtl_` variant here. I'll adjust the next PR to change it.

PR #139, opened 2023-06-23
Node overhaul 4: encoding, unit tests and bit machine

apoelstra:
This includes a few things at once, so I'm happy to split it up, but other than the encoding logic (the first commit) these are mostly just search-and-replacing types and patching the results up.

After this the only remaining code using the old-style nodes is the `policy` module, which will get its own overhaul series.


src/node/redeem.rs
```diff
@@ -175,6 +175,18 @@ impl<J: Jet> RedeemNode<J> {
...
+        arrow: FinalArrow,
+        inner: Inner<Arc<Self>, J, Arc<Value>>,
+        inner_cached: Inner<&Arc<RedeemData<J>>, J, Arc<Value>>,
+    ) -> Result<Self, types::Error> {
+        Ok(RedeemNode {
+            cmr: Cmr::unit(),
+            inner,
+            data: Arc::new(RedeemData::new(arrow, inner_cached)),
+        })
+    }
```
uncomputable, line 190: 4143ea8f374f1633e3c5077f2df25f1bd2d04d2d: I feel this should be removed along with `core::redeem::RedeemNode::to_node()` when we get to that. A node with unit CMR looks weird.

src/bit_encoding/decode.rs
```diff
@@ -781,19 +803,22 @@ mod tests {
     fn root_unit_to_unit() {
         // main = jet_eq_32 :: 2^64 -> 2 # 7387d279
         let justjet = vec![0x6d, 0xb8, 0x80];
-        // Should be able to decode this as a CommitNode...
+        // Should be able to decode this as an expression...
+        let mut iter = BitIter::from(&justjet[..]);
+        decode_expression::<_, Core>(&mut iter).unwrap();
```
uncomputable, line 807: 750abaaa8038045683030c40c0f84da2259ae605: We could even make this `ConstructNode::decode` as a contrast to the other two node types.

uncomputable:
750abaaa8038045683030c40c0f84da2259ae605: Does this mean we used to share witness / disconnect nodes in the encoding if they had the same type + CMR?

src/bit_encoding/decode.rs
```diff
@@ -734,25 +730,48 @@ mod tests {
...
+            // Same program but with each `witness` replaced by `comp iden witness`.
             (
                 vec![
-                    0xde, 0x87, 0x04, 0x08, 0xe6, 0x8c, 0x41, 0x08,
-                    0x38, 0x15, 0xc6, 0x22, 0x8d, 0xb7, 0x10, 0x46,
-                    0x02, 0x00,
+                    0xe0, 0x28, 0x70, 0x43, 0x83, 0x00, 0xab, 0x9a,
+                    0x31, 0x04, 0x20, 0xe0, 0x57, 0x18, 0x8a, 0x36,
+                    0xdc, 0x41, 0x18, 0x08, 
                 ],
```
uncomputable, line 775: 750abaaa8038045683030c40c0f84da2259ae605: Does this mean we used to share witness / disconnect nodes in the encoding if they had the same type + CMR?

src/node/mod.rs
```diff
@@ -528,12 +521,12 @@ impl<N: NodeData<J>, J: Jet> Node<N, J> {
...
     pub fn convert<S, M, C>(&self, converter: &mut C) -> Result<Arc<Node<M, J>>, C::Error>
     where
-        S: for<'a> SharingTracker<SwapChildren<&'a Self>> + Default,
+        S: for<'a> SharingTracker<&'a Self> + Default,
         M: NodeData<J>,
         C: Converter<N, M, J>,
     {
         let mut converted: Vec<Arc<Node<M, J>>> = vec![];
-        for data in self.rtl_post_order_iter::<S>() {
+        for data in self.post_order_iter::<S>() {
```
uncomputable, line 529: e97fbc8f3ffb9eb3aaf9f35dca953a4b9e64048b: Why does LTR / RTL not affect any unit tests? Is it because children are still visited before parents, and no implementation of `Converter` tracks the nodes it has seen? As long as `convert_witness()` and `convert_data()` are called with the right arguments (which they are) the same DAG will be built.

uncomputable (commented):
e97fbc8f3ffb9eb3aaf9f35dca953a4b9e64048b: This is basically good to go, but I would like to discuss the above points before we merge.

src/bit_encoding/decode.rs
apoelstra: I'd prefer to pretend that `ConstructNode`s have no canonical encoding, so you have to use this obscure non-exported function to "decode" them.

Though I'm open to arguments otherwise.

src/bit_encoding/decode.rs
apoelstra: Yep. Well, we used to allow it, anyway -- the old `CommitNode` would do no sharing checks, only `RedeemNode` was.

Then as it happened, our encoding would actually do it, which is why all our test vectors (which I made with the old code) had sharing that we would now consider illegitimate.

apoelstra: Once we have the human-readable encoding and you can verify these binary blobs, I hope you'll feel better about these test vectors :).

src/node/mod.rs
apoelstra: Yep, exactly. There's no "local" way to tell what direction the iterator is going in, so you would need to somehow be tracking nodes you've seen before...and the `convert` function is designed so that you don't need to do that.

The only place it would've mattered was in my old finalization code, which had some insane logic where it would be iterating an AST in parallel to the Simplicity DAG, and caching random bits of data which only made sense assuming a particular iteration order. I think it's good for multiple reasons that we decided to throw that away :)

src/node/redeem.rs
apoelstra: :eyes: lol great catch. Indeed, this funciton is only ever used by the temporary conversion code, and I had totally forgotten it existed (and forgotten to remove it in my later commit which deletes `core`).

I'm kinda surprised this passes all tests...it looks like, almost by accident, I only ever used it during encoding and for the policy tests, none of which check CMRs.

Anyway I think it's fine to merge, since it's only needed to shim tests, and we will delete it soon.

src/bit_encoding/decode.rs
apoelstra: Well, I guess, eventually we'd like a CLI tool that can decode expressions even if they're not unit->unit.

But we'd have to think a bit about what sharing rules we want to enforce. "No rules" would be simplest and probably makes the most sense. But I think, forbidding sharing of witnesses and disconnects maybe also makes sense. Or forbidding sharing entirely.

Should `ConstructNode`s have a human-readable representation too?

These are all good questions which we should resolve, but I don't think they're high-priority right now and I think there'd be nontrivial work no matter what.

uncomputable: This is a big topic, but we definitely need an encoding of raw expressions. People want to share their math gadgets with each other. `ConstructNode::{de,en}encode()` is public so encourage users to use it.

For now, let's merge this PR as is.

uncomputable (approved):
ACK e97fbc8f3ffb9eb3aaf9f35dca953a4b9e64048b

PR #140, opened 2023-06-24
Some simple `policy` module improvements and cleanups

apoelstra:
This PR does some (I think) uncontroversial cleanups to the `policy` module in preparation for translating it to use the new node types.


apoelstra:
Actually, the "encapsulate the enum in `Fragment`" commit might be controversial. It replaces the `Policy` enum with a `struct Policy<Pk> { fragment: Fragment<Pk> }` where `Fragment` is basically the same as the old enum.

The reason I did this was that I wanted to cache the serialized Simplicity expression, or at least its CMR, in the new `Policy` struct. But then I realized that I couldn't do that unless I knew how to turn my `Pk` type into a `Value`, which requires a `ToPublicKey` bound that I'd rather avoid except when actually serializing the program.

So I dunno, I guess we could compute and store resource bounds or something. But in all the code I've queued up to PR, I'm not caching anything so this commit is basically just churn.

Let me know and I can remove it.

src/policy/satisfy.rs
```diff
@@ -112,48 +112,30 @@ impl<Pk: MiniscriptKey + PublicKey32 + ToPublicKey> Policy<Pk> {
...
             }
-            Fragment::Or(sub_policies) => {
-                assert_eq!(
-                    2,
-                    sub_policies.len(),
-                    "Disjunctions must have exactly two sub-policies"
-                );
-
+            Fragment::Or { left, right } => {
                 // TODO: Replace std::u64::MAX in MSRV >=1.43.0
```
uncomputable, line 126: Let's fix these TODOs as well: https://gist.github.com/uncomputable/68e1b3c8842c9d4a38f2ef4d97219293

src/policy/mod.rs
```diff
@@ -25,14 +25,116 @@
...
+/// Policy that expresses spending conditions for Simplicity.
+///
+/// The policy can be encoded directly as a Simplicity program which can be committed to the
+/// blockchain; when finalized, it can executed on the Bit Machine.
+///
+/// Policies can be normalized and are amenable to semantic analysis.
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
+pub struct Policy<Pk: MiniscriptKey> {
+    fragment: ast::Fragment<Pk>,
+}
```
uncomputable, line 54: 04bbd20406315e83722fb7012c0e59b72fed9b51: Like @apoelstra already said, this doesn't add anything and makes the code more complex. I would remove the commit and keep it on the back burner in case we find data to cache.

uncomputable (changes requested):
5d2ae8a99363ef4967aa56b4b88d2c29dd0dc699: Let's remove the fragment encapsulation commit

src/policy/satisfy.rs
apoelstra: Lol, yeah, I'll add a commit here.

I'm sure I fixed these *somewhere* but I guess it didn't make it into this PR.

src/policy/mod.rs
apoelstra: Sounds good, will do.

apoelstra:
Removed the encapsulation module, and rebased.

PR #141, opened 2023-06-25
bit machine: get rid of a bunch of dead code

apoelstra:
This eliminates a ton of old unused code (though not all of it; much of `ElementsEnv` is still unused for example, but it seems like there may be value in keeping it around because it provides accessors to transaction data for future satisfiers, even if the current one doesn't need so much information). It also tightens up the encapsulation of the bit machine module, and rearranges some files.


apoelstra:
@uncomputable this is a big diff but the majority of it is deleting and moving code so I think it shouldn't be too bad to review.

I think we should get this in before you start working on doing any anti-DoS checks, because the current bit machine module is pretty hard to read.

src/jet/mod.rs
```diff
@@ -137,6 +137,7 @@ pub trait Jet: Copy + Eq + Ord + Hash + std::fmt::Debug + std::fmt::Display {
                 c_writeBit(&mut a_frame, bit);
             }
         }
+        mac.back(src_ty_bit_width);
```
uncomputable: 0e0e13f287052b18d4e8e09e94124f338781b7f3: Is that the semantics of jets? They peek their input and don't advance the read cursor? This doesn't seem to be documented in the tech report (the jet chapter is empty). All I can see is that the test from the next commit fails on master, and that somehow the other tests are unaffected, which is surprising. We probably used `1 → A` jets almost all of the time.

uncomputable:
This PR basically removes all remnants of implementing simplicity in pure Rust (bit machine, jets). This is a little bit sad, but our current approach focuses on the FFI. We want to get the library in shape, and most importantly we want to agree with the C consensus implementation. If it makes sense, we can add a pure-Rust implementation later and test it against the FFI.

uncomputable (approved):
ACK 2430ba603b6fcecd4254ffd491a0aaeeb63c4d9c

src/jet/mod.rs
apoelstra: Literally nothing advances the read cursor in Simplicity and leaves it advanced.

PR #142, opened 2023-06-26
Eliminate recursion from `value` and from most of `types`

apoelstra:
This gets rid of almost all the recursion in the library. In particular, there should be no remaining methods which recurse on `Value` nor on `Type` with the exception of the union-bound stuff, which I didn't do because this PR was already getting too big and I was running out of weekend and probably need to get back to the human-readable encoding.

Basically, it introduces a new pre-order iterator to dag.rs and then implements `DagLike` for a ton more structures. This often lets us get rid of recursion by just handing off to the pre-order iterator, and the resulting code is even smaller and easier to read.

Unfortunately refactoring the recursion in the various `types` functions required a nontrivial refactoring of the module. But I think the result is significantly easier to understand, and since I did the refactoring as a series of small-ish steps, hopefully not too bad to review. Essentially, I moved the union-bound algorithm into its own module, taking the worst of the mutex-manipulation with it, then I was able to fold `Constraint` into `Bound` (which I usually manipulate from behind a new `MutexBound` type which checks that we aren't editing bounds on complete types); then I was able to move the occurs-check stuff out of the data structures and into a local hashset in the `finalize` algorithm.

Aside from eliminating many stack-overflow vectors, this PR also:

* Exposes free variables and their names in the public API; previously these were not accessible except maybe by parsing `Debug` output
* Exposes the ability to iterate over types, values, etc., in pre-order and post-order.
* Significantly reduces the `Debug` output of `types::Final` by only displaying the top-level TMR and bitwidth, and showing a compact string representation of the types.
* Speeds up the unit tests by around 4x, by short-cutting unification of complete types with incomplete ones.
* FIxes a race condition in `Types::finalize` where we might incorrectly trigger the occurs check if the user is finalizing the same types at the same time from multiple threads. (This was a consequence of us storing the `occurs_check` boolean in the types themselves.)
* Fixes a bug where type unification still happens even when errors occur. So you can force inference to succeed by trying it, ignoring the error, and trying again.

This started as an exploratory weekend project and kinda got away from me, but the result is a significant improvement so I'd like to keep it.

### DIFF
--- a/src/dag.rs
+++ b/src/dag.rs
@@ -375,6 +375,14 @@ pub trait DagLike: Sized {
         .map(PostOrderIterItem::unswap)
     }
 
+    /// Obtains an iterator of all the nodes rooted at the DAG, in pre-order.
+    fn pre_order_iter<S: SharingTracker<Self> + Default>(self) -> PreOrderIter<Self, S> {
+        PreOrderIter {
+            stack: vec![self],
+            tracker: Default::default(),
+        }
+    }
+
     /// Obtains an iterator of all the nodes rooted at the DAG, in post order.
     ///
     /// Each node is only yielded once, at the leftmost position that it
@@ -943,5 +951,47 @@ impl<D: DagLike, S: SharingTracker<D>> PostOrderIter<D, S> {
             f.write_str("\n")?;
         }
         Ok(())
+    }
+}
+
+/// Iterates over a DAG in _pre order_.
+///
+/// Unlike the post-order iterator, this one does not keep track of indices
+/// (this would be impractical since when we yield a node we have not yet
+/// yielded its children, so we cannot know their indices). If you do need
+/// the indices for some reason, the best strategy may be to run the
+/// post-order iterator, collect into a vector, then iterate through that
+/// backward.
+#[derive(Clone, Debug)]
+pub struct PreOrderIter<D: DagLike, S: SharingTracker<D>> {
+    /// A stack of elements to be yielded. As items are yielded, their right
+    /// children are put onto the stack followed by their left, so that the
+    /// appropriate one will be yielded on the next iteration.
+    stack: Vec<D>,
+    /// Data which tracks which nodes have been yielded already and therefore
+    /// should be skipped.
+    tracker: S,
+}
+
+impl<D: DagLike, S: SharingTracker<D>> Iterator for PreOrderIter<D, S> {
+    type Item = D;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // This algorithm is _significantly_ simpler than the post-order one,
+        // mainly because we don't care about indices.
+        while let Some(top) = self.stack.pop() {
+            // Only yield if this is the first time we have seen this node.
+            if self.tracker.record(&top, 0, None, None).is_none() {
+                // If so, mark its children as to-yield, then yield it.
+                if let Some(child) = top.right_child() {
+                    self.stack.push(child);
+                }
+                if let Some(child) = top.left_child() {
+                    self.stack.push(child);
+                }
+                return Some(top);
+            }
+        }
+        None
     }
 }
--- a/src/dag.rs
+++ b/src/dag.rs
@@ -350,6 +350,19 @@ pub trait DagLike: Sized {
         }
     }
 
+    /// Number of children that the node has.
+    ///
+    /// This has a default implementation which simply inspects the children, but
+    /// in some cases it may be more efficient for implementors to implement this
+    /// manually.
+    fn n_children(&self) -> usize {
+        match self.as_dag_node() {
+            Dag::Nullary | Dag::Witness => 0,
+            Dag::Unary(..) => 1,
+            Dag::Binary(..) | Dag::Disconnect(..) => 2,
+        }
+    }
+
     /// Obtains an iterator of all the nodes rooted at the DAG, in right-to-left post order.
     ///
     /// An ordinary post-order iterator yields children in the order
@@ -379,6 +392,25 @@ pub trait DagLike: Sized {
     fn pre_order_iter<S: SharingTracker<Self> + Default>(self) -> PreOrderIter<Self, S> {
         PreOrderIter {
             stack: vec![self],
+            tracker: Default::default(),
+        }
+    }
+
+    /// Obtains a verbose iterator of all the nodes rooted at the DAG, in pre-order.
+    ///
+    /// See the documentation of [`VerbosePreOrderIter`] for more information about what
+    /// this does. Essentially, if you find yourself using [`Self::pre_order_iter`] and
+    /// then adding a stack to manually track which items and their children have been
+    /// yielded, you may be better off using this iterator instead.
+    fn verbose_pre_order_iter<S: SharingTracker<Self> + Default>(
+        self,
+    ) -> VerbosePreOrderIter<Self, S>
+    where
+        Self: Clone,
+    {
+        VerbosePreOrderIter {
+            stack: vec![PreOrderIterItem::initial(self, None)],
+            index: 0,
             tracker: Default::default(),
         }
     }
@@ -978,7 +1010,7 @@ impl<D: DagLike, S: SharingTracker<D>> Iterator for PreOrderIter<D, S> {
 
     fn next(&mut self) -> Option<Self::Item> {
         // This algorithm is _significantly_ simpler than the post-order one,
-        // mainly because we don't care about indices.
+        // mainly because we don't care about child indices.
         while let Some(top) = self.stack.pop() {
             // Only yield if this is the first time we have seen this node.
             if self.tracker.record(&top, 0, None, None).is_none() {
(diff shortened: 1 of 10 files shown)


apoelstra:
Will fix the 1.48.0 failure tomorrow. Will require I make the syntax a little bit more annoying. The PR is otherwise ready to review.

src/value.rs
```diff
@@ -235,28 +239,58 @@ impl fmt::Debug for Value {
...
+                    parent_stack.push(Parent::Sum);
+                }
+                Value::SumR(..) => {
+                    f.write_str("1")?;
+                    parent_stack.push(Parent::Sum);
+                }
+                Value::Prod(..) => {
+                    f.write_str("(")?;
+                    parent_stack.push(Parent::ProdL);
                 }
```
uncomputable: 522bf157cf79124b60f7bc346eaa9ee89b520471: This implementation is awfully bit-centric. `Left(Sum(Unit, Unit))` is rendered as `0(ε,ε)` for example. This is an artificial example, but don't we have these kinds of non-bit values in front of case expressions sometimes? I would render what we can as bits (or even bytes) and render the rest as `L` (left) and `R` (right). Since `ε` is usually the empty string, maybe we should use a different symbol for unit, like `●`.

We could open an issue and tackle this later.

src/dag.rs
```diff
@@ -350,6 +350,19 @@ pub trait DagLike: Sized {
         }
     }
 
+    /// Number of children that the node has.
+    ///
+    /// This has a default implementation which simply inspects the children, but
+    /// in same cases it may be more efficient for implementors to implement this
```
uncomputable, line 356: 87aadf914254570e099522f452942ce8e1d82f0a: typo: "some cases"

src/value.rs
apoelstra: This is exactly what the existing code does, and it matches what the Bit Machine does internally.

Happy to open an issue to change it but it's separate from this PR.

src/dag.rs
```diff
@@ -995,3 +1027,124 @@ impl<D: DagLike, S: SharingTracker<D>> Iterator for PreOrderIter<D, S> {
...
+    /// should be skipped.
+    tracker: S,
+}
+
+impl<D: DagLike + Clone, S: SharingTracker<D>> Iterator for VerbosePreOrderIter<D, S> {
+    type Item = PreOrderIterItem<D>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        // This algorithm is _significantly_ simpler than the post-order one,
+        // mainly because we don't care about indices.
```
uncomputable, line 1068: 87aadf914254570e099522f452942ce8e1d82f0a: copy-paste error: the verbose pre-order iterator does care about indices

apoelstra: Should probably say "child indices" in both cases.

src/dag.rs
```diff
@@ -995,3 +1027,124 @@ impl<D: DagLike, S: SharingTracker<D>> Iterator for PreOrderIter<D, S> {
...
     }
 }
+
+/// Iterates over a DAG in _pre order_, yielding extra state changes.
+///
+/// Unlike the regular pre-order iterator, this one may yield each node
+/// multiple times: once in pre-order and then again after each of its
+/// children has been yielded. The different cases can be distinguished
+/// by inspecting the [`PreOrderIterItem::n_children_yielded`] and
+/// [`PreOrderIterItem::is_complete`] fields of the yielded item.
```
uncomputable, line 1037: It might be clearer to call this a "verbose pre-order" where first the parent, then the left child, then the parent, then right child, and finally the parent are yielded.

uncomputable:
Checkpoint 1ecadac72bdb0d0d12bbeb106c31e56ed54c4f83: Waiting until CI passes and this PR is finalized.

apoelstra:
Addressed comments, rebased on `master` after #140 was merged, updated block comments in the `types` module, and found & fixed a new bug in the type inference logic.

Ready for review.

apoelstra:
Drafting again because I was able to break this with the fuzzer.

apoelstra:
This is a neat program:

```
d482848704080000003131313131313131313131313131313131313131313131313131313131313131313198310000000010d1d1d100e38e38e38e38e3ffffffffffffff0fd1d1d1d1d1aaaeffffffffffffff01ffffffaaab3800380000000400000003000400325e31383400000000000000000000621886611886
```

I haven't figured out how to decode it yet. The C code rejects it with an occurs-check error. On `master` it is rejected but the error deadlocks when printing it. If I try to run it through the latest compiling version of my human-readable tool, it deadlocks trying to print out the type of something.

With this PR I get a stack overflow in the union-bound algorithm.

Given that the behavior on master is also completely (though differently) broken, I don't think we need to hold up this PR over it. But I will try to fix it.

apoelstra:
Ok, so the "infinite loop when printing occurs check" problam I think is independent; we need to drop the `incomplete_bound` varible from the `OccursCheck` error **or** fix our printing code to limit its depth. But the value we're putting into there is inherently an infinitely-sized type so we're going to have trouble.

The bug I'm seeing I've been able to isolate to:
```
    #[test]
    fn occurs_check_3() {
        let iden = Arc::<ConstructNode<Core>>::iden();
        let injr = Arc::<ConstructNode<Core>>::injr(&iden);
        let pair = Arc::<ConstructNode<Core>>::pair(&injr, &iden).unwrap();
        let drop = Arc::<ConstructNode<Core>>::drop_(&pair);

        let case1 = Arc::<ConstructNode<Core>>::case(&drop, &drop).unwrap();
        let case2 = Arc::<ConstructNode<Core>>::case(&case1, &case1).unwrap();

        let comp1 = Arc::<ConstructNode<Core>>::comp(&case2, &case2).unwrap();
        let comp4 = Arc::<ConstructNode<Core>>::comp(&comp1, &case1).unwrap();

        comp4.finalize_types_non_program().unwrap_err();  // occurs check
    }   
```
This test passes on master but fails on this PR.

apoelstra:
Ok, I have to go to bed, but I think the code in `master` is "getting lucky". See https://github.com/BlockstreamResearch/rust-simplicity/blob/master/src/types/mod.rs#L608 -- you can see that this swaps `x_lock` and `y_lock` but forgets to swap `x_root` and `y_root`. You can see the correct swapping pattern a few lines above.

When I correct this on master, I get a deadlock during type inference "as expected". Though it's a deadlock in a different place than where this PR's code infinite-loops.

I've found some simplifications to all this logic which I can push tomorrow but I still haven't resolved the infinite recursion.

apoelstra:
Ok, resolved the infinite recursion. In my `unify` method I had a check to see if we were trying to unify two elements which were already in the same set, but I was doing this check incorrectly (comparing pointers to `UbElement`, which have no particular reason to be equal to each other, rather than the pointers *inside* the `UbElement`s which will definitely be equal for the representatives for two equal sets).

With some help from @roconnor-blockstream I was able to understand why this algorithm won't infinitely recurse, even with infinite types, assuming that it is implemented correctly. There's a fairly simple inductive argument: we start with some number of disjoint type variables (bounded above by 4 * the number of Simplicity nodes, or something like that), and then every call to `unify` reduces the set of disjoint variables by one. So if we call `unify` enough times, eventually every single variable will be unified with each other, and then the equality check will mean that further `unify` calls will do nothing at all.

Anyway, if the equality check itself is wrong, obviously this reasoning doesn't hold :).

Will let the fuzzer go for another few hours before updating the PR. I'm happy to see:
* There are no more "timeout" fuzz iterations. I had previously assumed these were programs that bypassed the DoS checks we hadn't implemented and legitimately took a long time to run, but now I believe they were deadlocking our old type inference logic. This no longer occurs.
* Relatedly, we can fuzz 3-4 times faster than we used to be able to.

src/types/mod.rs
```diff
@@ -368,66 +365,85 @@ impl Type {
...
+                            // A unit might match a Bound::Free(..) or a Bound::Complete(..),
+                            // and both cases were handled above. So this is an error.
+                            (CompleteBound::Unit, _) => Err(bind_error()),
+                            (
+                                CompleteBound::Product(ref comp1, ref comp2),
+                                Bound::Product(ref ty1, ref ty2),
+                            )
+                            | (
+                                CompleteBound::Sum(ref comp1, ref comp2),
+                                Bound::Product(ref ty1, ref ty2),
```
uncomputable: 969b1a64553b93e07c3bf9618fd01dc0e6c6ef9f: Unifying sums with products is a bug that is corrected in a later commit. Amazing that no unit tests fail.

src/types/mod.rs
```diff
@@ -368,66 +365,85 @@ impl Type {
...
+                        } else {
+                            Err(bind_error())
+                        }
+                    }
+                    // Binding an incomplete to a complete type requires recursion.
+                    (Bound::Complete(ref complete), incomplete)
+                    | (&incomplete, Bound::Complete(ref complete)) => {
+                        match (complete.bound(), incomplete) {
+                            // A unit might match a Bound::Free(..) or a Bound::Complete(..),
+                            // and both cases were handled above. So this is an error.
```
uncomputable: 969b1a64553b93e07c3bf9618fd01dc0e6c6ef9f: This means a `CompleteBound::Unit` can only be bound to a `Bound::Free` or a `Bound::Complete(CompleteBound::Unit)`, both cases were handled above, right?

apoelstra: Yeah, though I could maybe make the comment clearer. The `Bound::Free` cases are obviously handled in earlier match arms.

The `Bound::Complete` case is "handled" in the sense that the only valid thing to unify with a `Complete(Unit)` is another `Complete(Unit)`, and we handled the `Bound::Complete, Bound::Complete` in the previous match arm.

src/types/mod.rs
```diff
@@ -160,98 +163,121 @@ impl std::error::Error for Error {}
...
-    constraint: Constraint,
-    /// Used during finalization; whether this type has been seen before while
-    /// traversing the type tree. If it has, it means we have a type containing
-    /// itself, which is illegal. This is referred to as the "occurs check".
-    occurs_check: bool,
-    /// The rank of the type which provides an ordering within disjoint sets
-    /// during the union-bound type inference algorithm.
-    rank: usize,
-}
+mod bound_mutex {
```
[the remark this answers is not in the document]
apoelstra: It's a submodule so you can't directly access `inner`. It seemed too minor to give it its own file.

We could merge it into the parent but then it's just coding discipline preventing us from trying to directly lock the mutex.

src/types/final_data.rs
```diff
@@ -42,16 +42,104 @@ pub enum CompleteBound {
 
 /// Data related to a finalized type, which can be extracted from a [`super::Type`]
 /// if (and only if) it is finalized.
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
```
uncomputable: 9964c3ed8c229c85859f5acfbe7af164a9a83048: 4x speedup due to fewer `String` allocations and instant `Eq` and `Ord` computations (as opposed to recursive ones)?

apoelstra: The 4x speedup is because we compare TMRs of complete bounds during unification, instead of re-recursing into their incomplete types.

apoelstra: The `derive`s are removed because they're otherwise recursive, but there's not really any perf improvement (plausibly there would be a slight regression because of the extra indirection and heap allocations).

src/types/mod.rs
```diff
@@ -320,17 +308,69 @@ impl Bound {
...
+                Dag::Binary(ty1.bound.root().get(), ty2.bound.root().get())
+            }
+        }
+    }
+}
+
+/// Source or target type of a Simplicity expression
+///
+/// This is just a mutex holding the actual data, which defined by the
+/// `TypeInner` type.
```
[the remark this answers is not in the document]
apoelstra: I don't know what commit you're referring to (this code is not in 148dc06) or what code you're referring to, because Github is only showing two lines of context.

src/types/mod.rs
```diff
@@ -459,31 +457,44 @@ impl Type {
...
+        let mut occurs_check = HashSet::new();
+        for data in bound.verbose_pre_order_iter::<NoSharing>() {
+            if data.is_complete {
+                occurs_check.remove(&(data.node.as_ref() as *const _));
+            } else if data.n_children_yielded == 0
+                && !occurs_check.insert(data.node.as_ref() as *const _)
+            {
+                return Err(Error::OccursCheck);
+            }
         }
```
uncomputable: e690bde94cea795e028e1414f0d61d7c30be42c1: I would love to use the TMR for this, but we don't have finalized types yet. I guess we have to do pointer comparisons...

Also, is `occurs_check.remove(...)` an optimization to keep the hash set small?

uncomputable (commented):
148dc06c4ed0a45b3cfef6fa543724ae4e804772 Phew, made it to the end

src/types/final_data.rs
uncomputable: I wrote an [alternative commit](https://github.com/uncomputable/rust-simplicity/commit/576af55ed1359bada7dcbb6611beb683dc607fc7) that only changes `Debug` and `Display`. The speedup is there.

src/types/mod.rs
apoelstra: No, it's because types are allowed to be repeated, they just can't be nested. So you need to remove types from the set once you've recursed out of them.

src/types/mod.rs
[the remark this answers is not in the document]
apoelstra: I cannot find the text "This is just" in 18035d0 either.

In any case, I'll add a commit which updates the doccomment on `Type`, which I infer is what you're referring to.

src/types/final_data.rs
uncomputable: When I go one commit before the alternative commit, the tests are slow. After applying the commit, there is a speed up. I think the commit message was right. Anyway, not too important.

apoelstra: Huh, yeah, I think you're right. I guess I was sloppy pinpointing where the speedup came from. (FWIW I only even noticed it several commits later, then used `git rebase -x` to run the tests on every commit, and see what the timing was. I was not trying to speed anything up in this PR.)

Agreed, it's not too important, so I won't change the message back, but I wish I hadn't changed it now.

apoelstra: And looking at the code now, I think you're right that this is about the string allocations.

uncomputable (approved):
ACK ff7ec7444cd6cf901aef638a7a530f071bad1491

PR #143, opened 2023-06-26
Policy 2: update compiler and satisfier code

apoelstra:
This commit introduces a new `WitnessNode` type which is analogous to `ConstructNode` but supports attaching witnesses and pruning missing witnesses. It replaces the `compiler` module with a new `serialize` module which takes a `Policy` and serializes it to Simplicity, either using a `ConstructNode` converted to `CommitNode` (at commit time) or a `WitnessNode` converted to `RedeemNode` (at redeem time).

It leaves some undone work, specifically:
* We need to support shrinking witnesses, which can't happen with our current fragments, but in theory can happen with future ones. This happens when after pruning, some witness data is no longer needed.
* We need to support a final pruning pass where we run the bit machine to prune any un-taken branches that weren't recognized as prunable during satisfaction.
* We need to support a correct weight function (need some new bounds computations first)
* Relatedly we should figure out what bounds we can put on `WitnessNode` without needing to finalize it first.
* We need to support the `asm` fragment (but we probably want to introduce the human-readable encoding first).

But I think it's fine to leave all these things for future PRs. With this PR, the next one will be able to do some final cleanups and remove the old `core` module :).


src/policy/descriptor.rs
```diff
@@ -12,6 +11,16 @@ use elements_miniscript::{MiniscriptKey, ToPublicKey};
 use std::fmt;
 use std::str::FromStr;
 
+pub trait UnspendableKey: MiniscriptKey {
+    fn unspendable() -> Self;
+}
```
uncomputable, line 16: ad6f865b868b7add369d26c5aa1a6195b7e8b703: The name `UnspendableKey` can lead someone to believe that all keys of this type are unspendable. Maybe `HasUnspendable` is less confusing.

src/policy/satisfy.rs
```diff
@@ -38,208 +39,135 @@ impl<'a, Pk: ToPublicKey> Satisfier<Pk> for PolicySatisfier<'a, Pk> {
...
-                    right.witness.push_front(Value::u1(1));
-                    right.witness_cost += 1;
-                    right.program =
-                        compiler::or(left.program, right.program, UsedCaseBranch::Right).ok()?;
-                    right.program_cost += 6;
-                    Some(right)
+                    super::serialize::or(&left, &right.pruned(), Some(Arc::new(Value::u1(0))))
                 }
+                .prune_and_retype()
             }
```
uncomputable, line 114: 3030b553f76c1a51e79bd21cba68ed8e987b6565: The changing types after each pruning step make this very complicated: First we recursively satisfy the left and the right child. Then, in one `match` case, we finalize both which is another tree traversal. Combining left and right in an `or` is easy. Finally we reconstruct and type the resulting tree again.

That is the downside of the constructive approach. I'm not saying I have a better proposal right now... The fact that pruning causes non-local type changes forces us to reconstruct the entire tree from scratch again and again.

I hope that `WitnessNode::finalize()` is basically as fast as `Type::finalize()`.

src/policy/satisfy.rs
```diff
@@ -38,208 +39,135 @@ impl<'a, Pk: ToPublicKey> Satisfier<Pk> for PolicySatisfier<'a, Pk> {
...
-                for i in 1..sub_policies.len() {
-                    let child_sat = get_summand(i, &mut satisfiable_children, sub_policies);
-
-                    sat.witness.extend(child_sat.witness);
-                    sat.witness_cost += child_sat.witness_cost;
-                    sat.program = compiler::thresh_add(sat.program, child_sat.program).unwrap();
-                    sat.program_cost += child_sat.program_cost + 6;
+                let k = u32::try_from(k).expect("k less than 2^32");
+                if threshold_failed {
+                    super::serialize::threshold(k, &nodes, &witness_bits).pruned()
```
uncomputable, line 152: 3030b553f76c1a51e79bd21cba68ed8e987b6565: If we already know that the threshold fails, why construct it in the first place? Could we not do `super::serialize::trivial().pruned()`?

This doesn't work for the other occurrences of `.pruned()` because these expressions come from the sub policies and are already constructed.

apoelstra: No, `trivial` has a completely different CMR than the actual threshold. We can't replace branches with unrelated ones like that.

src/policy/satisfy.rs
apoelstra: There are two "complicated" steps here, neither of which are really necessary:

* We call `.finalize` (which does indeed only finalize types, it doesn't do much other work) in order to get the cells bound. But in theory we could get this information directly from the `WitnessNode` without finalizng, we just haven't implemented that.
* We call `prune_and_retype` which is just an optimization at this point, to get us better estimates when making decisions in other `or`s and `thresh`s.

src/node/witness.rs
```diff
@@ -0,0 +1,329 @@
...
+                left: &Arc<WitnessNode<J>>,
+                right: &Arc<WitnessNode<J>>,
+            ) -> Result<Hide, Self::Error> {
+                // If either child is marked as pruned, we hide it, which will cause this
+                // case node to become an assertl or assertr, potentially reducing the size
+                // of types since there will be fewer constraints to unify.
+                //
+                // If both children are marked pruned, this function will only hide the left
+                // one. This doesn't matter; in this case the node itself will be marked as
+                // pruned and eventually get dropped.
```
uncomputable, line 102: 5b76889ffee871cb1e537cf5a2529c9400428408: I may have to run `prune_and_retype()` arbitrarily many times before I finish, right?

For every number I can construct a chain of nested `case` combinators with one `fail` at the end. In every step the lowermost `case` will be converted into `assert{l,r}` which is marked `must_prune`. The ancestor `case` combinators will not be marked because both branches must be marked `must_prune` for `case` to be marked.

This is not necessarily an issue, but we should warn users in the documentation and we should be careful about our own code.

src/policy/satisfy.rs
uncomputable: Ah, makes sense. I guess constructing the threshold is not too bad if we have already constructed the children.

src/node/witness.rs
apoelstra: I think you're confused. If both branches if every `case` are prunable then they'll all be pruned in one pass. If only the topmost one is prunable, and the others have some non-prunable branches, then they won't be. Which is correct.

src/policy/satisfy.rs
apoelstra: Maybe let's file an issue and see if we can make a `WitnessNode` constructor that produces a `must_prune` node with a specified CMR, and an inner of `Unit` ... I need to think a bit but I think that'd be fine, and would let us do this.

Or don't file an issue, and I'll try to remember to do this :). But because it's an optimization I'd like to do it after this PR, which is otherwise a bit big.

uncomputable: We need to know the CMR first. And we have to construct the expression for that, don't we?

src/node/witness.rs
uncomputable: You're right. Is it fair to say that an expression will be fully pruned after two steps? If you want to do anything useful with the `WitnessNode` you're going to finalize it to `RedeemNode` which does the pruning for you. So I suppose the expression is "fully pruned" after one step.

uncomputable:
In a later PR, we should add a method that converts a `Policy` into a `CommitNode`. Basically we port `policy::compiler` to the new node types and possibly rename the module.

uncomputable (approved):
ACK 3030b553f76c1a51e79bd21cba68ed8e987b6565

src/policy/satisfy.rs
apoelstra: Ah, yes, you're right.

Though we could cache it by changing `Policy` to be a wrapper type around a `Fragment` enum ;)

apoelstra: Actually, that would be quite hard/annoying to do, because it would require `ToPublicKey` to compute the cached CMR, and we don't assume `ToPublicKey` until we're actually trying to do a satisfaction..

src/node/witness.rs
apoelstra: Right -- `prune_and_retype` will do all the "explicitly-marked" pruning, in one step, and you should do this whenever you might prune stuff for optimal results. But to actually use the result, you want to do the "real" pruning, which also happens in one step, during the conversion to `RedeemNode`.

PR #144, opened 2023-06-27
CPU cost

uncomputable:
Adds a CPU cost to the node bounds of a redeem node. We can detect programs that are too expensive, which is useful during fuzzing.

Depends on https://github.com/BlockstreamResearch/simplicity/pull/161.

The cost of a disconnect combinator is subject to change: https://github.com/BlockstreamResearch/simplicity/issues/159
For now, this PR uses the old definition.


src/analysis.rs
```diff
@@ -12,7 +12,64 @@
...
+        (self.0 + 999) / 1000
+    }
+
+    /// Return the cost of a type with the given bit width.
+    pub const fn from_type(bit_width: usize) -> Self {
+        Cost(bit_width)
+    }
+
+    /// Wrap the cost of a jet.
+    pub const fn from_jet(cost: usize) -> Self {
```
apoelstra, line 83: In c45f8d3cc6d26c6435b3a3e06b4dc72d9a79e3b7:

You should call this `from_milliweight` and update the Simplicity PR accordingly.

uncomputable:
The PR now uses the generated jets from the main repo.

src/analysis.rs
```diff
@@ -57,6 +83,22 @@ impl Cost {
...
+    pub fn is_consensus_valid(&self) -> bool {
+        self <= &Self::CONSENSUS_MAX
+    }
+
+    /// Return whether the cost is within the budget of
+    /// the given transaction input.
+    #[cfg(feature = "elements")]
+    pub fn is_budget_valid(&self, txin: &elements::TxIn) -> bool {
+        let budget = txin.witness.script_witness.iter().map(|el| el.len()).sum();
+        self.to_weight() <= budget
```
apoelstra, line 100: In 00683984959400de3dd8c4902bc4940c5ded159a:

You should replace this with `txin.witness.serialized_len()` + 50_000 to match the consensus code at https://github.com/ElementsProject/elements/blob/simplicity/src/script/interpreter.cpp#L3301

uncomputable: Good find. I wrote about the validation offset in the documentation, but I forgot it in this method.

uncomputable: I pushed an update.

src/analysis.rs
```diff
@@ -96,7 +99,12 @@ impl Cost {
     /// the given transaction input.
     #[cfg(feature = "elements")]
     pub fn is_budget_valid(&self, txin: &elements::TxIn) -> bool {
-        let budget = txin.witness.script_witness.iter().map(|el| el.len()).sum();
+        let mut witness_stack_encoding = Vec::new();
+        txin.witness
+            .script_witness
+            .consensus_encode(&mut witness_stack_encoding)
+            .expect("writing to vector never fails");
```
apoelstra, line 106: In a314da1acd8015e94b04e7fe9f7fd02d24f16368:

It does not seem reasonable to allocate a vector here. Why not use `Witness::serialized_len` which I even checked was available in rust-bitcoin 0.29.

uncomputable:
Rebased.

apoelstra (approved):
ACK 965a8220296e4c2375d0c93e4952551bf617e678

PR #145, opened 2023-06-28
Remove `core` module

apoelstra:
This PR completes the node overhaul :).

The first commit changes the `compiler` module so that it exposes a single function that just wraps `serialize::serialize_no_witnesses`. Should we just delete the module and move its unit tests to `serialize`?


uncomputable:
Let's move everything from `compiler` to `serialize`. The actual compiler part is almost a one liner. (We could rename it to `commit` and add a type finalization.) The unit tests more or less generalize to `WitnessNode` as well. At least, if everything works for `ConstructNode`, it is highly unlikely that something breaks for `WitnessNode`.

apoelstra:
I'll add a commit at the end.

src/node/redeem.rs
```diff
@@ -177,18 +177,6 @@ impl<J: Jet> RedeemNode<J> {
...
-        arrow: FinalArrow,
-        inner: Inner<Arc<Self>, J, Arc<Value>>,
-        inner_cached: Inner<&Arc<RedeemData<J>>, J, Arc<Value>>,
-    ) -> Result<Self, types::Error> {
-        Ok(RedeemNode {
-            cmr: Cmr::unit(),
-            inner,
-            data: Arc::new(RedeemData::new(arrow, inner_cached)),
-        })
-    }
```
uncomputable, line 190: 0da3e7baecc80e9c37a755f5bf9038324d3f1f3d: Thanks for remembering :)

apoelstra:
@uncomputable I just rebased this (I needed to because the "remove `Witness` from `dag`" commit stopped compiling since #142 added some new uses of `dag`.)

uncomputable (approved):
ACK d192d87b0adac2529002f0ccf3d9bf1e72e2e36c

PR #147, opened 2023-06-29
Cleanup

uncomputable:
This removes the obsolete `test_progs` and `simplicityC_test_inputs` that are not used anywhere.


apoelstra:
Agreed. The testprogs were there because in the old days we couldn't make our own test vectors and had to use Russell's; but now we not only have a much larger corpus of test vectors, we also have the fuzzer which is able to compare billions of examples between the C and Rust implementation.

As for the simplicityC_test_inputs, I believe these originate from gmaxwell doing fuzz testing on the C library very many years ago, and finding a reduced set of examples that had significant coverage. Sadly, these are no longer useful to us beacuse the encoding has changed.

apoelstra (approved):
ACK 146c840e139b69b907f04785a6d311d422f52a36

sanket1729:
@uncomputable @apoelstra these tests are not obselete. Only the testdata was obsolete. This was the only unit test that actually checked C and rust programs computed the cmr/amrs etc correctly. 

```rust
#[cfg(test)]
#[cfg(feature = "test-utils")]
mod tests {

    use simplicity_sys::c_jets::exec_ffi;

    use crate::BitIter;
    use crate::jet::Elements;
    use crate::merkle::common::MerkleRoot;
    use crate::RedeemNode;

    // // TODO: check AMR(if we implement it)
    // The new IMR also has type commitments, so we don't really need to check AMR
    fn check_merkle_roots(test: &exec_ffi::TestData) {
        let mut bits = BitIter::new(test.prog.iter().cloned());
        let prog = RedeemNode::<Elements>::decode(&mut bits).unwrap();
        assert_eq!(prog.cmr.into_inner(), test.cmr);
        // assert_eq!(redeem.amr.into_inner(), test.amr);
        assert_eq!(prog.imr.into_inner(), test.imr);
    }

    #[test]
    fn progs_cmr() {
        let hash_block = exec_ffi::hash_block_test_data();
        let schnorr0 = exec_ffi::schnorr0_test_data();
        let schnorr6 = exec_ffi::schnorr6_test_data();
        check_merkle_roots(&hash_block);
        check_merkle_roots(&schnorr0);
        check_merkle_roots(&schnorr6);
    }
}
```

apoelstra:
@sanket1729 we have tests in simplicity-sys/src/tests.c that check all of the example programs included in C, and we have a fuzz test that we've run ~1 trillion iterations on.

The old tests seemed to be redundant with those, and they weren't run, and wouldn't have passed (because their data was obsolete).

sanket1729:
> tests in simplicity-sys/src/tests.c 

These tests don't actually check against the rust implementation. They only check C implementation produces the correct result.

> The old tests seemed to be redundant with those, and they weren't run, and wouldn't have passed (because their data was obsolete).

The testdata was not being used at all. Only the 20 lines in mod.rs were used. We were linking to C FFI data as shown in the code snippet above. There are currently no unit tests that test whether the C and rust implementations match on cmr/amr etc.

Relying on the fuzzer to detect these is unreliable and I think having these tests is immense value given that it is only 20 LoC and provides quick feedback as a unit test.

apoelstra:
For sure -- we should add those back in. Sorry about that.

sanket1729:
No worries. I will raise a PR for this.

uncomputable:
I am surprised that the fuzzer is supposed to be unreliable. There are only a handful of combinators. If there is a false Merkle root computation for any of those, it should be able to detect it.

sanket1729:
I guess to better to rely on a deterministic test rather an randomized fuzz algorithm? I think it is unreliable in the sense it's hard to know what exactly the fuzzer tests or how long do we need to run it to gain confidence. 

We have been running the miniscript descriptor fuzzer for over 3 years, but it still could not produce a valid looking descriptor where the `MiniscriptKey` is not a part of standard ascii set for a long time. Our fuzzer finally found this issue last month for a bug that has been present since the first iteration of rust-miniscript. 

Even if we have a perfect fuzzer, having a unit test fail provides quicker feedback in code development. I almost never run the fuzzer locally before I make a PR.  Plus, lets make good use of the new FFI :)

apoelstra:
@sanket1729 almost every single unit test that checks CMRs against target values came from the fuzzer. If you find more failures, please add them as unit tests!

apoelstra:
And if you want to extend the existing `assert_program_deserializable` unit test helper to *also* test the C code, that would be nice as well :)

PR #148, opened 2023-06-30
Easy fuzz

### LINKED ISSUE #117: Easier local fuzz
Local fuzzing is a bit of a pain at the moment. `travis-fuzz.sh` is optimized for CI. We should write a README and some scripts that make it easier to run / debug local fuzz tests.

uncomputable:
Copies and adapts the fuzzer setup from rust-bitcoin. Updates our fuzzing CI. Fixes #117.

There are some open problems (see wip commits):

## Readme

- [x] Should we remove the section on weak cryptography if it breaks our Merkle roots? We use a fork of bitcoin-hashes so that part is already impossible to weaken.
- [x] `HFUZZ_BUILD_ARGS='--features honggfuzz_fuzz' cargo hfuzz run <target>` doesn't work on my machine. I have to drop the `HFUZZ_BUILD_ARGS` part.
    - `error: Package simplicity-fuzz v0.0.1 does not have the feature honggfuzz_fuzz`
    - Is that outdated documentation, or do I have to change my local setup?
- [x] I cannot reproduce failures as shown. When a fuzz test fails it merely increments the error counter and continues. The files in `hfuzz_workspace/` look like the expected output, but there is **no hex value** below the == line

## Generate CI

- [x] Anything that needs to be changed to work with Simplicity?
- [x] Test unit tests in fuzz crate? _(Was disabled by generate-fuzz.sh)_


apoelstra:
> Should we remove the section on weak cryptography if it breaks our Merkle roots? We use a fork of bitcoin-hashes so that part is already impossible to weaken.

Yes, we should remove it. When the ability to weaken the crypto returns we still won't want to do it.

>￼ HFUZZ_BUILD_ARGS='--features honggfuzz_fuzz' cargo hfuzz run <target> doesn't work on my machine. I have to drop the HFUZZ_BUILD_ARGS part.

Yeah, this is probably outdated in rust-bitcoin and has never applied to this crate.

>￼ I cannot reproduce failures as shown. When a fuzz test fails it merely increments the error counter and continues. The files in hfuzz_workspace/ look like the expected output, but there is no hex value below the == line

What fuzztests are failing right now? In any case, if you are just running `cargo hfuzz run` you are not going to get the nice hex output. You need to run `fuzz.sh` for this. To get the hex output you need to find the crash files in `hfuzz_workspace/<test>/SIG*` and run them through `xxd`.

uncomputable:
Fuzzing and error reports work for me now. I updated the README accordingly. _(There are no fuzz tests that fail. I artificially made them fail.)_

We could even push some of these updates to rust-bitcoin. I've separated the simplicity commits from the general ones.

I will open this for review.

uncomputable:
Please confirm that the CI looks as it should.

One nuisance that we might want to fix is that `checkReport` kills the nix shell when it exists with an error. Has this even come up in rust-bitcoin?

uncomputable:
I added `getReport` as a safe way to generate reports in the nix shell, and sorted the target in alphabetical orders.

File names with null characters and the like will not work, but those will probably be caught by the `checkWindowsFiles` check.

For me generate-files.sh is still clean. I could not reproduce your error. If this happens again, we need to debug this in more depth.

fuzz/README.md
```diff
@@ -10,15 +10,13 @@ run
 
 in this directory.
 
-To build honggfuzz, you must have libunwind on your system, as well as
-libopcodes and libbfd from binutils **2.38** on your system. The most
-recently-released binutils 2.39 has changed their API in a breaking way.
+You need honggfuzz installed on your system.
```
apoelstra, line 12: In 3ea836c9c0615fbcf2beb48d64bc7c52518af452:

This is only true on NixOS. On normal systems `cargo hfuzz` will build honggfuzz for you. I'm actually pretty surprised that you can just install it locally and things just work (if you have the right version of binutils) or just don't (otherwise).

I'm happy to have a simpler `nix-shell` invocation but we should maybe clarify that it's only needed if you get "libbfd" build errors.

We could also maybe mention that you need to run `cargo install honggfuzz` at some point, regardless of nix-shell invocation, but that the provided shellscripts will do this for you.

fuzz/README.md
```diff
@@ -104,12 +104,14 @@ FUZZ_FNAME: hfuzz_workspace/hashes_sha256/SIGABRT.PC.7ffff7c8abc7.STACK.18826d9b
 fff400610004
 ```
 
+If you don't see the summary because you fuzzed via `cargo`, run
+
+    checkReport <target>
+
```
apoelstra, line 89: In 3ea836c9c0615fbcf2beb48d64bc7c52518af452:

We should reword the original text to asy "If fuzz.sh or cycle.sh fails, it will exit..." and there here say "if you can `cargo hfuzz` manually, use `checkReport <target>` to see the report".

fuzz/README.md
```diff
@@ -104,12 +104,14 @@ FUZZ_FNAME: hfuzz_workspace/hashes_sha256/SIGABRT.PC.7ffff7c8abc7.STACK.18826d9b
...
 The final line is a hex-encoded version of the input that caused the crash. You
 can test this directly by editing the `duplicate_crash` test to copy/paste the
 hex output into the call to `extend_vec_from_hex`. Then run the test with
 
-    cargo test
-
-Note that if you set your `RUSTFLAGS` while fuzzing (see above) you must make
-sure they are set the same way when running `cargo test`.
+    RUSTFLAGS=--cfg=fuzzing cargo test
```
apoelstra, line 94: In 3ea836c9c0615fbcf2beb48d64bc7c52518af452:

This is incorrect. The old text was correct.

fuzz/README.md
apoelstra: Also, without manually running `source fuzz-util.sh` this won't work, and once you do, it'll kill your shell. I'll leave it to you whether to update it and update the README, or whether you just want to change the text to use `xxd` directly.

fuzz/fuzz_targets/c_rust_merkle.rs
```diff
@@ -55,8 +55,8 @@ fn main() {
     }
 }
 
-#[cfg(test)]
-mod unit_tests {
+#[cfg(all(test, fuzzing))]
```
apoelstra, line 58: In 2a3b18e919b54cea75383ae03ac2b83f0e002d88:

Why gate on `fuzzing`? This is super annoying.

apoelstra:
Done reviewing b50ad1c516789f64716b55529143471f4f0aff40

fuzz/fuzz_targets/c_rust_merkle.rs
uncomputable: [I did the same as in rust-bitcoin](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/fuzz/fuzz_targets/bitcoin/deser_net_msg.rs#L16).

fuzz/README.md
uncomputable: I need to add `cargo install honggfuzz` to the nix shell, actually. I thought the command checks if you have honggfuzz installed and takes that instead of building it from scratch, but I am no longer so convinced. Adding the build dependencies to the shell seems like the safe option.

fuzz/README.md
uncomputable: This has to do with the `fuzzing` feature gates on the unit tests. Those were introduced [in this commit](https://github.com/rust-bitcoin/rust-bitcoin/commit/64677282023295812c8b95212e0e5e96021040f2) to rust-bitcoin. Then [this second commit](https://github.com/rust-bitcoin/rust-bitcoin/commit/6649e1519305973a96d6d16e839711352c6044d8) added support for weak crypto where the `fuzzing` flag is disallowed. We don't have any of that in rust-simplicity, so we could stick with the original lines. If we want to be open to future changes, I would at least add a line that explains what the typical command in the typical scenario would looks like, so it's easy to copy and paste.

fuzz/fuzz_targets/c_rust_merkle.rs
apoelstra: If anything is broken by `cfg(fuzzing)` then we should fix the broken stuff, not inconvenience every user of the fuzztests ever.

fuzz/README.md
apoelstra: @uncomputable the build process is actually super confusing. When you do `cargo install honggfuzz` it installs the `cargo-honggfuzz` binary which is basically a build script. Then when you do `cargo hfuzz run` it rebuilds rust-simplicity "inside of" honggfuzz, with some extra instrumentation and a nice UI. It's the latter step where the nix-installed honggfuzz seems to be overriding something and allowing the build to succeed.

uncomputable:
Rebased and addressed the comments:

No more `fuzzing` gates, fixed README, a better nix shell that installs honggfuzz, `getReport` for generating error reports without crashing the shell.

apoelstra:
Could you rebase this so that

* the new README is in one commit (ok to split this into 3, one that copies rust-bitcoin, one that tweaks it in upstreamable ways, and one that does rust-simplicity-specific stuff)
* the new scripts are in one commit (ok to make this two, one that copies rust-bitcoin and one that tweaks it, if that makes it easier to upstream)
* Then the 'unify duplicate_crash` commit could go anywhere.
* Then deleting CI and running the the new scripts could be one commit, or two, your choice.

Right now this PR feels like it's updating the same files repeatedly in an unclear order, and it's hard to say, for any given problem, whether it's going to be fixed later or what.

In any case, `#[cfg(fuzzing)]` still appears in several places and in the README.

uncomputable:
Done. Originally I kept the commits separate for upstream cherry picks, but I feel the upstream commits will require a lot of changes, so it's not worth the effort.

fuzz/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+    ./fuzz.sh
+
+in this directory.
+
+You need to install the Rust crate honggfuzz-rs.
+
+To build the crate you must have libunwind on your system, as well as
+libopcodes and libbfd from binutils **2.38** on your system. The most
+recently-released binutils 2.39 has changed their API in a breaking way.
```
apoelstra, line 16: In 08801420c213b16598eb0d5926c7173ce6f0fd4d

This paragraph and the sentence above are confusing. The actual utility you need to install is `cargo-honggfuzz`. There is no crate called `honggfuzz-rs`; that is the name of the git repo where it comes from, if you install it from Github, which you won't if you use `cargo install`.

There's also no need to explicitly run `cargo install` because the shell script does it for you.

You also don't need libunwind or libbfd to install or build `cargo-honggfuzz`. You need these to *run* `cargo hfuzz run`, and you only need to care about this if the system-provided ones are the wrong version.

fuzz/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+    nix-shell
+
+Then run fuzz.sh as above.
+
+# Long-term fuzzing
+
+To see the full list of targets, the most straightforward way is to run
+
+    listTargetNames
```
apoelstra, line 32: Or you can just run `cargo hfuzz run x` and it will give you the list, and then you don't need to source anything.

fuzz/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+You may need to source the fuzz utils, unless you are already in the nix shell
```
apoelstra, line 34: In 08801420c213b16598eb0d5926c7173ce6f0fd4d:

We should not assume people are using nix-shell. If you want to instruct people to run bash functions then you need to start by telling them to source stuff.

fuzz/README.md
```diff
@@ -0,0 +1,110 @@
...
+ORIG_FNAME: 00000000000000000000000000000000.00000000.honggfuzz.cov
+FUZZ_FNAME: hfuzz_workspace/decode_natural/SIGABRT.PC.7ffff7e42adc.STACK.1a1d230819.CODE.-6.ADDR.0.INSTR.mov____%eax,%ebx.fuzz
+...
+=====================================================================
+fff400610004
+```
+
+If you run `cargo hfuzz` manually, run the following to see the report
+
+    checkReport <target>
```
apoelstra, line 96: In 08801420c213b16598eb0d5926c7173ce6f0fd4d:

If you are telling people to run this then they need to source the utils script first.

fuzz/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+    cargo test
+
+
+Note that if you set your `RUSTFLAGS` while fuzzing (see above) you must make
+sure they are set the same way when running `cargo test`.
+
+For example, you can enable the `fuzzing` feature as follows
+
+    RUSTFLAGS=--cfg=fuzzing cargo test
```
apoelstra, line 110: In 08801420c213b16598eb0d5926c7173ce6f0fd4d:

Nothing suggests to set the `fuzzing` flag, even in the original README, so this is confusing. You've also removed all the other RUSTFLAGS stuff so you might as well remove this too.

apoelstra:
Done reviewing 38937d17d058cf6ebbb3477808ef2f0d126bba88. The rest looks good.

uncomputable:
Addressed the comments.

fuzz/README.md
```diff
@@ -0,0 +1,109 @@
...
+This script uses the `chrt` utility to try to reduce the priority of the
+jobs. If you would like to run for longer, the most straightforward way
+is to edit `cycle.sh` before starting. To run the fuzz-tests in parallel,
+you will need to implement a custom harness.
+
+To manually run a single target indefinitely, run
+
+    cargo hfuzz run <target>
+
+You need to install cargo-honggfuzz
```
apoelstra, line 44: In d4852ef27c8720437f05bc10a4d84e04289a71d8:

I'd say

```
You may need to install cargo-honggfuzz:

    cargo install honggfuzz --no-default-features

though it may already be installed. It is installed by the shell-scripts and is available in the provided nix-shell.

If you have linker errors see the above information about libbfd and libopcodes.
```

uncomputable:
Updated the README.

apoelstra:
CI failure is unrelated serde stuff.

PR #149, opened 2023-07-02
Node refactors prior to human-readable encoding

apoelstra:
This PR does some first steps to making the `Node` types more general and powerful. The next one will similarly generalize things to allow `disconnect` nodes to have missing children (and correctly have `CommitNode` without disconnected expressions and `ReedemNode` with). Once we've done that we can start merging the human-readable encoding. Or if you get impatient we can re-order things :).

My full branch, which has some example programs etc as well as a parser, is at https://github.com/BlockstreamResearch/rust-simplicity/tree/2023-06--simplang . I will try to keep it up to date. Note the `2023-06-`. There is also a `2023-05-` branch with the same name which is badly out of date.

### DIFF
--- a/src/node/commit.rs
+++ b/src/node/commit.rs
@@ -19,8 +19,8 @@ use crate::{encode, types};
 use crate::{Amr, BitIter, BitWriter, Cmr, Error, FirstPassImr, Imr};
 
 use super::{
-    Construct, ConstructData, ConstructNode, Constructible, Converter, Inner, NoWitness, Node,
-    NodeData, Redeem, RedeemNode,
+    Construct, ConstructData, ConstructNode, Constructible, Converter, Inner, Marker, NoWitness,
+    Node, Redeem, RedeemNode,
 };
 
 use std::io;
@@ -30,7 +30,7 @@ use std::sync::Arc;
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
 pub enum Commit {}
 
-impl<J: Jet> NodeData<J> for Commit {
+impl<J: Jet> Marker<J> for Commit {
     type CachedData = Arc<CommitData<J>>;
     type Witness = NoWitness;
     type SharingId = Imr;
--- a/src/node/commit.rs
+++ b/src/node/commit.rs
@@ -28,7 +28,7 @@ use std::marker::PhantomData;
 use std::sync::Arc;
 
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
-pub struct Commit<J: Jet> {
+pub struct Commit<J> {
     /// Makes the type non-constructible.
     never: std::convert::Infallible,
     /// Required by Rust.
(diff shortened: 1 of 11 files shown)


src/human_encoding/README.md
```diff
@@ -0,0 +1,148 @@
...
+
+the name `node` is substituted by `unit` both places that it appears. We can see
+that starting from the name `main`, by recursive substitution, we obtain a single
+Simplicity expression.
+
+The following restrictions apply:
+
+* The name `main` is special and must occur once in every encoding; recursive substitution starts from this symbol.
+* No cycles are allowed; if a name occurs anywhere in the expansion of its expression, this is an error.
+* Every name must be reachable from `main`. FIXME this is currently not enforced.
```
apoelstra, line 109: Need to clarify that names can be defined multiple times, but exactly one must have an expression. (The others, if any, can add type bounds.)

src/human_encoding/README.md
```diff
@@ -0,0 +1,148 @@
...
+An ENTROPY is a VALUE whose size is between 128 and 512 bits inclusive. Internally
+it is 0-padded out to 512 bits.
+
+Finally, TYPE is
+
+* a literal `_`, indicating no type bound;
+* a NAME;
+* a literal `1`, indicating the unit type;
+* a literal `2`, indicating the bit type;
+* `2^n`, where `n` is any power of two, in decimal with no spaces or punctuation;
```
apoelstra, line 78: Should decide what exactly "any power of 2" means.

apoelstra:
Renamed `jet_word` to `const`; reserved names starting with `prim`. No other changes.

apoelstra:
Drafting while I re-architect this a bit to support decoding multiply-rooted "programs". Probably the changes for this PR will be limited to minor renaming of structs etc., but I haven't nailed things down on the parsing side.

This will probably also require changes to deal with `disconnect` but we may be able to do those after the fact.

apoelstra:
Undrafting since I have a working parser now that handles DAG forests. I'm not certain about the API here but I know this is workable, and we can tweak it.

src/node/commit.rs
```diff
@@ -64,16 +64,6 @@ pub struct CommitData<J> {
...
 
-impl<J> PartialEq for CommitData<J> {
-    // Two nodes are equal if they both have IMRs and those IMRs are equal.
-    fn eq(&self, other: &Self) -> bool {
-        self.imr
-            .zip(other.imr)
-            .map(|(a, b)| a == b)
-            .unwrap_or(false)
-    }
-}
```
uncomputable: d7021c2447a3304a818c9c2848750c2c227ca0e9: I think this is good since we might want to use `CommitNode` in hash maps and the likes.

I don't like the recursive equality check on `FinalArrow` but we can fix that later.

uncomputable: My bad; I should have looked a level deeper :) In this case, we are good.

uncomputable:
I like the changes to the generics by the way. A lot of the power of the new node and dag system comes from the fact that it is so flexible and versatile. Removing trait bounds from structures is a step in the same direction. Interesting stackoverflow discussion.

uncomputable (approved):
ACK d7021c2447a3304a818c9c2848750c2c227ca0e9

PR #151, opened 2023-07-06
node: allow `disconnect` nodes to have missing children

apoelstra:
This PR replaces the "stop code" with a single child disconnect node. When parsing `CommitNode`s, this must be used in place of disconnect (i.e. disconnects cannot have right children); when parsing `RedeemNode`s, it may *not* be used (i.e. disconnects must have right children). For `ConstructNode` and `WitnessNode` you can use either one.

Naturally something that affects the DAG structure of every node is going to be a major undertaking, but I nonetheless apologize for the size of this PR. I tried to split it into sensible commits, and where changes required a ton of mechanical changes, I tried to isolate those changes in their own commits so that the "real" changes would be in smaller ones.

Let me know fi you want me to split this up. The first 7 commits are cleanups that don't (really) have anything to do with the main PR.


src/dag.rs
```diff
@@ -1,4 +1,4 @@
-// Rust Simplicity Library
+//) Rust Simplicity Library
```
uncomputable, line 1: 047f4bcf1a7aaf41495e90a346ab6185e306af4e: typo

src/node/commit.rs
```diff
@@ -226,18 +226,19 @@ impl<J: Jet> CommitNode<J> {
...
                 &mut self,
                 _: &PostOrderIterItem<&CommitNode<J>>,
-                inner: Inner<&Arc<ConstructNode<J>>, J, &Arc<ConstructNode<J>>, &NoWitness>,
+                inner: Inner<&Arc<ConstructNode<J>>, J, &Option<Arc<ConstructNode<J>>>, &NoWitness>,
             ) -> Result<ConstructData<J>, Self::Error> {
                 let inner = inner
                     .map(|node| node.arrow())
-                    .map_disconnect(|node| node.arrow());
+                    .map_disconnect(|maybe_node| maybe_node.as_ref().map(|node| node.arrow()));
+                let inner = inner.disconnect_as_ref(); // lol sigh rust
```
uncomputable, line 241: 741909e80a0bc153b7e8215f586e529edd425d61: If the disconnected type `&X` creates so many problems, we could use `X` instead. This would be similar to the witness type `W`.

uncomputable (commented):
5e88ca07ed091adb27f4ac866b3e9b88f4393a9c Looks good. There are some minor comments.

src/node/commit.rs
apoelstra: This would cause a different set of problems :).

apoelstra:
Addressed the two comments.

uncomputable (approved):
ACK 9313898b987191ec1d6b815c6541733de556314b

PR #152, opened 2023-07-07
Human-readable encoding 1: serialization and README

apoelstra:
This PR introduces the human-readable serialization of Simplicity programs. Right now it will serialize disconnect nodes as having only one child (and the next PR will parse such programs). Later, when we introduce typed holes, we will correct this so that disconnect nodes always have two children (the rightmost one being a named hole). But for now disconnect is more-or-less unusable. To avoid too much complexity at once I think this is a reasonable order of operations.

It also slightly changes the syntax from the last PR to allow `'` in symbol names. This is so I can do stuff like `x' := pair x unit` or whatever, where I use the prime symbol to indicate a slightly tweaked version of a different expression.

### DIFF
--- a/simpcli/src/main.rs
+++ b/simpcli/src/main.rs
@@ -1,0 +1,112 @@
+// Simplicity "Human-Readable" Language
+// Written in 2023 by
+//   Andrew Poelstra <simplicity@wpsoftware.net>
+//
+// To the extent possible under law, the author(s) have dedicated all
+// copyright and related and neighboring rights to this software to
+// the public domain worldwide. This software is distributed without
+// any warranty.
+//
+// You should have received a copy of the CC0 Public Domain Dedication
+// along with this software.
+// If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
+//
+
+use simplicity::human_encoding::Forest;
+use simplicity::node::CommitNode;
+use simplicity::{self, BitIter};
+
+use base64::engine::general_purpose::STANDARD;
+use std::env;
+use std::str::FromStr;
+
+/// What set of jets to use in the program.
+// FIXME this should probably be configurable.
+type DefaultJet = simplicity::jet::Elements;
+
+fn usage(process_name: &str) -> Result<(), String> {
+    eprintln!("Usage:");
+    eprintln!("  {} disassemble <base64>", process_name);
+    eprintln!();
+    eprintln!("For commands which take an optional expression, the default value is \"main\".");
+    eprintln!();
+    eprintln!("Run `{} help` to display this message.", process_name);
+    Err("invalid usage".into())
+}
+
+enum Command {
+    Disassemble,
+    Help,
+}
+
+impl FromStr for Command {
+    type Err = String;
+    fn from_str(s: &str) -> Result<Self, String> {
+        match s {
+            "disassemble" => Ok(Command::Disassemble),
+            "help" => Ok(Command::Help),
+            x => Err(format!("unknown command {}", x)),
+        }
+    }
+}
+
+impl Command {
+    fn takes_optional_exprname(&self) -> bool {
+        match *self {
+            Command::Disassemble => false,
+            Command::Help => false,
+        }
+    }
+}
+
+fn main() -> Result<(), String> {
+    let mut args = env::args();
+    let process_name = args.next().unwrap();
+    let process_name = match process_name.rfind('/') {
+        Some(idx) => &process_name[idx + 1..],
+        None => &process_name[..],
+    };
+
+    // Parse command-line args into (command, first_arg, expression)
+    let command = match args.next() {
+        Some(cmd) => match Command::from_str(&cmd) {
+            Ok(cmd) => cmd,
+            Err(e) => {
+                eprintln!("Error: {}.", e);
+                eprintln!();
+                return usage(&process_name);
+            }
+        },
+        None => return usage(&process_name),
+    };
+    let first_arg = match args.next() {
+        Some(s) => s,
+        None => return usage(&process_name),
+    };
+    let _expression = if command.takes_optional_exprname() {
+        args.next().unwrap_or("main".to_owned())
+    } else {
+        String::new()
+    };
+    if args.next().is_some() {
+        usage(&process_name)?;
+    }
+
+    // Execute command
+    match command {
+        Command::Disassemble => {
+            let v = base64::Engine::decode(&STANDARD, first_arg.as_bytes())
+                .map_err(|e| format!("failed to parse base64: {}", e))?;
+            let mut iter = BitIter::from(v.into_iter());
+            let commit = CommitNode::decode(&mut iter)
+                .map_err(|e| format!("failed to decode program: {}", e))?;
+            let prog = Forest::<DefaultJet>::from_program(commit);
+            println!("{}", prog.string_serialize());
+        }
+        Command::Help => {
+            let _ = usage(&process_name);
+        }
+    }
+
+    Ok(())
+}
--- a/simpcli/src/main.rs
+++ b/simpcli/src/main.rs
@@ -24,13 +24,17 @@ use std::str::FromStr;
 // FIXME this should probably be configurable.
 type DefaultJet = simplicity::jet::Elements;
 
-fn usage(process_name: &str) -> Result<(), String> {
+fn usage(process_name: &str) {
     eprintln!("Usage:");
     eprintln!("  {} disassemble <base64>", process_name);
     eprintln!();
     eprintln!("For commands which take an optional expression, the default value is \"main\".");
     eprintln!();
     eprintln!("Run `{} help` to display this message.", process_name);
+}
+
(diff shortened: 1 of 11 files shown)


apoelstra:
I have a crappy CLI tool that I've been mostly interacting with this code through. If you'd like I can clean that up a little and add it to this PR, so you can feed it hex-encoded Simplicity programs and get the serialization.

Could also finally switch to base64..

sanket1729:
I started reviewing this PR. Only read the README so far. Will play around with this over the next couple days. 


Any CLI tool would be helpful. it doesn't have to be cleaned up. I can navigate through the CLI code to figure out stuff

apoelstra:
@sanket1729 you may also be interested in the full branch (which I keep reasonably up to date) https://github.com/BlockstreamResearch/rust-simplicity/tree/2023-06--simplang which includes the CLI tool and also parsing logic.

apoelstra:
Updated README and also removed the `name_map` from `Program` because I don't use it yet and it was complicating some later code. Will restore it in a later PR.

src/human_encoding/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Forest<J: Jet> {
+    programs: HashMap<Arc<str>, Program<J>>,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct Program<J: Jet> {
+    root: Arc<NamedCommitNode<J>>,
+}
```
uncomputable, line 44: cf2ea20bbbf0297347b132d7c1b382e88f8997e4: We should think about how we handle programs of other node types. So far, we use `&Node` or `Arc<Node>` in methods like `BitMachine::for_program()`.

src/human_encoding/mod.rs
```diff
@@ -52,4 +54,96 @@ impl<J: Jet> Program<J> {
...
+
+        // Pass 2: actually print everything
+        let mut ret = String::new();
+        if !witness_lines.is_empty() {
+            ret += "-- Witnesses\n";
+            ret += &print_lines(&witness_lines, false);
+            ret += "\n";
+        }
+        if !const_lines.is_empty() {
+            // FIXME detect scribes
```
uncomputable, line 138: 0fe9d92823b1aaddd39ca0ae83374db1884205f0: [I have some code that detects scribe.](https://github.com/uncomputable/hal-simplicity/blob/master/src/compress.rs) I can rebase it and make a PR later if you want.

src/human_encoding/README.md
```diff
@@ -0,0 +1,189 @@
...
+## Syntax
+
+The syntax is defined in `src/human_encoding/parse/ast.rs`. It currently uses the
+`santiago` parser generator, but we would like to move away from this, probably to
+an ad-hoc parser, to avoid poor asymptotic behavior and to get better error messages.
+
+Comments are started by `--` and end at the next newline. This is the only aspect
+in which whitespace is significant.
+
+Simplicity expressions are a series of **definitions** of the form
```
uncomputable, line 25: aed42d0045f0403f05e263c737bf6dc3c4153d9a: The word "expression" has a different meaning here than in the rest of the document. A (semantic) Simplicity expression is defined by a series of (syntactic) definitions.

Maybe "A Simplicity file consists of a series of DEFINITIONS of the form..."

src/human_encoding/README.md
```diff
@@ -0,0 +1,189 @@
...
+Simplicity expression. The type checker will ensure that the target type of `node`
+is 1.
+
+In general, we do not need to obtain a single expression. It is permissible to
+encode a "DAG forest".
+
+The name `main` is special for several reasons:
+* An expression named `name` implicitly has the type ascription `1 -> 1`. That is, it must always be a program.
+* To generate a commitment-time program from an expression, it must be called `main`.
+* Type ascriptions for `main` and its children are enforced **after** type inference has completed, so they act as sanity checks but cannot change the output of type inference. For other expressions, type ascriptions are enforced **before** and may guide inference.
```
uncomputable, line 122: aed42d0045f0403f05e263c737bf6dc3c4153d9a: We should define what a Simplicity program is, and also what a (semantic) Simplicity expression is.

src/human_encoding/README.md
```diff
@@ -0,0 +1,189 @@
...
+* `const` followed by a VALUE (defined below);
+* `fail` followed by an ENTROPY (defined below); or
+* `(` followed by another EXPRESSION followed by `)`.
+
+Note that while we allow parenthesis to help group parts of expressions for human
+understanding, they are never needed for disambiguation and are essentially
+ignored by the parser.
+
+A HOLE is the literal `?` followed by a NAME. It indicates an expression that has
+yet to be defined. Holes have a different namespace than other names.
```
uncomputable, line 63:
```suggestion
yet to be defined. Holes have a different namespace than names / NAMES.
```

src/human_encoding/serialize.rs
```diff
@@ -0,0 +1,37 @@
...
+        // The default value serialization shows the whole structure of
+        // the value; but for words, the structure is always fixed by the
+        // length, so it is fine to just serialize the bits.
+        if let Ok(hex) = self.0.try_to_bytes() {
+            f.write_str("0x")?;
+            f.write_str(&hex.to_hex())?;
+        } else {
+            f.write_str("0b")?;
+            self.0
+                .do_each_bit(|b| f.write_str(if b { "1" } else { "0" }).expect("FIXME"));
```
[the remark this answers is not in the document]
apoelstra: I think we should actually just use the `PreOrderIter` which is made for exactly this and is a little more Rustic. (In Rust it's usually clearer to use external (for-loop-based) rather than internal (closure-based) iteration.) This FIXME predates `Value` being `DagLike`.

src/human_encoding/README.md
apoelstra: Yes, "Simplicity expression" has a different meaning than "expression". I am careful about this but I agree that it's confusing and potentially misleading.

src/human_encoding/README.md
apoelstra: Yes, but probably not in the README for the human-readable encoding.

src/human_encoding/mod.rs
apoelstra: That'd be awesome, though I think we need to decide what to do about const word jets and whether a "scribe" includes them.

apoelstra: I'd also hold off on this until the parser is PR'd.

src/human_encoding/mod.rs
apoelstra: Well, right now we take a `RedeemNode` in the bit machine because that's the only node type that is guaranteed to have witness values, complete types, and disconnected expressions. If we add another such node type we'll need to add a trait describing this.

apoelstra: Nothing in my PR stack involves a `NamedRedeemNode` and I haven't found a use for one ... I think, if you want to execute a program, you should just convert it to `RedeemNode` beforehand.

simpcli/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+type DefaultJet = simplicity::jet::Elements;
+
+fn usage(process_name: &str) -> Result<(), String> {
+    eprintln!("Usage:");
+    eprintln!("  {} disassemble <base64>", process_name);
+    eprintln!();
+    eprintln!("For commands which take an optional expression, the default value is \"main\".");
+    eprintln!();
+    eprintln!("Run `{} help` to display this message.", process_name);
+    Err("invalid usage".into())
```
uncomputable, line 34: 2459f566bfea051bec2225f82bbdf68a69abfcd1: We should not return an error on `simpcli help`.

uncomputable (commented):
2459f566bfea051bec2225f82bbdf68a69abfcd1: Looks good. There are some comments. The biggest question is what "expression" should mean.

src/human_encoding/serialize.rs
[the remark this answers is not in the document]
apoelstra: BTW @uncomputable for future reference, to upload patches to PRs, you can literally just push (or force-push) to the branch on my repo. Because I opened a PR here, Github has made that branch writeable to maintainers of this repo.

apoelstra:
Force pushed to

* fix FIXME in the serialization logic for constant words
* update README in a couple places, most notably to change "definition" to "named expression"

apoelstra:
* Got rid of the confusingly-named `Program` type in favor of `Forest`.
* Added a CI job which just runs `cargo test` on the simpcli crate.

No other changes.

uncomputable (commented):
851928f4f61082f3773011dbabb8b6279568b23a: Looks good. I think we should split "named expressions" into "definitions" and "type bounds" / "declarations". We should also emphasize that there are three namespaces: one for expressions, one for holes and one for types.

apoelstra:
Changed `::` to `:` and updated README again.

We have another TODO to change `#` to `#{...}` but I think we can do that in a followup PR. Up to you.

PR #154, opened 2023-07-11
Ranged descriptors

uncomputable:
Adds rudimentary support for ranged descriptors. ~We could add more convenience methods, but this requires us to copy / implement a lot of traits such as `ForEachKey`.~

### DIFF
--- a/src/policy/descriptor.rs
+++ b/src/policy/descriptor.rs
@@ -9,8 +9,8 @@ use elements::taproot::{
 };
 use miniscript::descriptor::ConversionError;
 use miniscript::{
-    translate_hash_clone, DefiniteDescriptorKey, DescriptorPublicKey, MiniscriptKey, ToPublicKey,
-    Translator,
+    translate_hash_clone, DefiniteDescriptorKey, DescriptorPublicKey, ForEachKey, MiniscriptKey,
+    ToPublicKey, Translator,
 };
 use std::fmt;
 use std::str::FromStr;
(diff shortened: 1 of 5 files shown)


uncomputable:
Loosened the trait bound in `Descriptor<Pk> as FromStr`. Now `Pk` can be pretty much any `MiniscriptKey`.

uncomputable:
Fixed the issue. Now we can decode arbitrary `Policy`s and `Descriptor`s.

uncomputable:
`str::split_once()` is available from 1.52.0 so I need to find a workaround.

uncomputable:
Fixed and rebased on master.

src/policy/descriptor.rs
```diff
@@ -276,29 +277,50 @@ impl Descriptor<DescriptorPublicKey> {
 
 impl<Pk: MiniscriptKey> fmt::Display for Descriptor<Pk> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        fmt::Display::fmt(&self.policy, f)
+        write!(f, "sim({},{})", self.internal_key, self.policy)
```
apoelstra, line 433: In afae99a2e01351a1b8d1eb7b93802e05b0501498:

This commit seems to introduce a new `sim` descriptor. What is the value of having such a descriptor? What is the difference between a `sim` and a normal `tr` descriptor?

src/policy/descriptor.rs
```diff
@@ -276,29 +277,50 @@ impl Descriptor<DescriptorPublicKey> {
...
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        let policy = Policy::from_str(s)?;
-        Ok(Self::single_leaf(policy))
+        let (internal_key, policy) = s
+            .strip_prefix("sim(")
+            .and_then(|s| s.strip_suffix(')'))
+            .and_then(|s| {
+                let mut split = s.split(',');
+                let x = split.next()?;
+                let y = split.next()?;
```
apoelstra, line 453: In afae99a2e01351a1b8d1eb7b93802e05b0501498:

This code will do the wrong thing if there is more than one `,`. You should use `splitn` rather than `split`.

src/policy/descriptor.rs
uncomputable: `sim` takes a Simplicity policy as second argument while `tr` takes Miniscript. We can overload `tr` if you want, but that might be confusing.

src/policy/descriptor.rs
uncomputable: Fixed.

uncomputable:
Same CI failure as in #158 . Cherry-picked the fix.

src/policy/descriptor.rs
apoelstra: I'm confused. The difference between Simplicity and Script happens at the tapleaf level. It wouldn't be "overloading" `tr` to use it for different kinds of taproot descriptors.

uncomputable:
Superseded by #162. We will use ideas from this PR when we implement Simplicity descriptors over in elements-miniscript. But there is no reason to keep this PR open.

PR #158, opened 2023-07-13
Human-readable encoding 2: cleanups and syntax changes

### LINKED ISSUE #157: Support option types
I use 'S A' in the tech-report for the option type.  The idea is that 'S' is shorthand for 'successor', which in turn means 'plus one'.  However feel free to use whatever syntax you are most comfortable with.  The above is a little bit obtuse.

apoelstra:
After/during review of #152 we decided to change the syntax in a couple ways:

* We introduced the `?` syntax in types where `A?` is shorthand for `1 + A`
* We changed the CMR syntax from `#expr` to `#{expr}`
* We added a new `#<64 hex chars>` format for directly specifying CMRs without giving an expression

This PR also includes some cleanups and new types and utility methods that will be used by the parser. Let me know if you want me to move any to the next PR, or if you'd like me to pull the parser into this one.

Fixes #157

### DIFF
--- a/src/human_encoding/README.md
+++ b/src/human_encoding/README.md
@@ -52,8 +52,8 @@ and EXPRESSION is
 * `unit`, `iden`, or `witness`;
 * `injl`, `injr`, `take`, or `drop` followed by another EXPRESSION;
 * `case`, `comp`, or `pair` followed by two EXPRESSIONs;
-* `assertl` followed by an EXPRESSION, a literal `#`, and another EXPRESSION;
-* `assertr` followed by a literal `#` and two EXPRESSIONs;
+* `assertl` followed by an EXPRESSION and a CMR (defined below);
+* `assertr` followed by CMR and an EXPRESSION;
 * a jet, which begins with `jet_` and must belong to the list of jets (FIXME define this list);
 * `const` followed by a VALUE (defined below);
 * `fail` followed by an ENTROPY (defined below); or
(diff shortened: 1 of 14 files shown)


apoelstra:
Sigh, CI failure is a rustc bug https://github.com/rust-lang/rust/issues/110475

I will need to patch it in elements-miniscript.

src/human_encoding/README.md
```diff
@@ -151,7 +159,7 @@ Expressions may be
 * one of the core combinators `unit`, `iden`, `comp`, `injl`, `injr`, `case`, `take`, `drop`, `pair`, followed by subexpression(s) as needed;
 * the `disconnect` combinator followed by an expression and a hole;
 * the `witness` combinator which currently allows no subexpressions;
-* the assertions, `assertl` or `assertr`, which take two subexpressions, one of which will be hidden in the decoded program. The hidden subexpression should be prefixed by `#` which indicates to the parser to take the CMR of that expression, not the expression itself.
+* the assertions, `assertl` or `assertr`, which take a subexpressions and a CMR. The CMR is encoded as a full expression prefixed by `#{` and suffixed by `}`; but in the bit-encoding the expression does not appear, only its CMR;
```
uncomputable, line 162: 64c6ea43a69054a1af61ccaf325f18586ec69cf1: Typo, and the definition of CMR was extended. I would leave out the latter part because it is already explained in an earlier section. 
```suggestion
the assertions, `assertl` or `assertr`, which take a subexpression and a CMR;
```

uncomputable (approved):
ACK b608ec19cc326b3bca55af8c3dbf01a3a886bf11 Feel free to address the nit.

The named construct node and the error sets look interesting. It is not being used at the moment, but I'm sure the next PR is already waiting.

PR #159, opened 2023-07-18
Move policy

uncomputable:
We move the policy module over to elements-miniscript where we can integrate the Simplicity descriptors etc. into the existing infrastructure. This PR only removes the policy module and the dependency on elements-miniscript.

I will leave this as a draft until we have a working PR on the elements-miniscript side.


uncomputable:
Superseded by #162

PR #160, opened 2023-07-18
Latest miniscript

uncomputable:
Updates simplicity to the latest bitcoin(-hashes), elements and elements-miniscript.

I had to copy hex.rs because bitcoin-hashes dropped support. We might want to use a stable workaround instead once it exists.

There are also problems in the fuzz and bench crate that need to be resolved.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,11 +16,11 @@ name = "simplicity"
 path = "src/lib.rs"
 
 [dependencies]
-bitcoin = { version = "0.29.2", optional = true }
-bitcoin_hashes = "0.11"
+bitcoin = { version = "0.30.0", optional = true }
+bitcoin_hashes = "0.12"
 byteorder = "1.3"
-elements = { version = "0.21.1", optional = true }
-elements-miniscript = { git = "https://github.com/apoelstra/elements-miniscript", tag = "2023-07--rust-simplicity-patch" }
+elements = { version = "0.22.0", optional = true }
+elements-miniscript = "0.2.0"
 simplicity-sys = { version = "0.1.0", path = "./simplicity-sys" }
 actual-serde = { package = "serde", version = "1.0.103", features = ["derive"], optional = true }
 
--- a/src/jet/elements/environment.rs
+++ b/src/jet/elements/environment.rs
@@ -131,11 +131,11 @@ impl ElementsEnv {
 impl ElementsEnv {
     /// Return a dummy Elements environment
     pub fn dummy() -> Self {
-        Self::dummy_with(elements::PackedLockTime::ZERO, elements::Sequence::MAX)
+        Self::dummy_with(elements::LockTime::ZERO, elements::Sequence::MAX)
     }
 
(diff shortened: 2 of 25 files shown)


uncomputable:
Rebased on master. Fuzzing is still broken because we need a patched version of bitcoin-hashes. There is also a weird bug for 1.48.0.

apoelstra:
As suggested on IRC please try rust-bitcoin/rust-bitcoin commit c1c76756a65d1eb046ac934425603c83dbcd0d94 for bitcoin_hashes in fuzz/Cargo.toml.

uncomputable:
Heads up, `quote` and `proc-macro2` no longer compile for 1.48.0. They need to be patched: 7a1b268

Cargo.toml
```diff
@@ -16,11 +16,11 @@ name = "simplicity"
...
 [dependencies]
-bitcoin = { version = "0.29.2", optional = true }
-bitcoin_hashes = "0.11"
+bitcoin = { version = "0.30.0", optional = true }
+bitcoin_hashes = "0.12"
 byteorder = "1.3"
-elements = { version = "0.21.1", optional = true }
-elements-miniscript = { git = "https://github.com/apoelstra/elements-miniscript", tag = "2023-07--rust-simplicity-patch" }
+elements = { version = "0.22.0", optional = true }
+elements-miniscript = "0.2.0"
```
apoelstra, line 23: In 66c07d2ce8474e20b9bc66a566f654c6b8c5f3b4:

Heads up the commit message still says "patched version of elements-miniscript". But we are actually using the released version.

src/jet/elements/environment.rs
```diff
@@ -70,7 +70,7 @@ pub struct ElementsEnv {
     /// Optional Annex.
     annex: Option<Vec<u8>>,
     /// Genesis block hash
-    genesis_hash: elements::BlockHash,
+    genesis_hash: BlockHash,
```
apoelstra, line 73: In 66c07d2ce8474e20b9bc66a566f654c6b8c5f3b4:

Please use the `elements::` or `bitcoin::` prefix for blocks, transactions, and any related types.

uncomputable: I think I had conflicting versions of elements at some point. This line even didn't need to change.

apoelstra:
> As suggested on IRC please try rust-bitcoin/rust-bitcoin commit c1c76756a65d1eb046ac934425603c83dbcd0d94 for bitcoin_hashes in fuzz/Cargo.toml.

For future reference: on IRC we noticed that this commit actually doesn't work, because it includes changes to the `hash_newtype!` macro that cause compilation failures when used in rust-bitcoin. Instead we had to create a one-off commit based on 0.30.1, which I tagged on apoelstra/rust-bitcoin as `2023-07--0.30.1-with-fuzzcfg`.

apoelstra (approved):
ACK 91ecf06b7a3ccdbfd745f5183962036104c08697

PR #162, opened 2023-07-20
Invert miniscript dependency

uncomputable:
We remove elements-miniscript from rust-simplicity so that the former can depend on the latter. It will be much easier to implement Simplicity descriptors in elements-miniscript where there exists most of the required infrastructure. Doing this in rust-simplicity would lead to a lot of copy and pasting.

We decided to create a new abstraction layer for Simplicity policy (keys, satisfiers, ...) to keep all the messy details about constructing Simplicity programs in rust-simplicity. elements-miniscript can simply implement these traits for the existing analogues (MiniscriptKey, Satisfier, ...) to make use of them.

One drawback of this PR is that Policy no longer implements FromStr. We would have to copy a lot of code from elements-miniscript to adapt the approach from master, and we don't use FromStr outside of tests. elements-miniscript can use its Tree trait to parse Policy (see commit message).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,7 +9,7 @@ edition = "2018"
 [features]
 default = [ "bitcoin", "elements" ]
 test-utils = ["simplicity-sys/test-utils"]
-serde = ["actual-serde", "bitcoin/serde", "elements-miniscript/serde"]
+serde = ["actual-serde", "bitcoin/serde", "elements/serde"]
 
 [lib]
 name = "simplicity"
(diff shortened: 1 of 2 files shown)


apoelstra:
I think it's ok that we can't use `FromStr` directly on `Policy`. `Policy` shouldn't occur outside of descriptors anyway, and elements-miniscript can do string-parsing there using its own `FromTree` trait (which it can impl on `Policy`), as you observe.

apoelstra (approved):
ACK d7e66a0111f2d501d424362f07a9aed8561a6872

PR #163, opened 2023-07-20
Easy CI

uncomputable:
This is the second part to #148 where we copy the CI infrastructure from rust-bitcoin.

We use a script at contrib/test.sh to run all our checks. The Github yml files simply set up a machine and run the script with certain arguments. This means that there is one job for each architecture + rust version.

Having a bash script should make it easier to run CI locally.

I enabled rustdoc tests because they were already part of the script. It was easy to fix the existing errors and this additional check shouldn't slow us down.


uncomputable:
We need to introduce lockfiles to this repo and add a few lines from https://github.com/rust-bitcoin/rust-bitcoin/blob/master/contrib/test.sh.

contrib/test.sh
```diff
@@ -0,0 +1,130 @@
...
+
+    # We should not have any duplicate dependencies. This catches mistakes made upgrading dependencies
+    # in one crate and not in another (e.g. upgrade bitcoin_hashes in bitcoin but not in secp).
+    duplicate_dependencies=$(
+        # Only show the actual duplicated deps, not their reverse tree, then
+        # whitelist the 'syn' crate which is duplicated but it's not our fault.
+        cargo tree  --target=all --all-features --duplicates \
+            | grep '^[0-9A-Za-z]' \
+            | grep -v 'syn' \
+            | wc -l
```
apoelstra, line 50: In b2aabc679853cb221b7705487b5952b3f055235f:

Should use `grep -vc` rather than `grep -v | wc -l`.

contrib/test.sh
```diff
@@ -0,0 +1,130 @@
...
+fi
+
+# Run benchmarks if told to
+if [ "$DO_BENCH" = true ]
+then
+    echo "********* Benchmarks *************"
+    cargo install cargo-criterion
+    cd jets-bench
+    cargo test
+    cargo criterion --no-run
```
apoelstra, line 109: In b2aabc679853cb221b7705487b5952b3f055235f:

You should add a `cd ..` every time you `cd` into something, so that you have have multiple `DO_*` flags set at once. (Or you can use bash rather than sh, and use `pushd`/`popd`.)

contrib/test.sh
```diff
@@ -0,0 +1,130 @@
...
+    cd simpcli
+    cargo test
+fi
+
+# Use as dependency if told to
+if [ "$AS_DEPENDENCY" = true ]
+then
+    cargo new dep_test 2> /dev/null # Mute warning about workspace, fixed below.
+    cd dep_test
+    echo 'bitcoin = { path = "..", features = ["serde"] }\n\n' >> Cargo.toml
```
apoelstra, line 125: In b2aabc679853cb221b7705487b5952b3f055235f:

If you want to use `\n` you need to use `printf` rather than `echo`. The behavior of `echo` isn't consistent across shells, and I think on mine it'll do the wrong thing.

contrib/test.sh
```diff
@@ -0,0 +1,130 @@
...
+    # Adding an empty workspace section excludes this crate from the rust-bitcoin workspace.
+    echo '[workspace]\n\n' >> Cargo.toml
+
+    cargo test --verbose
```
apoelstra, line 129: In b2aabc679853cb221b7705487b5952b3f055235f:

Should try to clean this up afterward. Probably need to use an `if` statement around `cargo test` so that your cleanup code will run even if it fails.

Alternately we could just commit the `dep_test` crate to the repo.

contrib/test.sh
```diff
@@ -0,0 +1,130 @@
...
+    fi
+fi
+
+echo "********* Testing no features *************"
+# Test without any features
+cargo test --locked --verbose --no-default-features
+
+echo "********* Testing default *************"
+# Then test with the default features
+cargo test --verbose
```
apoelstra, line 65: In b2aabc679853cb221b7705487b5952b3f055235f:

Let's add an `cargo test --all-features` here as well. We should also do `cargo build` before every `cargo test`, since building does a slightly different compilation.

uncomputable:
I will redo this in nix, possibly using code from https://github.com/apoelstra/local-nix-ci.

PR #164, opened 2023-07-26
Human-readable encoding 3: add parser

apoelstra:
This uses the `santiago` parser generator, which requires rustc 1.58. Bump the MSRV to 1.58, though we will revert this later when we replace the generated parser with a more efficient ad-hoc one.

Since #152 added serialization, parsing gives us the ability to round-trip. So update many unit tests and add a fuzz test to test this.

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -34,7 +34,7 @@ decode_program,
           key: cache-${{ matrix.target }}-${{ hashFiles('**/Cargo.toml','**/Cargo.lock') }}
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.58
+          toolchain: 1.64
           override: true
           profile: minimal
       - name: fuzz
--- a/simpcli/src/main.rs
+++ b/simpcli/src/main.rs
@@ -16,9 +16,10 @@ use simplicity::human_encoding::Forest;
 use simplicity::node::CommitNode;
 use simplicity::{self, BitIter};
 
+use base64::display::Base64Display;
 use base64::engine::general_purpose::STANDARD;
-use std::env;
 use std::str::FromStr;
+use std::{env, fs};
 
 /// What set of jets to use in the program.
 // FIXME this should probably be configurable.
@@ -26,6 +27,7 @@ type DefaultJet = simplicity::jet::Elements;
 
 fn usage(process_name: &str) {
     eprintln!("Usage:");
+    eprintln!("  {} assemble <filename>", process_name);
     eprintln!("  {} disassemble <base64>", process_name);
     eprintln!();
     eprintln!("For commands which take an optional expression, the default value is \"main\".");
@@ -39,6 +41,7 @@ fn invalid_usage(process_name: &str) -> Result<(), String> {
 }
 
 enum Command {
+    Assemble,
     Disassemble,
     Help,
 }
@@ -47,6 +50,7 @@ impl FromStr for Command {
     type Err = String;
     fn from_str(s: &str) -> Result<Self, String> {
         match s {
+            "assemble" => Ok(Command::Assemble),
             "disassemble" => Ok(Command::Disassemble),
             "help" => Ok(Command::Help),
             x => Err(format!("unknown command {}", x)),
@@ -57,8 +61,23 @@ impl FromStr for Command {
 impl Command {
     fn takes_optional_exprname(&self) -> bool {
         match *self {
+            Command::Assemble => false,
             Command::Disassemble => false,
             Command::Help => false,
+        }
+    }
+}
+
+fn parse_file(name: &str) -> Result<Forest<DefaultJet>, String> {
+    let s = fs::read_to_string(name).map_err(|e| format!("failed to read file {}: {}", name, e))?;
+    match Forest::parse(&s) {
+        Ok(prog) => Ok(prog),
+        Err(mut errs) => {
+            errs.add_context(std::sync::Arc::from(s));
+            eprintln!("Errors:");
+            eprintln!("{}", errs);
+            eprintln!();
+            Err(format!("failed to parse file {}", name))
         }
     }
 }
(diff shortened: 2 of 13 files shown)


uncomputable:
We could have different MSRV for different sub crates. We are not going to replace santiago with our own implementation any time soon.

Could you also explain why you bumped the Rust version of the fuzz tests?

apoelstra:
@uncomputable okay, but the parser is part of the main crate, so I don't see how changing MSRV for different crates would help anything.

And I bumped the Rust version for the fuzztests because the `regex` dependency wasn't compiling and I didn't feel like figuring out the right way to add version pins to the fuzzer script.

src/human_encoding/parse/ast.rs
```diff
@@ -0,0 +1,846 @@
...
+            } else {
+                unreachable!("expected string, got something else")
+            }
+        };
+        "expr" => rules "(" "expr" ")" => |toks| toks[1].clone();
+        "expr" => rules "nullary" => Ast::from_combinator;
+        "expr" => rules "unary" "expr" => Ast::from_combinator;
+        "expr" => rules "binary" "expr" "expr" => Ast::from_combinator;
+
+        // TODO should we allow CMRs as literals for constant words?
```
uncomputable, line 623: 8e711fd6b8a34c454d7ec83a8ad672cc6a8e12f1: What does this todo mean? Do CMRs (`#...` / `#{...}`) have a role outside of assertions?

src/human_encoding/parse/ast.rs
```diff
@@ -0,0 +1,846 @@
...
+    fn from_type_lexeme(lexemes: &[&std::rc::Rc<Lexeme>]) -> Self {
+        assert_eq!(lexemes.len(), 1);
+        let position = &lexemes[0].position;
+        match lexemes[0].raw.as_str() {
+            "1" => Ast::Type(Some(Type::One)),
+            "2" => Ast::Type(Some(Type::Two)),
+            other => {
+                assert_eq!(&other[..2], "2^");
+                match str::parse::<u32>(&other[2..]) {
+                    // TODO how many of these should we support?
```
uncomputable, line 289: 8e711fd6b8a34c454d7ec83a8ad672cc6a8e12f1: Ideally we should support as many powers as in `types::precomputed::N_POWERS`, but the numbers get pretty big. We could add SI prefixes for kibi and mebi. Maybe in another PR.

src/human_encoding/parse/ast.rs
```diff
@@ -0,0 +1,846 @@
...
+impl Type {
+    /// Convert to a Simplicity type
+    pub fn reify(self) -> types::Type {
+        match self {
+            Type::Name(s) => types::Type::free(s),
+            Type::One => types::Type::unit(),
+            Type::Two => types::Type::sum(types::Type::unit(), types::Type::unit()),
+            Type::Product(left, right) => types::Type::product(left.reify(), right.reify()),
+            Type::Sum(left, right) => types::Type::sum(left.reify(), right.reify()),
+            Type::TwoN(n) => types::Type::two_two_n(n as usize), // cast OK as we are only using tiny numbers
```
uncomputable, line 105: 8e711fd6b8a34c454d7ec83a8ad672cc6a8e12f1: Something weird is going on here. `Type::TwoN(n)` means 2^n while `types::Type::two_two_n(n)` means 2^(2^n).

src/human_encoding/parse/mod.rs
```diff
@@ -0,0 +1,629 @@
...
+            match finalized {
+                Ok(root) => {
+                    roots.insert(Arc::clone(name), root);
+                }
+                Err(errs) => errors.merge(&errs),
+            }
+        }
+    }
+
+    // Finally, build name map.
```
uncomputable, line 574: 8e711fd6b8a34c454d7ec83a8ad672cc6a8e12f1: Outdated comment. The map is already built.

simpcli/src/main.rs
```diff
@@ -104,6 +123,27 @@ fn main() -> Result<(), String> {
 
     // Execute command
     match command {
+        Command::Assemble => {
+            let prog = parse_file(&first_arg)?;
+
+            let roots = prog.roots();
+            let mut error = false;
+            for name in roots.keys() {
+                if **name != *"main" {
```
uncomputable, line 132: bac6647aa3601812908c45642d48b884d38f11c3 `name.as_ref() != "main"` seems more canonical. I have an aversion against long sequences of `*` and `&`.

uncomputable (commented):
34075113e7873d4dfd15a2b1161b2ebbcffcabcc: Finished reviewing. I'm mostly confused about the seeming off-by-one error in the powers of two type annotations and why the existing programs work. I will write a unit test. I will also write a fuzz test for the parser. The rest looks good.

src/human_encoding/parse/ast.rs
```diff
@@ -0,0 +1,846 @@
...
+    /// The unit type 1
+    One,
+    /// The bit type 1+1
+    Two,
+    /// A product type (A * B)
+    Product(Box<Type>, Box<Type>),
+    /// A sum type (A + B)
+    Sum(Box<Type>, Box<Type>),
+    /// An exponential type 2^n where n is one of 2, 4, 8, 16, ..., 1024.
+    TwoN(u32),
```
uncomputable, line 93: 8e711fd6b8a34c454d7ec83a8ad672cc6a8e12f1: `TwoN(n)` means 2^n many bits. `Two(9)` is converted to the Simplicity type `2^512` and so on. We could rename it to `Bits` and update the documentation to make this clear. I was really confused why `TwoN(n)` was passed as `two_two_n(n)` and everything worked.

.github/workflows/fuzz.yml
```diff
@@ -34,7 +34,7 @@ decode_program,
           key: cache-${{ matrix.target }}-${{ hashFiles('**/Cargo.toml','**/Cargo.lock') }}
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.58
+          toolchain: 1.64
```
uncomputable, line 37: 34075113e7873d4dfd15a2b1161b2ebbcffcabcc: fuzz/generate-files.sh needs to be updated.

src/human_encoding/parse/ast.rs
apoelstra: @uncomputable yes, they are provided as input to the first child of the `disconnect` combinator.

src/human_encoding/parse/ast.rs
apoelstra: `TwoN(n)` is two^(two^n). I should probably rename the variant.

src/human_encoding/parse/ast.rs
apoelstra: Will leave for a later PR.

uncomputable (approved):
ACK cb597021439e1611764ae70ed59030446f86c635

sanket1729:
post merge ACK. I have playing with this locally and so far did not encounter any unexpected issues.

PR #166, opened 2023-08-01
Fuzz human encoding parser

uncomputable:
This test checks if there are byte strings that crash the parser. Roundtrip checks turn out to be complicated to implement and potentially slow, unfortunately.


uncomputable:
Sure. Let's make a PR for the typed holes.

PR #167, opened 2023-08-09
Update simplicity to e5b477f3c5d0bd4af5fa2b364d29cf8a00907074

sanket1729:
Also fix broken test cases because of CMR mismatch.

The last commit splits out heavy jets so we can run them longer in our benchmarks.


sanket1729:
I suggest ignoring the clippy error as that is something auto-generated by Haskell script which can be fixed at a later point

sanket1729:
The third commit is clunky because we need to fix all the upgrade bugs in the same update commit.

apoelstra:
Wow! Thanks for doing all this.

apoelstra (approved):
ACK 8e5c533823e35c1aa1708109d4c6ff12fa858fc2

apoelstra:
We should fix the clippy thing or patch CI, but we can do this in a later PR since this one is pretty big.

PR #169, opened 2023-08-14
Introduce typed holes; improve disconnect support in human-readable encoding

apoelstra:
This introduces the notion of "typed holes" which can be used as placeholders in the text encoding of Simplicity. It uses these holes for disconnect, so that at commitment time users can set their right children to placeholders. The old behavior, which allowed putting a complete expression into `disconnect` that wouldn't be committed to, was dangerous.

### DIFF
--- a/simpcli/example_programs/fixpoint.simpl
+++ b/simpcli/example_programs/fixpoint.simpl
@@ -1,0 +1,32 @@
+
+-- Witnesses
+-- IMR: [none]
+wit_input := witness
+
+-- All jets have source type 1; but to use the `pair` combinator we want one
+-- with source type 2^256. To get this, we compose it with unit.
+sha256_init : 2^256 -> _
+sha256_init := comp unit jet_sha_256_ctx_8_init
+
+-- Using this, we can write a self-contained "take 32 bytes and compute their
+-- sha2 hash" function.
+-- IMR: 8e341445...
+sha256 : 2^256 -> 2^256
+sha256 := comp
+    comp
+        pair sha256_init iden
+        jet_sha_256_ctx_8_add_32
+    jet_sha_256_ctx_8_finalize
+
+-- Check eq
+assert_fixpoint : 2^256 -> 1
+assert_fixpoint := comp
+    comp
+        pair (comp iden sha256) iden
+        jet_eq_256
+    jet_verify
+
+-- IMR: [none]
+main := comp wit_input assert_fixpoint
+
+
--- a/simpcli/example_programs/twowit.simpl
+++ b/simpcli/example_programs/twowit.simpl
@@ -1,0 +1,14 @@
+--
+-- Attempts to reuse a witness name.
+--
+-- By uncommenting wit2 and replacing one of the wit1 references with wit2,
+-- it'll work...but is it what you intended?
+--
+
+wit1 := witness : 1 -> 2^32
+--wit2 := witness : 1 -> 2^32
+wit_tuple := pair wit1 wit1 : 1 -> 2^64
+
+check_eq := comp wit_tuple jet_eq_32
+main := comp check_eq jet_verify
+
--- a/src/human_encoding/mod.rs
+++ b/src/human_encoding/mod.rs
@@ -141,6 +141,7 @@ impl<J: Jet> Forest<J> {
         let mut witness_lines = vec![];
         let mut const_lines = vec![];
         let mut program_lines = vec![];
+        let mut disc_idx = 0;
         // Pass 1: compute string data for every node
         for root in self.roots.values() {
             for data in root.as_ref().post_order_iter::<MaxSharing<_>>() {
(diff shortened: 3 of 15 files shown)


simpcli/example_programs/fixpoint.simpl
```diff
@@ -0,0 +1,32 @@
+
+-- Witnesses
+-- IMR: [none]
+wit_input := witness
+
+-- All jets have source type 1; but to use the `pair` combinator we want one
+-- with source type 2^256. To get this, we compose it with unit.
+sha256_init :: 2^256 -> _
```
uncomputable, line 8: 04f8ffad7e92ce024363816b3b13c7d457d29c15: should be single `:`

src/human_encoding/mod.rs
```diff
@@ -175,11 +175,13 @@ impl<J: Jet> Forest<J> {
...
+                let cmr_str = /*if let node::Inner::Witness(..) = node.inner() {
                     String::new()
-                } else {
-                    format!("-- cmr {:.8}...", node.cmr())
-                };
+                } else { */
+                    format!("-- cmr {}...", node.cmr())
+                    + &format!("  -- imr {}...", node.imr().as_ref().map(ToString::to_string).unwrap_or("???".to_string()))
+//                };
+;
```
uncomputable, line 184: 04f8ffad7e92ce024363816b3b13c7d457d29c15: We probably want to remove the special case for witness entirely.

simpcli/example_programs/hidden.simpl
```diff
@@ -0,0 +1,2 @@
+-- Nothing but a hidden node. Should not parse.
+main := hidden 0000000000000000000000000000000000000000000000000000000000000000
```
uncomputable, line 2: b0f12ba6c86439d797a76de9832fe72d73aab268: Does not parse

simpcli/example_programs/twowit.simpl
```diff
@@ -0,0 +1,14 @@
+--
+-- Attempts to reuse a witness name.
+--
+-- By uncommenting wit2 and replacing one of the wit1 references with wit2,
+-- it'll work...but is it what you intended?
+--
+
+wit1 := witness :: 1 -> 2^32
```
uncomputable, line 8: b0f12ba6c86439d797a76de9832fe72d73aab268: Should be `:`

simpcli/example_programs/disconnect.simpl
```diff
@@ -0,0 +1,22 @@
+-- Fuzz crash:
+-- dc800000000000000140d48704000000ffffffffff87c700000001000007ffffffffffffffffffffffffffff000000d934ffffddd9fff8ffffffd907ff2b0dff08280000000fffffffffffffff0202000000000000000600
+
+-- Witnesses
+wit1 := witness : (((2^256 * B) * (2^256 * B)) + D) -> F
```
uncomputable, line 5: c6eb2128b21ece2da95f8c9c63a12070a951bd9c
```
    5 | wit1 := witness : (((2^256 * B) * (2^256 * B)) + D) -> F 
      | ^
      \-- failed to apply bound `1` to existing bound `2^256 × B`: binding source type annotation
```

src/human_encoding/mod.rs
apoelstra: Oops, this code just shouldn't be there. *Or* we should step back and consider whether we want to add CMRs, IMRs, or both (and how). Right now it's an ad-hoc choice based on what I currently needed to generate blog posts.

apoelstra:
Rebased and renamed last commit. No other changes. I will fix the example programs tomorrow.

apoelstra:
FIxed the example programs. Deleted the hidden and disconnect ones, since those test bugs that are no longer architecturally possible.

src/human_encoding/mod.rs
uncomputable: Let's fix this.

uncomputable (commented):
7494aab3b5e643d33165d931b5e6604b6faaf2a0: The example programs work now. Let's resolve the printing of CMRs and IMRs, and then this PR is ready to be merged.

src/human_encoding/mod.rs
apoelstra: I think `malleable` is misleading. How about `[undetermined]`? Or even `[undetermined at commitment time]` since we've got a lot of room for characters here.

uncomputable (commented):
5465612f211aed546ebad22b63cc76a0c9b60afb: The formatting looks good.

I could reproduce @sanket1729's error on my end. The disassembly output has a `main`, but the assembler doesn't find it.

uncomputable (approved):
5465612f211aed546ebad22b63cc76a0c9b60afb ACK. Sanket's bug is already present on master.

uncomputable:
I opened #173

PR #170, opened 2023-08-16
Policy fixes

uncomputable:
Adds fixes to make Policy work with Miniscript, descriptors and the Simplicity wallet.

The only controversial change should be the implementation of Simplicity key traits for every Miniscript key. This introduces a diamond pattern: rust-miniscript → rust-simplicity → elements-miniscript and rust-miniscript → elements-miniscript. The situation is not ideal, but it avoids violating the orphan rule that caused us so much trouble.

In particular we need to convert `elements_miniscript::Satisfier` (with Miniscript keys) into `simplicity::Satisfier` (with Simplicity keys). We could make `elements_miniscript::Satisfier` a subtrait of simplicity::Satisfier, but we would have to do that for every generic implementation of `elements_miniscript::Satisfier`. We would need to copy code to rust-simplicity and implement `simplicity::Satisfier` for nonsensical types such as `HashMap<Pk, ElementsSig>`.

The diamond pattern seems like a good solution that will work until we have a better architecture. It allows us to use Miniscript and Simplicity keys interchangeably, while all the messy Simplicity details stay inside rust-simplicity.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ path = "src/lib.rs"
 
 [dependencies]
 bitcoin = { version = "0.30.0", optional = true }
+bitcoin-miniscript = { package = "miniscript", version = "10.0" }
 byteorder = "1.3"
 elements = { version = "0.22.0", optional = true }
 hashes = { package = "bitcoin_hashes", version = "0.12" }
--- a/src/policy/ast.rs
+++ b/src/policy/ast.rs
@@ -215,6 +215,19 @@ impl<Pk: SimplicityKey> Policy<Pk> {
             _ => {}
         }
     }
+
+    /// Return an iterator over the fragments of the policy.
+    pub fn iter(&self) -> PolicyIter<'_, Pk> {
+        PolicyIter::new(self)
+    }
+
+    /// Return an iterator over the public keys of the policy.
+    pub fn iter_pk(&self) -> impl Iterator<Item = Pk> + '_ {
+        self.iter().filter_map(|fragment| match fragment {
+            Policy::Key(key) => Some(key.clone()),
+            _ => None,
+        })
+    }
 }
 
 impl<Pk: SimplicityKey> fmt::Debug for Policy<Pk> {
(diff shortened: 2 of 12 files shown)


src/policy/ast.rs
```diff
@@ -37,7 +37,7 @@ use super::serialize;
 /// given a witness.
 ///
 /// Furthermore, the policy can be normalized and is amenable to semantic analysis.
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord)]
+#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, std::hash::Hash)]
```
apoelstra, line 40: In f37a401dbd982830233d1c2f8a36b8cd5748c6a1:

You should just say `Hash` here. I had no idea this fully-qualified syntax was legal, and it's not necessary/idiomatic.

src/policy/sighash.rs
```diff
@@ -0,0 +1,95 @@
...
+            elements::sighash::Prevouts::All(prevouts) => *prevouts,
+            _ => return Err(elements::sighash::Error::PrevoutKind),
+        };
+        let utxos: Vec<_> = all
+            .iter()
+            .map(|o| ElementsUtxo::from(O::borrow(o).clone()))
+            .collect();
+
+        let simplicity_env = crate::jet::elements::ElementsEnv::new(
+            Arc::new((*self.tx).clone()),
```
apoelstra, line 83: In daacacea3044825c6898afc9c9f6dfde05265735:

We should change `ElementsEnv` to be generic over `T` as well so we don't need to do this allocation.

src/policy/ast.rs
```diff
@@ -215,6 +215,19 @@ impl<Pk: SimplicityKey> Policy<Pk> {
             _ => {}
         }
     }
+
+    /// Return an iterator over the fragments of the policy.
```
apoelstra, line 219: In a036b0bc3a7930202e5ae0ed0a459f0c0bf36aeb:

You mention maybe using the dag module in the commit message -- actually we should use the `miniscript::iter` module which will be available in rust-miniscript 11.0. https://github.com/rust-bitcoin/rust-miniscript/pull/567

But for now it's fine  to have a manual implementation.

apoelstra:
Will ACK, but I'd kinda like the `std::hash::Hash` thing to be changed before merging.

Cargo.toml
```diff
@@ -17,6 +17,7 @@ path = "src/lib.rs"
 
 [dependencies]
 bitcoin = { version = "0.30.0", optional = true }
+bitcoin-miniscript = { package = "miniscript", version = "10.0" }
```
apoelstra, line 20: In eb055fa8a5bce506e7752e27426543c54faf8597:

I am tempted to change this from `10.0` to a range from 8.0.1 to 10. Later we will just bump the `10`. Using a range will minimize the amount of grief that we'll have with coordinating new crate versions. (Though arguably we have no need to support miniscript 8 and 9 since those are in the past.)

apoelstra:
Actually I do have one request before ACKing -- @uncomputable can you rearrange the first commits so that the miniscript one comes first? That way I can test the entire PR with the same lockfiles, and not have to worry about Cargo.toml changing partway through.

src/policy/ast.rs
uncomputable: I see. Miniscript's `iter` module is the cousin of Simplicity's `dag` module. I look forward to using it.

src/policy/ast.rs
uncomputable: My goal was to differentiate `std::hash::Hash` from `hashes::Hash`, but just `Hash` is probably more canonical.

uncomputable:
Reordered the commits and fixed the two nits. We will need to update the Haskell generators to include a new use statement.

apoelstra:
Thanks -- but now the benchmarks are not compiling.

uncomputable:
Fixed the benchmarks. Sorry for that. I didn't test them in my local CI.

apoelstra (approved):
ACK d64e3b220fa3cd04e3dea1ac570ca6775b04fa25

apoelstra:
Heads up that I disabled my "checked-in files match those from Simplicity" check for this ACK.

PR #171, opened 2023-08-21
Policy asm fragment

uncomputable:
This commit adds the assembly (asm) fragment to Policy, which allows us to include arbitrary Simplicity programs. The Policy includes just the CMR. It is the Satisfier's responsibility to provide a matching satisfied program at spending time. 

One of the challenges is that Policy with asm fragments does not have a serialization at commitment time. This why we introduce separate getters for the CMR and the program commitment.

### DIFF
--- a/src/merkle/cmr.rs
+++ b/src/merkle/cmr.rs
@@ -12,8 +12,14 @@
 // If not, see <http://creativecommons.org/publicdomain/zero/1.0/>.
 //
 
+use std::sync::Arc;
+
 use crate::impl_midstate_wrapper;
 use crate::jet::Jet;
+use crate::node::{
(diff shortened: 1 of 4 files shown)


apoelstra:
Nice! As you noticed, the `Constructible` trait has different sigantures than the existing CMR constructors. (More importantly than the reference stuff is that the trait methods return `Result` in a bunch of places where the CMR constructors do not.)

apoelstra:
concept ACK. Will really ACK shortly. I'm not certain this is the right approach but it makes sense and will let us move forward, and aligns with my original idea for how `Constructible` could be used.

apoelstra (approved):
ACK b133ffced054d5f06e875282001ab1bdbc43119d

PR #172, opened 2023-08-21
Arc<Value>

uncomputable:
This PR makes the library use `Arc<Value>` everywhere. We already used `Arc<Value>` in some places, but we had to convert it using `.map(Arc::new)` or similar. We also had to write verbose `Box::new(...)` expressions. This PR fixes these issues.

The diff is large but there should be only mechanical changes.


apoelstra:
Would you like me to review/merge this before or after #171?

apoelstra (approved):
ACK 9cdd6f7330f8be4647f594e56476da2762a6d6ff

uncomputable:
Thanks for merging #171. I rebased this PR. It should be good to go now.

apoelstra (approved):
ACK efdea5f5d53440790f25279c984d492eba7bae5a

sanket1729:
I can understand why we did this, and it does simplify some of our code here. But I do think it is the responsibility of the caller to put things to do `Box`, or `Arc` if they need to. As a downstream user, now I am forced to use `Arc<Value>`. 

For example, we use `Value::u32()` to construct `Value::u64()`, so it is convenient for rust-simplicity to change Value::u32() to return Arc<Self>, but the downstream user really just wants to use `Value::u32` as `Value` without the Arc overhead.

sanket1729:
I am not suggesting that we change this, we already have lots of Arcs lying around and there are no downstream users right now. But still something we should consider.

uncomputable:
I feel this is a general API question that includes `Node` and other types. We can open an issue if you like.

apoelstra:
We can open an issue, but I'm strongly in favor of just making users use `Arc` everywhere. This is almost always the correct decision, and since e.g. `Value::u32()` uses a ton of `Arc`s internally so it's not much of a savings to let the user reduce the `Arc` count by one.

We could hide the `Arc` inside the `Value` type the way that we do with `Type`.

PR #175, opened 2023-09-06
Finalize human encoding

uncomputable:
This PR adds methods to convert a `NamedCommitNode` (parsed from text) into a `WitnessNode` with populated witness nodes and populated disconnected branches. This can be finalized into a `RedeemNode` for spending.

A hash map from names to values provides the witness data. This can be obtained from a JSON file.

The disconnected branches come from the forest of Simplicity expressions that are already parsed from the text file. The witness hash map covers the main program and (recursively) all its disconnected expressions. All expressions share a name space and must use unique node names.

### DIFF
--- a/src/human_encoding/named_node.rs
+++ b/src/human_encoding/named_node.rs
@@ -15,15 +15,19 @@
 //! Human-readable Nodes
 
 use crate::dag::{InternalSharing, MaxSharing, PostOrderIterItem};
-use crate::encode;
 use crate::human_encoding::{Error, ErrorSet, Position, WitnessOrHole};
 use crate::jet::Jet;
-use crate::node::{self, Commit, CommitData, CommitNode, Converter, NoDisconnect, NoWitness, Node};
+use crate::node::{
+    self, Commit, CommitData, CommitNode, Converter, Inner, NoDisconnect, NoWitness, Node, Witness,
+    WitnessData,
+};
 use crate::node::{Construct, ConstructData, Constructible};
 use crate::types;
 use crate::types::arrow::{Arrow, FinalArrow};
+use crate::{encode, Value, WitnessNode};
 use crate::{BitWriter, Cmr, Imr};
 
+use std::collections::HashMap;
 use std::io;
 use std::marker::PhantomData;
 use std::sync::Arc;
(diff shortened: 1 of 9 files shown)


src/value.rs
```diff
@@ -193,6 +194,29 @@ impl Value {
         )
     }
 
+    /// Encode a byte slice as a value.
+    ///
+    /// The length of the slice must be a power of two.
+    pub fn from_slice(v: &[u8]) -> Arc<Self> {
```
sanket1729, line 200: I feel this can be mis-used. We cannot really parse a value from byte slice without knowing the exact type of value that we are parsing. For example, length of encoding for types `u16 * u1` is same as `u8 * u8` as it is important for caller to dis-ambiguate that. 

Do you want some form of convenience function that only operates on powers of two, and outputs (2^8^(2^n)) type? In this case, I would create a private function and re-name like something `from_pow_two_8_type`

sanket1729: Also, I think the implementation can be something like `BitIter::read_value(powers_of_two(log(v.len() + 3)))`.

src/human_encoding/named_node.rs
```diff
@@ -124,23 +124,26 @@ impl<J: Jet> NamedCommitNode<J> {
...
+            type Error = crate::Error;
 
             fn convert_witness(
                 &mut self,
                 data: &PostOrderIterItem<&Node<Named<Commit<J>>>>,
                 _: &NoWitness,
             ) -> Result<Option<Arc<Value>>, Self::Error> {
                 let name = &data.node.cached_data().name;
-                let value = self.0.get(name).ok_or(())?;
+                let value = self
```
sanket1729, line 143: I guess for conditionals, even if we are not using the branch, we expect the user to specify some value of the correct type?

sanket1729 (commented):
mostly ACK. Left a few comments/questions. Once resolved, we can merge

src/value.rs
uncomputable: You're right; `from_slice` is ambiguous. I renamed the method to `power_of_two`, similar to `Value::u{1,2,4,8,16,32,64,128}`.

I don't want to use `BitIter` because `power_of_two` is supposed to replace `u{256,512}_from_slice`.

src/human_encoding/named_node.rs
uncomputable: Good catch. I changed the method to keep empty the witness nodes that don't have a mapping. There will be an error during finalization if an unpruned node is empty. Similar for disconnected branches / holes.

uncomputable:
I reordered and squashed the commits. Sorry for the large diff compared to the previous state. This PR was in bad shape.

sanket1729 (approved):
ACK 386f3031e03f3fee4c36905bf3a29334f63f2310

PR #179, opened 2023-11-06
Cost: Compute padding

uncomputable:
Extend the `Cost` struct to compute the necessary padding based on a given witness stack. Test the new methods.

Besides of `get_padding()`, the public API stays unchanged.


uncomputable:
Rebased, squashed and ready for review

src/analysis.rs
```diff
@@ -100,13 +100,11 @@ impl Cost {
     }
 
     /// Return whether the cost is within the budget of
-    /// the given transaction input.
+    /// the given script witness of a transaction input.
     #[cfg(feature = "elements")]
-    pub fn is_budget_valid(&self, txin: &elements::TxIn) -> bool {
+    pub fn is_budget_valid(&self, script_witness: &Vec<Vec<u8>>) -> bool {
```
apoelstra: In 11c65439e77b18d4255e5e756d307e735b284576

Even though we expect the user to just pass `txin.witness.script_witness`, we should take a `&[Vec<u8>]` rather than a `&Vec<Vec<u8>>`.

src/analysis.rs
```diff
@@ -99,17 +99,27 @@ impl Cost {
...
             .consensus_encode(&mut sink)
             .expect("writing to sink never fails");
-        let budget = witness_stack_serialized_len + 50;
-        // Cast safety: witness stack serialized length cannot be more than 2^32 - 51
-        self.less_equal_weight(budget as u32)
+        assert!(
+            witness_stack_serialized_len <= u32::MAX as usize, // cast safety: Assume 32-bit machine or higher
+            "Serialized witness stack must be shorter than 2^32"
+        );
+        let budget = witness_stack_serialized_len as u32; // cast safety: witness_stack_serialized_len <= u32::MAX
```
apoelstra, line 126: In dfb7f0be6fc35608cfdc9f8cc0a3e423995198b2:

We should use `u32::try_from` (or `usize::try_into` or whatever) and `expect` this, rather than doing an explicit assert followed by a cast.

src/analysis.rs
```diff
@@ -99,6 +99,19 @@ impl Cost {
...
+        // TODO: Wait for rust 1.73.0
+        // self.0.div_ceil(1000)
+        let base = self.0 / 1000;
+        let remainder = self.0 % 1000;
+        if remainder > 0 {
+            base.saturating_add(1)
+        } else {
+            base
+        }
+    }
```
apoelstra, line 113: In f09a045c552aaddda7181dd92b1e007e06bf9089:

This would be much clearer if the whole function were just `base.saturating_add(999) / 1000`. With a comment saying that `saturating_add` is just there to eliminate the panic path and that we don't expect it to ever saturate.

src/analysis.rs
```diff
@@ -134,6 +134,39 @@ impl Cost {
...
+        let required_budget = self.required_budget();
+        let current_budget = Self::get_budget(script_witness);
+        if required_budget <= current_budget {
+            return None;
+        }
+
+        debug_assert!(
+            required_budget > current_budget,
+            "Subtraction cannot overflow"
+        );
```
apoelstra, line 154: In 83b963e205a35b36f8c28b6991e595958e87b1fd:

I'd drop this `debug_assert`. You check this condition in the line immediately above and return `None`.

apoelstra: Or if you want to be fancy, you can do the subtraction with `checked_sub` and return `None` in the error case.

src/analysis.rs
```diff
@@ -134,6 +134,39 @@ impl Cost {
...
+        //
+        // 1. The encoded annex starts with the annex byte length
+        // 2. The first annex byte is always 0x50
+        //
+        // The remaining padding is done by adding (zero) bytes to the annex.
+        //
+        // Cast safety: assuming 32-bit machine or higher
+        let remaining_padding_len = required_budget.saturating_sub(2) as usize;
+        let mut annex_bytes = vec![0x50];
+        annex_bytes.extend(vec![0u8; remaining_padding_len]);
```
apoelstra, line 166: In 83b963e205a35b36f8c28b6991e595958e87b1fd:

nit: it is a little more efficient to create the vector with `Vec::with_capacity` since you know the total size upfront.

apoelstra:
f28ec2f353b158ed7d49d9c5b99bdedd12a97f2d looks great! All my comments are fairly minor.

src/analysis.rs
uncomputable: [consensus_encode](https://docs.rs/bitcoin/latest/bitcoin/consensus/encode/trait.Encodable.html#tymethod.consensus_encode) is not implemented for `&[Vec<u8>]`. This is why clippy didn't show a warning, which I was wondering about. I will update the commit message.

I could combine the encodings of each `Vec<u8>`, but this requires knowledge about the encoding.

src/analysis.rs
uncomputable: `self.0.saturating_add(999) / 1000` would be cleaner, but its semantics are different:
`4294967001.saturating_add(999) / 1000 = 4294967` while `4294967001.div_ceil(1000) = 4294968`. The correct budget is `ceil(4294967.001) = 4294968`, which is also what `required_budget` computes.

I think it is important to compute the budget correctly, even though it exceeds `CONSENSUS_MAX`. We can use `div_ceil` once we have Rust 1.73.0.

uncomputable:
Rebased and addressed some of the nits. Thanks for the helpful comments.

src/analysis.rs
apoelstra: Ah, I see.

In that case we should update the doccomment of the function.

src/analysis.rs
apoelstra: So, the only difference is when we're nearing u32::MAX? But even your code does the wrong thing when we are literally at the maximum.

apoelstra: If we're going to use `saturating_add` at all then we are accepting that it's ok to get wrong answers when we are near the limits of the type. And in this case, my code is clearer.

uncomputable: We can remove `saturating_add` in my code. It's unnecessary because we divided `base` by 1000.

I think `self.0.saturating_add(999) / 1000` is a good option.

Let me just clarify that my code is equivalent to `div_ceil` even at `u32::MAX`: [Rust playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=1d40eb1a7e7d30768e1ba202b924e3e0)

apoelstra: I believe that it's equivalent but it has probably 100x the cognitive load for readers of the code.

I'm fine with the existing code (though I'd prefer you remove `saturating_add` which is simply confusing, if it can never saturate) if you add a comment explaining that it's equivalent to the idiomatic `(x + 999) / 1000` except that it handles 999 more values correctly.

uncomputable:
Rebased and addressed the remaining comments.

apoelstra:
In a followup PR we should maybe replace `Vec<Vec<u8>>` with [`Witness`](https://docs.rs/bitcoin/latest/bitcoin/blockdata/witness/struct.Witness.html) which is more efficiently implemented (in particular it doesn't make you serialize to a sink to get its length).

But I think that would require changes in rust-elements first which maybe we don't want to bother with (in general, we are just keeping rust-elements on life support until rust-bitcoin 1.0 when we will need to more-or-less rewrite it to match the rust-bitcoin API).

apoelstra (approved):
ACK 7b4cd462c1d7c619fee0ff991ead107b4c3ac4b9

PR #180, opened 2023-11-25
Precise type splits

uncomputable:
Split sum and product types with separate methods for better readability.


apoelstra (approved):
ACK e0a2caed3342b9694a47a89fc88a2b34a368c069

uncomputable:
Awesome, thanks. The red icon looks scary, but the merge is what counts.

PR #182, opened 2023-12-05
Fix word parser

uncomputable:
The shift by 32 bits causes an overflow on 32-bit systems. AFAIU, the maximum word length is 2^31 bits, so this should be a shift by 31 bits.


apoelstra:
Good catch. I believe you are correct about the maximum.

PR #188, opened 2023-12-20
Run clippy on all targets

uncomputable:
Run clippy on all targets, fix lints and add to CI.


apoelstra (approved):
ACK 5502e5fbf333b3c0e2953b33b6e3238865cd3a3e

PR #190, opened 2024-01-05
build(windows): fix windows compilation errors

delta1:
fixes windows compilation errors

upstream fixes for simplicity-sys are at BlockstreamResearch/simplicity#213

### DIFF
--- a/simplicity-sys/build.rs
+++ b/simplicity-sys/build.rs
@@ -42,6 +42,7 @@ fn main() {
     let include = simplicity_path.join("include");
 
     cc::Build::new()
+        .std("c11")
         .flag_if_supported("-fno-inline-functions")
         .files(files)
         .files(test_files)


simplicity-sys/depend/wrapper.c
```diff
@@ -10,7 +10,7 @@
 
 #include <stdalign.h>
 
-const uint64_t c_overhead = overhead;
+const uint64_t c_overhead = 100;
```
delta1, line 13: this was necessary to resolve `[C2099: initializer is not a constant` ... 

happy to change it if there's a better way, but all the stuff i tried resulted in the same issue 

https://stackoverflow.com/questions/6131455/compile-error-c2099-initializer-is-not-a-constant

uncomputable: We will have to keep `c_overhead` in sync with `overhead`. Ideally this is automated with a script.

delta1: Agreed, I've added [a change upstream](https://github.com/BlockstreamResearch/simplicity/pull/213/commits/d3baff3578e361032d8732e2ba41d8459cc9d34e#diff-07c9188115e9cbfbfa2a581496c14eee1c64e590c13bc6a8bc2fccd1b1bdebbbR44) that works for this. So once [#213](https://github.com/BlockstreamResearch/simplicity/pull/213) is merged and we can revendor simplicity here, this will work. 

```suggestion
const uint64_t c_overhead = overhead_value;
```

delta1: fixed upstream, now unnecessary

uncomputable (approved):
ACK 44d0126bad306f65bea245c6a0c9a33b529dd4fd

PR #196, opened 2024-01-17
Release 0.2.0

uncomputable:
First release of rust-simplicity. Releases will make it easier to use rust-simplicity inside other projects such as elements-miniscript.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "simplicity"
-version = "0.1.0"
+version = "0.2.0"
 authors = ["Andrew Poelstra <apoelstra@wpsoftware.net>"]
 license = "CC0-1.0"
 edition = "2021"
@@ -23,13 +23,13 @@ elements = { version = "0.24.0", optional = true }
 hashes = { package = "bitcoin_hashes", version = "0.13" }
 hex = { package = "hex-conservative", version = "0.1.1" }
 santiago = "1.3"
-simplicity-sys = { version = "0.1.0", path = "./simplicity-sys" }
+simplicity-sys = { version = "0.2.0", path = "./simplicity-sys" }
 actual-serde = { package = "serde", version = "1.0.103", features = [
     "derive",
 ], optional = true }
 
 [dev-dependencies]
-simplicity-sys = { version = "0.1.0", path = "./simplicity-sys", features = [
+simplicity-sys = { version = "0.2.0", path = "./simplicity-sys", features = [
     "test-utils",
 ] }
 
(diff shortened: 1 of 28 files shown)


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "simplicity"
```
RCasatta, line 2: to release on crates we need a different name -> https://crates.io/crates/simplicity

uncomputable:
Depends on #198 and #199.

simplicity-sys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "simplicity-sys"
```
RCasatta, line 2: should we change also this in `simplicity-lang-sys` ?

uncomputable: I like `simplicity-sys` because it is shorter. There will be a lot of projects with the `simplicity` infix.

We invade the namespace of the existing `simplicity` project on crates.io, but the last update was 3 years ago so it seems pretty dead.

uncomputable:
@apoelstra Let's merge the release unless you see remaining roadblocks

RCasatta:
I think some mandatory fields in the `Cargo.toml` are missing such as `description` and `documentation` https://doc.rust-lang.org/cargo/reference/publishing.html#before-publishing-a-new-crate.

it can be tried out using `cargo publish --dry-run`

apoelstra:
On my end there are just warnings @RCasatta. I think we're ok.

apoelstra:
@uncomputable could you squash these three commits? (Or at least the last two; the middle one does not compile.)

uncomputable:
I squashed so each commit compiles.

I added the missing Cargo fields so `cargo publish` doesn't emit any warnings. Thank you @RCasatta for providing the command for dry runs!

I added the missing license to the `sys` crate. There were missing or long license headers, which I resolved. These changes belong to a separate PR, strictly speaking, but I feel like merging quickly :)

apoelstra:
We have a nice symmetry now with our workspaces being `simplicity-lang`, `simplicity-sys` and `simplicity-fuzz` :)

apoelstra:
I'm getting a bunch of clippy warnings, which maybe showed up since the edition-2021 change, but we can fix those in a followup PR.

uncomputable:
Unintended naming symmetry, but I like it, too

apoelstra (approved):
ACK 9bcef5155709e34b4b22379825cc10e280139d6d

apoelstra:
Tagged and published.

uncomputable:
Awesome. Cheers to the release 🎉

PR #198, opened 2024-01-23
Update cargo

uncomputable:
Updates to the Cargo and CI configuration.


.github/workflows/fuzz.yml
```diff
@@ -11,7 +11,7 @@ on:
 jobs:
   fuzz:
     if: ${{ !github.event.act }}
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
```
RCasatta, line 14: maybe personal taste but what do you think about updating to a fixed version `ubuntu-22.04` ?

uncomputable: We already use `ubuntu-latest` in our [main CI](https://github.com/BlockstreamResearch/rust-simplicity/blob/master/.github/workflows/main.yml#L8) so I would do the same for our fuzz CI.

[rust-bitcoin](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/.github/workflows/rust.yml#L13), [rust-secp](https://github.com/rust-bitcoin/rust-secp256k1/blob/master/.github/workflows/rust.yml#L8) and [rust-elements](https://github.com/rust-bitcoin/rust-bitcoin/blob/master/.github/workflows/rust.yml#L13) seem to be doing the same.

uncomputable:
Rebased

Cargo.toml
```diff
@@ -34,10 +34,7 @@ simplicity-sys = { version = "0.1.0", path = "./simplicity-sys", features = [
...
 [workspace]
-members = ["simpcli", "simplicity-sys"]
+members = ["simpcli", "simplicity-sys", "fuzz"]
 # Should be manually/separately tested since it has a massive dep tree
 # and not follow MSRV
-# FIXME we also need to include 'fuzz' in here because it currently uses
-# a patched bitcoin_hashes. When we remove that, we can move it back
-# to "members"
-exclude = ["jets-bench", "fuzz"]
+exclude = ["jets-bench"]
```
apoelstra, line 40: In 4a7c76c985a04e30e6995e3760e76e7d3a484565:

While we can do this, it means that we need `honggfuzz` (and all its dependencies) in any lockfile for this project. Generally people don't look inside their lockfiles so this is probably just an annoyance (it is for me, since I need to regenerate the lockfiles for my local CI :)) but if people *do* look at them, it's suddenly a much bigger file to check through.

uncomputable: I don't have strong opinions about this. Let's exclude `fuzz` and update explanation why it is excluded.

rust-bitcoin has a different folder structure, but `fuzz` seems to be a member of the workspace.

apoelstra: Ohh, actually, I do want us to un-exclude it, sorry :).

I am unable to run the fuzztests in my local CI because of the exclusion.

uncomputable: I am confused why CI worked with excluded `fuzz` on `master` and now it doesn't work, but I am not running your CI.

uncomputable:
Reverted to ff0dfd4 (`fuzz` included in workspace)

Cargo.toml
apoelstra: @uncomputable there are two separate issues:

* One is that whenever we change our Cargo.toml pretty-much at all, it forces me to regenerate Cargo.lock. (If we had lockfiles in the repo, like in rust-bitcoin or rust-secp256k1 or rust-bech32, I wouldn't need to maintain separate lockfiles and this wouldn't be an issue.) This "breakage" happens to me all the time.
* The other is that I can't test things that aren't workspaces (due to limits of crates2nix). So in this crate, I had just commented out the fuzztesting parts of my local CI. But now I can uncomment them :)

apoelstra (approved):
ACK ff0dfd44ac8d130588a805f179a4f3ff51197065

PR #201, opened 2024-01-29
Enable PRODUCTION flag

uncomputable:
The PRODUCTION flag needs to enabled for benchmarks and normal library usage. The flag needs to be disabled for fuzzing.


apoelstra (approved):
ACK ba30e2f579cfd992c8e841e4ee6b7b4931a56d48

PR #203, opened 2024-01-30
Update jets script

uncomputable:
Update the update jets script to handle wrapped jets.

Simplify the update jets script to use the currently checked-out version of libsimplicity instead of fiddling with git checkout. This might be controversial? The change makes my life easier. It might also break existing CI.

### DIFF
--- a/update_jets.sh
+++ b/update_jets.sh
@@ -1,33 +1,30 @@
 #!/bin/bash
 set -e
 
+# Requires (nix-shell with) cabal
 
 if [ -z "$1" ]; then
-  echo "\$1 parameter must be the simplicity directory"
-  echo "\$2 parameter (optional) can be the revision to check out. If this is not specified, we
-  use the simplicity revision specified in the simplicity-sys/depend/simplicity-HEAD-revision.txt.
-  If this is specified the revision must be available in the simplicity directory."
+  echo "\$1 path to libsimplicity repository root"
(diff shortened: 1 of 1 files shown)


apoelstra:
For sure, we shouldn't be running `git checkout` in peoples' repos like this. Probably removing that logic is an improvement in that it doesn't unexpectedly clobber stuff.

But I think you should preserve the warning. Rather than doing `git checkout "$REV"` just check what the checked-out rev is with `REV=$(git rev-parse HEAD)` and compare that to `VENDORED_HEAD`.

(And we should also check whether the repo is dirty, if there's a straightforward way to do that.

https://stackoverflow.com/questions/2657935/checking-for-a-dirty-index-or-untracked-files-with-git

suggests we can use `test -n "$(git status --porcelain)"` which will basically check whether `git status` runs clean.)

apoelstra:
1ef3ffc1fa1ad9c51f0767c79f0d0890fa599efd looks good other than that. I haven't tested locally; I suspect the new file will require a bit of tweaking to my local scripts.

uncomputable:
I added warnings about different libsimplicity versions and an unclean libsimplicity repo. The checks seem to work after a quick testing.

apoelstra (approved):
ACK 7556587051626eed04046c7c170703b5d493e571

PR #204, opened 2024-01-31
WASM FFI

uncomputable:
Upgrades the FFI to work with WASM.

Remove dependency on libc. Fix definition of `UWORD`, which is `u16` on WASM != `u32 = usize`. Prepare for wrapped jets.


simplicity-sys/src/types.rs
```diff
@@ -0,0 +1,61 @@
...
+        pub static WASM32_SIZE_T_ALIGN: c_uchar;
+
+        pub static WASM32_UNSIGNED_CHAR_SIZE: c_uchar;
+        pub static WASM32_UNSIGNED_CHAR_ALIGN: c_uchar;
+
+        pub static WASM32_PTR_SIZE: c_uchar;
+        pub static WASM32_PTR_ALIGN: c_uchar;
+    }
+    unsafe {
+        assert_eq!(size_of::<c_int>(), WASM32_INT_SIZE as usize);
```
apoelstra, line 51: In 04933cfb72dbb1ff31987f543f3611ed50524427:

I think we should actually make these checks unconditional. So rather than calling them `WASM32_` call them `C_` and always compare the size/alignment as computed by the C compiler against the ones implied by the `type size_t = usize` etc.

apoelstra: You can also unconditionally include `wasm.c` which I guess ought to be renamed to `ffi_checks.c` or something.

apoelstra: The include is in 076fe4a75f21c5fa62e37f5af0df8d2cde925db1. Also, that patch needs to come before the one that adds the unit tests.

apoelstra: Wait -- we already have all this stuff. The constants are in `depend/wrapper.c` (we need to add some new ones for int, pointers and size_t) (we really don't need to test `unsigned_char` this is guaranteed by the C standard to have size and alignment 1).

Then the tests are in `simplicity-sys/src/c_jets/mod.rs` `sanity_checks`.

Why add a new wasm.c rather than just extending the existing stuff?

uncomputable: I copied from rust-secp-sys, but it seems we already have some of the functionality. I will remove the duplicates tomorrow.

apoelstra: I think rust-secp-sys copied its stuff from here :)

uncomputable:
Rebased onto master. Extended the existing C type definitions and tests. Removed wrapper.h because the file will be vendored.

simplicity-sys/src/c_jets/jets_ffi.rs
```diff
@@ -1,6 +1,6 @@
 /* This file has been automatically generated. */
 
-use std::ffi::c_void;
+use crate::ffi::c_void;
```
apoelstra, line 3: In 8e7f59bcbdac781c0472d1643ef246436ab59e15:

You are not allowed to change this file directly. It needs to be changed upstream and then pulled in by the vendoring script.

uncomputable:
Let's merge this PR first and then match the changes in https://github.com/BlockstreamResearch/simplicity/pull/218

apoelstra:
My test suite will not pass if you are modifying files that originate in the upstream library.

PR #205, opened 2024-02-13
WASM FFI 2

uncomputable:
Make rust-simplicity compile in WASM.

Move test-only FFI bindings into separate module, separating them from jet-only FFI bindings. The latter compile in WASM while the former don't.

Define more C types (checked in WASM) and use these local types. Remove libc dependency entirely.

Redefine allocation in Rust.


apoelstra:
CI failure looks real.

uncomputable:
Fixed the fuzz tests. The import paths changed and `test-utils` needed to be activated.

apoelstra:
a37068e5c8219ba50f482ab591382a39319cbfef looks fine to me. Cannot test locally because #204 broke my local CI.

uncomputable:
Sorry for the broken CI :) I will get the Haskell Gen PR up to speed in the net 24 hours.

uncomputable:
Depends on #208. Rebased onto #208 and added allocation.

uncomputable:
Updated to use the allocator functions from https://github.com/BlockstreamResearch/simplicity/pull/222.

MSRV broke CI again. This error likely exists on master. I will work on a separate CI overhaul PR. In the meantime, let's ignore the CI failure and move forward with the present PR.

uncomputable:
CI is fixed in #210

uncomputable:
Rebased on master and updated lock file

simplicity-sys/src/alloc.rs
```diff
@@ -0,0 +1,104 @@
...
+    let ptr_prefix = allocator(layout);
+    if ptr_prefix.is_null() {
+        alloc::handle_alloc_error(layout);
+    }
+    // Write the number of allocated bytes to memory
+    (ptr_prefix as *mut usize).write(size_prefixed_bytes);
+    // Return a pointer to the size_bytes many allocated bytes behind the counter
+    // We need to offset the pointer by MIN_ALIGN to keep alignment
+    // This means there is a gap of MIN_ALIGN - sizeof(size_t) many unused bytes
+    // We asserted that MIN_ALIGN >= sizeof(size_t), so this gap is nonnegative
```
apoelstra, line 40: In 22c8741c74c73fab7588d969ec2d21fc6d3b95c3

Curiously, in rust-secp we actually *do* assert this, but here we only comment it. (I see that there's a unit test about this, but we should also put an assertion into `allocate` itself.)

uncomputable:
Added runtime assertions to the allocation function.

apoelstra (approved):
ACK 2b95398fc2f0786eb2effe3aabce5f60a5d70714

apoelstra:
Note that I had to explicitly whitelist the simplicity_alloc.h file when checking the diff between this repo and upstream. I think that's the right approach.

PR #207, opened 2024-02-14
Refactor exec_jec

uncomputable:
A small change that fits into no other PR.

### DIFF
--- a/src/bit_machine/mod.rs
+++ b/src/bit_machine/mod.rs
@@ -85,8 +85,8 @@ impl BitMachine {
     /// Drop the active read frame
     fn drop_frame(&mut self) {
         let active_read_frame = self.read.pop().unwrap();
-        self.next_frame_start -= active_read_frame.len;
-        assert_eq!(self.next_frame_start, active_read_frame.start);
+        self.next_frame_start -= active_read_frame.bit_width();
+        assert_eq!(self.next_frame_start, active_read_frame.start());
     }
 
     /// Write a single bit to the active write frame
--- a/src/bit_machine/mod.rs
+++ b/src/bit_machine/mod.rs
@@ -8,15 +8,16 @@
 
 mod frame;
 
+use std::fmt;
+use std::sync::Arc;
+use std::{cmp, error};
+
 use crate::analysis;
 use crate::dag::{DagLike, NoSharing};
 use crate::jet::{Jet, JetFailed};
 use crate::node::{self, RedeemNode};
 use crate::{Cmr, FailEntropy, Value};
 use frame::Frame;
-use std::fmt;
-use std::sync::Arc;
-use std::{cmp, error};
 
 /// An execution context for a Simplicity program
 pub struct BitMachine {
@@ -175,6 +176,19 @@ impl BitMachine {
                 Value::Prod(..) => {}
             }
         }
+    }
+
+    /// Return the bit width of the active read frame.
+    fn active_read_bit_width(&self) -> usize {
+        self.read.last().map(|frame| frame.bit_width()).unwrap_or(0)
+    }
+
+    /// Return the bit width of the active write frame.
+    fn active_write_bit_width(&self) -> usize {
+        self.write
+            .last()
+            .map(|frame| frame.bit_width())
+            .unwrap_or(0)
     }
 
     /// Add a read frame with some given value in it, as input to the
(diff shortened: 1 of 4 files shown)


uncomputable:
Added another bugfix that fits nowhere else.

src/bit_machine/mod.rs
```diff
@@ -365,67 +369,74 @@ impl BitMachine {
         }
     }
 
-    fn exec_jet<J: Jet>(&mut self, jet: J, env: &J::Environment) -> Result<(), JetFailed> {
-        use simplicity_sys::c_jets::frame_ffi::{c_readBit, c_writeBit, CFrameItem};
-        use simplicity_sys::c_jets::uword_width;
-        use simplicity_sys::ffi::UWORD;
+    /// Return a read frame that is equivalent to the active read frame.
+    fn get_input_frame(&mut self, bit_width: usize, buffer: &mut Vec<UWORD>) -> CFrameItem {
```
apoelstra, line 373: In 63c2d233fb29a8837e930d79507d89d06a4443d9:

`get_input_frame` and `get_output_frame` both need to be marked `unsafe` since they return a data structure containing raw pointers into a `Vec`, with no protection on those pointers being invalidated.

Probably these methods should just be defined inside of `exec_jet` since they're so wildly unsafe to use.

src/bit_machine/mod.rs
```diff
@@ -365,67 +369,74 @@ impl BitMachine {
...
+        let buffer_width = uword_width(bit_width);
+        buffer.clear();
+        buffer.resize(buffer_width, 0);
```
apoelstra, line 376: In 63c2d233fb29a8837e930d79507d89d06a4443d9:

I would instead accept a `&mut [u8]` and `assert_eq` that its length is correct, and make the caller do the resizing and initialization. Then at least we're not invalidating other `CFrameItem`s that refer to this object from within this function.

src/bit_machine/mod.rs
```diff
@@ -365,67 +369,74 @@ impl BitMachine {
...
 
-        if !res {
+        // Return output frame as write frame
+        let buffer_end = unsafe { buffer.as_mut_ptr().add(buffer_width) };
+        unsafe { CFrameItem::new_write(bit_width, buffer_end) }
+    }
+
+    /// Return the value of the given type that is contained in the buffer.
+    fn read_output_frame(&mut self, bit_width: usize, buffer: &mut Vec<UWORD>) {
```
apoelstra, line 406: In 63c2d233fb29a8837e930d79507d89d06a4443d9:

Here you can take a `&[UWORD]` and you don't even need `mut`.

src/bit_machine/mod.rs
uncomputable: I moved the new functions inside `exec_jet`. Because these functions are pure, we need to pass `self` as a parameter. I think this is a better solution than closures which cannot be unsafe.

src/bit_machine/mod.rs
uncomputable: I will look into this tomorrow

src/bit_machine/mod.rs
```diff
@@ -366,66 +367,79 @@ impl BitMachine {
...
 
     fn exec_jet<J: Jet>(&mut self, jet: J, env: &J::Environment) -> Result<(), JetFailed> {
-        use simplicity_sys::c_jets::frame_ffi::{c_readBit, c_writeBit, CFrameItem};
-        use simplicity_sys::c_jets::uword_width;
-        use simplicity_sys::ffi::UWORD;
+        use crate::ffi::c_jets::frame_ffi::{c_readBit, c_writeBit, CFrameItem};
+        use crate::ffi::c_jets::uword_width;
+        use crate::ffi::ffi::UWORD;
+
+        /// Return a read frame that is equivalent to the active read frame.
```
apoelstra, line 374: In 494a232fa8e6144be845940a1bfe22c88e98359e:

I'm much happier now that this lives inside `exec_jet` and is clearly a single-use function. But we should improve the comment still, to say something like
```
/// Creates a new read frame, backed by `buffer`, and copy the `bit_width` bits from the active read frame into it.
///
/// The value of `buffer`, including possibly its backing store, will be lost and no pointers
/// to it should be retained. Similarly the returned frame should not outlive `buffer` or
/// any modifications to it.
///
/// `bit_width` should be the width of the active read frame.
```

src/bit_machine/mod.rs
```diff
@@ -366,66 +367,79 @@ impl BitMachine {
...
-        let src_ty_bit_width = jet.source_ty().to_bit_width();
-        let target_ty_bit_width = jet.target_ty().to_bit_width();
-
-        let a_frame_size = uword_width(src_ty_bit_width);
-        let b_frame_size = uword_width(target_ty_bit_width);
-        // a_frame_size + b_frame_size must be non-zero unless it is a unit to unit jet
-        if a_frame_size == 0 && b_frame_size == 0 {
-            return Ok(());
+
+        /// Return a write frame that is equivalent to the active write frame.
```
apoelstra, line 398: In 494a232fa8e6144be845940a1bfe22c88e98359e:

This comment should just say "returns a new output frame backed by `buffer` of width `bit_width`". And have similar warnings about the safety of `buffer` and maybe say that `bit_width` is expected to be the size of the active write frame.

src/bit_machine/mod.rs
```diff
@@ -366,66 +367,79 @@ impl BitMachine {
...
-        // 1. Write into C bitmachine using A write frame(= src_ptr_end)
-        // Precondition satisfied: src_ptr_end is one past the end of slice of UWORDs for A.
-        let mut a_frame = unsafe { CFrameItem::new_write(src_ty_bit_width, src_ptr_end) };
-        for _ in 0..src_ty_bit_width {
-            let bit = self.read_bit();
-            unsafe {
-                c_writeBit(&mut a_frame, bit);
+
+        /// Return the value of the given type that is contained in the buffer.
+        fn read_output_frame(mac: &mut BitMachine, bit_width: usize, buffer: &[UWORD]) {
```
apoelstra, line 410: In 494a232fa8e6144be845940a1bfe22c88e98359e:

Could this simply be a method on `Frame`? The frame itself should know its width now.

As written, is this unsafe if the user passes too large a value for `bit_width`?

uncomputable: `Frame`s don't own their data so we would have to pass the data as a parameter, which is messy.

uncomputable:
Rebased on master. Creating the buffers inside the helper methods is much cleaner than passing mutable references. Added assertions that the active read / write frame are large enough. Resolved misunderstanding that active read / write frame is as wide as jet source / target type.

src/bit_machine/mod.rs
```diff
@@ -366,66 +380,93 @@ impl BitMachine {
...
+        use crate::ffi::ffi::UWORD;
+
+        /// Create new C read frame that contains `bit_width` many bits from active read frame.
+        ///
+        /// Return C read frame together with underlying buffer.
+        ///
+        /// ## Panics
+        ///
+        /// Active read frame has fewer bits than `bit_width`.
+        unsafe fn get_input_frame(
```
apoelstra, line 394: In b561354638232b5884f8197a47e8cfda415a16f5:

Ok, awesome, this is looking much cleaner.

Should add a `## Safety` section to the doc simply saying that it returns a `CFrameItem` which points into the returned `Vec` and that the `Vec` must outlive the `CFrameItem`.

*Alternately* we could create a new opaque type `struct BackedCFrameItem` which holds the `Vec`, implements `Deref<CFrameItem>`, and which would then ensure that you were obeying the lifetime rules. I *think* that if you did this, you could eliminate `unsafe` from this method.

src/bit_machine/mod.rs
```diff
@@ -366,66 +380,93 @@ impl BitMachine {
...
-            let bit = self.read_bit();
-            unsafe {
-                c_writeBit(&mut a_frame, bit);
+
+        /// Write `bit_width` many bits from `buffer` into active write frame.
+        ///
+        /// ## Panics
+        ///
+        /// Active write frame has fewer bits than `bit_width`.
+        fn update_active_write_frame(mac: &mut BitMachine, bit_width: usize, buffer: &[UWORD]) {
```
apoelstra, line 437: In b561354638232b5884f8197a47e8cfda415a16f5:

Here you still need `unsafe` because the passed buffer might now be wide enough for your `c_readBit` calls. I think you can add another `assert!` that `buffer.len()` is the right length to avoid this unsafety.

src/bit_machine/mod.rs
uncomputable: Ok, will do. I haven't been working much with unsafe, so these comments on how to write unsafe Rust (somewhat) safely are appreciated :)

apoelstra (approved):
ACK 44d20a96f123fc48f7b3007b962a8e7f8cb7cc99

PR #208, opened 2024-02-16
Jet wrappers

uncomputable:
Update libsimplicity and use generated jet wrappers.


simplicity-sys/vendor-simplicity.sh
```diff
@@ -1,45 +1,47 @@
...
+echo "$HEAD" >> "$ORIGINAL_PATH/$DEPEND_PATH/simplicity-HEAD-revision.txt"
 
-# Only get the C directory. move C into parent folder, and then move it back
-mv "${DIR}/C" "./${DIR}_C"
-rm -rf "${DIR}"
-mv "${DIR}_C" "${DIR}"
+# Copy C folder to simplicity-sys/depend/simplicity
+# Use rsync to copy only files tracked by git
+cd C
+git ls-files | rsync -av --files-from=- . "$ORIGINAL_PATH/$DEPEND_PATH/$VENDORED_SIM_DIR"
```
apoelstra, line 40: In 089847766a3ce08527511548c27a67895cb8d691:

What do you mean "only files tracked by git"? rsync does not understand git and you haven't given it any interesting filtering options.

apoelstra: Oh, derp, you're piping `ls-files` into `rsync`. Interesting.

I think I'm OK with the rsync dependency then.

simplicity-sys/vendor-simplicity.sh
```diff
@@ -1,45 +1,47 @@
 #!/bin/bash
```
apoelstra, line 1: In 089847766a3ce08527511548c27a67895cb8d691:

If you want this to be nix-compatible we have to replace `/bin/bash` with `/usr/bin/env bash`.

apoelstra (approved):
ACK 6a7684197f119c1fc15d9eff7454e394adb599ce

uncomputable:
I fixed the bash shebang. I also fixed the header of the autogenerated HEAD revision, which was different depending on the location of the vendor script relative to pwd.

apoelstra (approved):
ACK c9d5c00bb8d6bdbaf7d328ad00a11af3f6e12a3c

PR #209, opened 2024-02-22
Better human

### LINKED ISSUE #181: Incorrect hole naming
The hole in this program is named `disc1` instead of `hole`. The following code runs:

```
let s = "
    id1 := iden : 2^256 * 1 -> 2^256 * 1
    disc1 := unit
    main := comp (disconnect id1 ?hole) unit -- fixme: ?hole is named disc1
";
let empty_witness = HashMap::new();
let forest = Forest::parse(s).unwrap();
let program = forest.to_witness_node(&empty_witness).finalize().unwrap();
```

Replacing `disc1` with `hole` makes the code fail.

uncomputable:
Fixes #181

### DIFF
--- a/src/human_encoding/mod.rs
+++ b/src/human_encoding/mod.rs
@@ -339,6 +339,20 @@ mod tests {
     }
 
     #[test]
+    fn filled_hole() {
+        let empty = HashMap::new();
+        assert_finalize_ok::<Core>(
+            "
+                id1 := iden : 2^256 * 1 -> 2^256 * 1
+                main := comp (disconnect id1 ?hole) unit
+                hole := unit
+            ",
+            &empty,
+            &(),
+        );
+    }
+
+    #[test]
     fn unfilled_hole() {
         let empty = HashMap::new();
         assert_finalize_err::<Core>(
--- a/src/human_encoding/mod.rs
+++ b/src/human_encoding/mod.rs
@@ -208,6 +208,9 @@ impl<J: Jet> Forest<J> {
         ret
     }
 
+    /// Convert the forest into a witness node.
+    ///
+    /// Succeeds if the forest contains a "main" root and returns `None` otherwise.
     pub fn to_witness_node(
         &self,
         witness: &HashMap<Arc<str>, Arc<Value>>,
--- a/src/human_encoding/mod.rs
+++ b/src/human_encoding/mod.rs
@@ -249,29 +249,6 @@ mod tests {
     }
 
     #[test]
-    fn parse_duplicate_witness_in_disconnected_branch() {
-        let s = "
-            wit1 := witness
-            main := comp wit1 comp disconnect iden ?dis2 unit
-
-            wit1 := witness
-            dis2 := wit1
-        ";
-
-        match Forest::<Core>::parse(s) {
-            Ok(_) => panic!("Duplicate witness names should fail"),
-            Err(set) => {
-                let errors: Vec<_> = set.iter().collect();
-                assert_eq!(1, errors.len());
-                match errors[0] {
-                    super::error::Error::NameRepeated { .. } => {}
-                    error => panic!("Unexpected error {}", error),
-                }
-            }
-        }
-    }
-
-    #[test]
     fn to_witness_node_unfilled_hole() {
         let s = "
             wit1 := witness
--- a/src/human_encoding/mod.rs
+++ b/src/human_encoding/mod.rs
@@ -223,76 +223,107 @@ impl<J: Jet> Forest<J> {
 #[cfg(test)]
 mod tests {
     use crate::human_encoding::Forest;
-    use crate::jet::Core;
-    use crate::{Error, Value};
+    use crate::jet::{Core, Jet};
+    use crate::{BitMachine, Value};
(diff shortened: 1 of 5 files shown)


src/human_encoding/named_node.rs
```diff
@@ -33,7 +33,7 @@ pub struct Named<N> {
 impl<J: Jet> node::Marker for Named<Commit<J>> {
     type CachedData = NamedCommitData<J>;
     type Witness = <Commit<J> as node::Marker>::Witness;
-    type Disconnect = <Commit<J> as node::Marker>::Disconnect;
+    type Disconnect = Arc<str>;
     type SharingId = Arc<str>;
```
apoelstra, line 37: in 428fdf3628534e4677e31f9f37075f446901b4ef:

Can wait for a future PR but these `Arc<str>`s should both be newtypes I think.

src/human_encoding/mod.rs
```diff
@@ -223,76 +223,108 @@ impl<J: Jet> Forest<J> {
...
+    ) {
+        let program = match Forest::<J>::parse(s)
+            .expect("Failed to parse human encoding")
+            .to_witness_node(witness)
+            .expect("Forest is missing expected root")
+            .finalize()
+        {
+            Ok(program) => program,
+            Err(error) => {
+                assert_eq!(error.to_string().as_str(), err_msg);
```
apoelstra, line 260: In 60efcbf0a34204b28bc063804df609f523374c46:

You definitely do not need to write `to_string().as_str()` anywhere. Use `&error.to_string()`.

src/human_encoding/mod.rs
```diff
@@ -223,76 +223,108 @@ impl<J: Jet> Forest<J> {
...
-
-        witness.insert(Arc::from("wit2"), Value::unit());
+        let wit2_is_pruned = HashMap::from([(Arc::from("wit1"), Value::u1(0))]);
+        assert_finalize_ok::<Core>(s, &wit2_is_pruned, &());
+
+        let wit2_is_missing = HashMap::from([(Arc::from("wit1"), Value::u1(1))]);
+        // FIXME The finalization should fail
+        // This doesn't happen because we do bottom-up pruning:
+        // Unpopulated witness nodes are pruned, regardless of whether they will be executed
+        // Top-down pruning would execute the program and throw an error at unpopulated witness nodes
```
apoelstra, line 327: In 60efcbf0a34204b28bc063804df609f523374c46:

I think this comment is confusingly worded. Really the problem is that we don't execute during finalization, so we can't determine in general exactly which nodes need to be pruned. "Top-down" vs "bottom-up" is beside the point.

My original plan was to just add an execution pass at the end of `finalize()` with some extra context passed into the execution engine to track which nodes should be pruned. Maybe you have a more efficient idea?

apoelstra:
Done reviewing 2c12d1272522fba763475a0ad5c29f92ef65653f. Overall these changes are great, thank you!

src/human_encoding/named_node.rs
uncomputable: We can introduce the newtype `HoleName` and use it in `WitnessOrHole` and in `NamedCommitNode::Disconnect`. `Arc<str>` is used in 428fdf3628534e4677e31f9f37075f446901b4ef only for hole names, unless I am missing something.

src/human_encoding/mod.rs
uncomputable: Running the Bit Machine at the end of the current finalization method would work.

In the future, I think we can drop the existing pruning and retyping, to simplify the algorithm. Bit Machine execution is necessary in general and sufficient to detect all unused branches.

We finalize the types of the shared program and run the Bit Machine to detect unused branches. We might need to re-run the program because types changed. TBH, I am still confused how much the types can change in each pruning iteration and what the worst case is, but the execution-only algorithm should work.

uncomputable:
For now I fixed all nits except for the newtype.

src/human_encoding/named_node.rs
apoelstra: Yep, I believe you introduce and use it only for hole names.

We should make a second newtype for the `SharingId`. That way we can't get them confused when we're threading all this stuff through the abstract node infrastructure.

src/human_encoding/mod.rs
apoelstra: We shouldn't need to re-run the program. After pruning we should find that some times have shrunk, but only to eliminate entirely unused branches of types.

One complication is that we need to parse witnessess knowing that the user might provide a larger witness than is needed. For example, you could imagine a witness node that could take a type `Prod(X, Y)` but only one of the `X` or `Y`s are actually used depending on some other witness bit. Ideally we would be ok if the user actually provided both, and as part of our pruning we would shrink the witness. (The case where the user wants to provide only one, because he knows which one will actually be used, is a bit trickier. Initially I think we will need to just request he fill out the "whole type" perhaps by adding dummy data.)

But we could add a `debug_assert` or something that re-running the program succeeds and does not result in any further pruning.

apoelstra (approved):
ACK 05c81baef9e6ecd523994876f7b5a2023059c133

PR #210, opened 2024-03-14
Locked CI

uncomputable:
Add a lock file that works with all supported rust versions. Use this lock file in CI.


apoelstra (approved):
ACK 207fd32292229fd01110077ea64de0e4212165ea

PR #212, opened 2024-03-15
Typed value

uncomputable:
Add type checker to values and use it in the Bit Machine. This is particularly useful for running expressions with non-unit input types.

### DIFF
--- a/src/bit_machine/mod.rs
+++ b/src/bit_machine/mod.rs
@@ -16,6 +16,7 @@ use crate::analysis;
 use crate::dag::{DagLike, NoSharing};
 use crate::jet::{Jet, JetFailed};
 use crate::node::{self, RedeemNode};
+use crate::types::Final;
 use crate::{Cmr, FailEntropy, Value};
 use frame::Frame;
 
@@ -30,6 +31,8 @@ pub struct BitMachine {
     read: Vec<Frame>,
     /// Write frame stack
     write: Vec<Frame>,
+    /// Acceptable source type
+    source_ty: Arc<Final>,
 }
 
 impl BitMachine {
@@ -42,6 +45,7 @@ impl BitMachine {
             next_frame_start: 0,
             read: Vec::with_capacity(program.bounds().extra_frames + analysis::IO_EXTRA_FRAMES),
             write: Vec::with_capacity(program.bounds().extra_frames + analysis::IO_EXTRA_FRAMES),
+            source_ty: program.arrow().source.clone(),
         }
     }
 
@@ -193,11 +197,14 @@ impl BitMachine {
 
     /// Add a read frame with some given value in it, as input to the
     /// program
-    pub fn input(&mut self, input: &Value) {
-        // FIXME typecheck this
+    pub fn input(&mut self, input: &Value) -> Result<(), ExecutionError> {
+        if !input.is_of_type(&self.source_ty) {
+            return Err(ExecutionError::InputWrongType(self.source_ty.clone()));
+        }
         self.new_frame(input.len());
         self.write_value(input);
         self.move_frame();
+        Ok(())
     }
 
     /// Execute the given program on the Bit Machine, using the given environment.
@@ -484,6 +491,8 @@ impl BitMachine {
 /// Errors related to simplicity Execution
 #[derive(Debug)]
 pub enum ExecutionError {
+    /// Provided input is of wrong type
+    InputWrongType(Arc<Final>),
     /// Reached a fail node
     ReachedFailNode(FailEntropy),
     /// Reached a pruned branch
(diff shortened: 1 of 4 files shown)


src/value.rs
```diff
@@ -91,6 +91,35 @@ impl Value {
             .count()
     }
 
+    /// Check if the value is a unit.
+    pub fn is_unit(&self) -> bool {
+        matches!(self, Value::Unit)
+    }
+
+    /// Unwrap a left sum value and return `None` on other values.
+    pub fn unwrap_left(&self) -> Option<Arc<Value>> {
```
apoelstra, line 106: In e10f2df43b42b5865f5cd7be6ecd11d3f52cb498:

`unwrap_` is a super confusing name for a method that returns an `Option`. I would suggest calling it `to_left`.

You could also return an `Option<&Arc<Value>>`, avoid the `Arc::clone`, and call it `as_left`.

apoelstra: I think we should rename the existing `split_sum` and `split_product` to `as_sum` and `as_product` as well. (Or `to_sum`/`to_product` if we are returning non-borrowed Arcs.)

src/value.rs
```diff
@@ -293,6 +294,36 @@ impl Value {
 
         (bytes, bit_length)
     }
+
+    /// Check if the value is of the given type.
+    pub fn is_of_type(self: Arc<Self>, ty: Arc<Final>) -> bool {
```
apoelstra, line 304: In d9ca9252c58e16c6375d4d570020d77b6b3c96a6:

I'm fairly sure we could take an `&Arc<Self>` and `&Arc<Final>` here and save a bunch of clones. We would also want `split_*`/`unwrap_*` to return references.

Personally I would suggest having a family `to_sum`/`to_left`/etc which returns `Arc<Self>` as they do now, and also a family `as_sum`/`as_left`/etc which return `&Self`. I think you can then use the `as_` methods, have your stack be holding references rather than Arcs, and save yourself a lot of atomic refcounts.

src/bit_machine/mod.rs
```diff
@@ -179,11 +183,14 @@ impl BitMachine {
 
     /// Add a read frame with some given value in it, as input to the
     /// program
-    pub fn input(&mut self, input: &Value) {
-        // FIXME typecheck this
+    pub fn input(&mut self, input: Arc<Value>) -> Result<(), ExecutionError> {
+        if !input.clone().is_of_type(self.source_ty.clone()) {
```
apoelstra, line 187: In f45da7ac65eb7843e2a5875f3b3414298220477c

...then both these `.clone`s can go away.

apoelstra: And the function can keep taking an `&Value`.

src/bit_machine/mod.rs
```diff
@@ -187,9 +187,12 @@ impl BitMachine {
         if !input.clone().is_of_type(self.source_ty.clone()) {
             return Err(ExecutionError::InputWrongType(self.source_ty.clone()));
         }
-        self.new_frame(input.len());
-        self.write_value(&input);
-        self.move_frame();
+        // Unit value doesn't need extra frame
```
apoelstra, line 190: In 440298b621bf0f9f0e68544fe7005494d0581891:

This optimization is fine but where is the panic that you claim exists in the commit message?

apoelstra:
Done reviewing 564fa2566b4a65705c29af0c03d20fa929188c5a

src/bit_machine/mod.rs
uncomputable: The panic occurs during execution, a few lines down the code. The error message is "Program requires a non-empty input to execute". This error is triggered if there is a frame on the input stack (unit value) but the input width is zero (unit type). The check is made less confusing in 564fa2566b4a65705c29af0c03d20fa929188c5a.

src/value.rs
uncomputable: I've been wrestling with the question `Arc<Self>` vs `&Self` for a while. Sometimes we need an owned `Arc<Self>`. Sometimes we want to avoid unnecessary clones using `&Self`. Supporting both sounds like a good compromise.

uncomputable:
I changed the `split_*` and `unwrap_*` methods to `as_*` methods. References are returned. Right now we don't need any owned `Arc<Self>`s.

src/bit_machine/mod.rs
apoelstra: I think the assertion was just wrong. It is very weirdly structured. But in any case it is removed in 563fa25 as you say.

src/bit_machine/mod.rs
```diff
@@ -193,11 +197,14 @@ impl BitMachine {
 
     /// Add a read frame with some given value in it, as input to the
     /// program
-    pub fn input(&mut self, input: &Value) {
-        // FIXME typecheck this
+    pub fn input(&mut self, input: Arc<Value>) -> Result<(), ExecutionError> {
+        if !input.clone().is_of_type(self.source_ty.as_ref()) {
```
apoelstra, line 201: In 23bdebe5ef4f2c4dc622ba0bb1d2c783839d887b:

Please remove this `clone`. Also you shouldn't need to explicitly call `as_ref`. Just pass `&self.source_ty`.

apoelstra:
Done reviewing d81cf5aa196d9b8e3835dd0387728a44eb90d905. Much better!

uncomputable:
Rebased onto master and fixed the `input` method.

src/bit_machine/mod.rs
uncomputable: I overlooked this instance of `clone`. Thanks for pointing this out.

PR #214, opened 2024-04-23
Fix AMR

uncomputable:
Fix an error that was introduced in #212. Fix clippy.

### DIFF
--- a/simplicity-sys/src/alloc.rs
+++ b/simplicity-sys/src/alloc.rs
@@ -7,6 +7,7 @@ use std::mem;
 /// [16 bytes is the greatest alignment of any architecture Rust currently supports](https://github.com/rust-lang/rust/blob/61223975d46f794466efa832bc7562b9707ecc46/library/std/src/sys/pal/common/alloc.rs).
 #[repr(align(16))]
 #[derive(Default, Copy, Clone)]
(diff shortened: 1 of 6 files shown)


src/bit_encoding/decode.rs
```diff
@@ -236,7 +236,7 @@ pub fn decode_expression<I: Iterator<Item = u8>, J: Jet>(
         converted.push(new);
     }
 
-    converted[len - 1].get().map(Arc::clone)
+    converted[len - 1].get().cloned()
```
apoelstra, line 239: Aside from violating our MSRV, this code is objectively worse because it hides the fact that the `clone` is merely an arc clone.

If clippy is really complaining about this, you probably need to file separate bugs about (a) the lint not respecting MSRV and (b) the lint being stupid.

uncomputable: I was struggling to write `.cloned()` in a way that includes `Arc`. The lint is stupid and I disabled it now.

apoelstra:
Yeah, I think the design of `Literal` did change. I believe this was a very early idea for doing `disconnect` that turned out not to be sensible.

apoelstra (approved):
ACK 32d30e4c7c344753ed86195958c81598f54ff7ee

PR #215, opened 2024-05-08
Fix jets

uncomputable:
Fix two things to make jets more usable.

### DIFF
--- a/src/bit_machine/mod.rs
+++ b/src/bit_machine/mod.rs
@@ -477,7 +477,7 @@ impl BitMachine {
         let (mut output_write_frame, output_buffer) = unsafe { get_output_frame(output_width) };
 
         let jet_fn = jet.c_jet_ptr();
-        let c_env = jet.c_jet_env(env);
+        let c_env = J::c_jet_env(env);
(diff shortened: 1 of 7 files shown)


uncomputable:
The `c_jet_env` is useful in generic code over `J: Jet`. It is the only way to transform `J::Environment` to `J::CJetEnvironment`. I use `c_jet_env` inside the web IDE to implement jets.

It would be cleaner to enforce a trait for `J::Environment` which provides `c_jet_env` as a method.

apoelstra:
Ok, I understand. We'll keep it as-is for now. It's not obvious that this method should be on `J::Environment` when plausibly the same environment type corresponds to different C environments, depending on the jet generic.

apoelstra:
f98a088856008e1279eaadea75fd3d100965f4a1 is fine but you're not allowed to edit the Haskell-generated files directly like this.

uncomputable:
Oops, I didn't realize that I am touching auto-generated files. I will make a PR to libsimplicity.

uncomputable:
Depends on https://github.com/BlockstreamResearch/simplicity/pull/230

PR #216, opened 2024-05-08
Display expression

uncomputable:
Display Simplicity expressions as linear strings, like in the tech report. Include some simple transformations like infix `pair` and `comp`, and `true` or `false`, and the `{O,I,H}` notation. This will make debugging much easier.


apoelstra (approved):
ACK 4aa2fc5d2d95941916d7b2052863ae193b892abc

PR #217, opened 2024-05-27
Construct Types

uncomputable:
Make the constructors for `Final` public and update documentation.

I would like these constructors to return `Self` instead of `Arc<Self>`, because then the caller can choose how he wants to save the returned value. However, we would need to change the union bound code for that, which seems too invasive for now.


apoelstra (approved):
ACK c9276b48438c5ee1499ed5bc72fe9ba8e86479dd

PR #221, opened 2024-06-25
Cap maximum depth of incomplete type display

### LINKED ISSUE #177: Infinite finalization loop
If I try to decode the bytes of the program `case (drop iden) iden` from the [occurs check test](https://github.com/BlockstreamResearch/simplicity/blob/69eb1719b6a13a4346abde9b221eebc99002cc6f/C/test.c#L278), the binary gets stuck at the line that does the decoding.

```
let program_bytes = vec![0xc1, 0x07, 0x20, 0x30];
let mut bits = simplicity::BitIter::new(program_bytes.iter().copied());
let program = simplicity::CommitNode::<simplicity::jet::Core>::decode(&mut bits).expect("decode");
/// This is never printed
println!("{}", program.cmr());
```

apoelstra:
This limit may still be too high -- I think you can output a 2^64 sized type which will wreck any terminal. But at least it's finite now.

Every instance of `_iter` methods called on an incomplete type (in the types/ module anyway) now has a maximum depth on it, which should eliminate all the places where we might loop forever with incomplete types.

Fixes #177

### DIFF
--- a/src/node/commit.rs
+++ b/src/node/commit.rs
@@ -538,4 +538,41 @@ mod tests {
         // I created this unit test by hand
         assert_program_not_deserializable::<Core>(&[0xa9, 0x48, 0x00], &Error::NotInCanonicalOrder);
     }
+
+    #[test]
+    fn regression_177() {
+        // `case (drop iden) iden` from upstream occurs-check test. Has an infinitely sized
+        // input type. Will fail trying to unify the input type with the unit type before
+        // doing the occurs check, putting an infinitely-sized type into the error variant.
+        //
+        // The human-readable encoding will keep going and then also hit the occurs check.
+        //
+        // Check that both error types can be generated and printed in finite space/time.
+        let bad_prog = "
+            id := iden
+            main := case (drop id) id
+        ";
+        match Forest::<Core>::parse(bad_prog) {
+            Ok(_) => panic!("program should have failed"),
+            Err(set) => {
+                let mut errs_happened = (false, false);
+                for err in set.iter() {
+                    match err {
+                        crate::human_encoding::Error::TypeCheck(e @ types::Error::Bind { .. }) => {
+                            errs_happened.0 = true;
+                            e.to_string();
+                        }
+                        crate::human_encoding::Error::TypeCheck(
+                            e @ types::Error::OccursCheck { .. },
+                        ) => {
+                            errs_happened.1 = true;
+                            e.to_string();
+                        }
+                        x => panic!("unexpected error {x:?}"),
+                    }
+                }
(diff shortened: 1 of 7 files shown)


src/node/commit.rs
```diff
@@ -538,4 +538,18 @@ mod tests {
...
         assert_program_not_deserializable::<Core>(&[0xa9, 0x48, 0x00], &Error::NotInCanonicalOrder);
     }
+
+    #[test]
+    fn regression_177_1() {
+        // `case (drop iden) iden` from upstream occurs-check test. Has an infinitely sized
+        // input type. Will fail trying to unify the input type with the unit type *before*
+        // doing the occurs check, putting an infinitely-sized type into the error variant.
+        // Check that this error type can be printed in finite space/time.
+        let mut iter = BitIter::from([0xc1, 0x07, 0x20, 0x30].iter().copied());
```
uncomputable, line 548: 50293ca9872e6ad6dddd42aa020f17a7d43336e6 This test case would be better maintainable if we wrote it in the human encoding. We should strive to remove all byte blobs from the code base.

src/node/commit.rs
```diff
@@ -538,4 +538,18 @@ mod tests {
         // I created this unit test by hand
         assert_program_not_deserializable::<Core>(&[0xa9, 0x48, 0x00], &Error::NotInCanonicalOrder);
     }
+
+    #[test]
+    fn regression_177_1() {
```
uncomputable, line 543: 50293ca9872e6ad6dddd42aa020f17a7d43336e6: Adding this test case _before_ the infinite loop is fixed means that CI runs forever on this commit. @apoelstra Does your local CI have a timeout?

uncomputable (changes requested):
868964a912ccdcddae7f230f8dec831c69d4f258 Looks great, but let's move the test commit to the back.

src/node/commit.rs
apoelstra: Oh, lol, let me reorder the commits. I don't know if I have a timeout, but if I do, it is in exceess of an hour..

src/node/commit.rs
apoelstra: I'll take a look, but our options for decoding `ConstructNode`s are pretty limited.

apoelstra:
Reordered commits and rewrote the regression test to use the human-readable encoding. Conveniently with the human-readable encoding I can test *both* the bind error and the occurs-check error at once, since it attempts to continue after the first error.

uncomputable (approved):
ACK 66b2ddb2792b373853786b116d00782bde27248c

PR #222, opened 2024-06-25
occurs check: replace exponential-time algorithm with linear-time one

apoelstra:
Builds on #221, because the changes to the `OccursCheck` error variant would conflict with it if I PR'd it separately. But this is conceptually independent of #221.

Our old occurs-check loop would use the pre-order iterator from dag.rs with the `NoSharing` adaptor. This would iterate over the full explicit type, which may have exponential size in the program length. The correct algorithm notices when it is done checking a type bound and all its children, and does not check that bound again.

I wasn't able to produce a test vector showing the exponential-time behavior of the old algorithm, but I am hitting this issue in a fuzzing project I'm working on locally, and I think it's clear from the diff what the issue is and how this change fixes it.

### DIFF
--- a/src/dag.rs
+++ b/src/dag.rs
@@ -278,13 +278,15 @@ pub trait DagLike: Sized {
     /// yielded, you may be better off using this iterator instead.
     fn verbose_pre_order_iter<S: SharingTracker<Self> + Default>(
         self,
+        max_depth: Option<usize>,
     ) -> VerbosePreOrderIter<Self, S>
     where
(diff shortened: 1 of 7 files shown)


uncomputable:
Superseded by #221 ?

apoelstra:
No, but it needs to be rebased on #221.

apoelstra:
Oh, no, I guess I accidentally(?) pulled this commit into #221. It is superseded.

PR #223, opened 2024-06-25
minor fixes for recent rustcs

apoelstra:
I recently updated my local CI to run nightly clippy and check doc links. There are a few things that need to be cleaned up here to pass.


uncomputable (approved):
ACK 3418a6c89750e9b980fc203735be3c75a39d8cbe

PR #225, opened 2024-06-28
types: avoid accessing BoundMutex::inner in derived Debug

### LINKED ISSUE #224: Deadlock in type inference
If you add the following unit test to src/types/mod.rs it will deadlock while printing:

```rust
    #[test]
    fn memory_leak() {
        let iden = Arc::<WitnessNode<Core>>::iden();
        let drop = Arc::<WitnessNode<Core>>::drop_(&iden);
        let case = Arc::<WitnessNode<Core>>::case(&iden, &drop).unwrap();

        println!("{:?}", case.arrow());
    }
```

Occurs even with #221 and #222 . The issue here is not that the types are logically infinite (though I think they are) but that we have a actual refcount cycle in our `Arc<Mutex>`es. This leads to a deadlock in `BoundMutex::get`.

If you don't try to print the types, the test will complete, but it will leak memory, which can be seen by running the test in valgrind.

Investigating.

apoelstra:
We have a submodule `bound_mutex` whose purpose is to ensure that no deadlocks are possible. We do this by having a short submodule which only accesses the BoundMutex::inner member in the `get` and `set` methods. These methods lock the mutex and immediately unlock it. The member is never accessed by any other method. This is easy to check by searching the code for `inner`.

HOWEVER, there is actually a hidden access to `inner` in the `#[derive(Debug)]` line on `BoundMutex`, and this access is incorrect. Rather than locking the mutex, cloning the Arc within, unlocking, then processing the cloned Arc, it just locks the mutex and then recursively calls stuff while it's locked. Stupid.

Fix this by manually implementing `fmt::Debug`, calling `get` rather than directly accessing `inner`.

Fixes #224.

### DIFF
--- a/src/types/mod.rs
+++ b/src/types/mod.rs
@@ -142,13 +142,19 @@ impl std::error::Error for Error {}
 
 mod bound_mutex {
     use super::{Bound, CompleteBound, Error, Final};
+    use std::fmt;
     use std::sync::{Arc, Mutex};
 
     /// Source or target type of a Simplicity expression
-    #[derive(Debug)]
     pub struct BoundMutex {
         /// The type's status according to the union-bound algorithm.
         inner: Mutex<Arc<Bound>>,
+    }
+
+    impl fmt::Debug for BoundMutex {
+        fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+            self.get().fmt(f)
+        }
     }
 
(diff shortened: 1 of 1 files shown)


uncomputable (approved):
ACK 24fb77a559b3c493a80e77ad69857058df5e027e Good catch.

PR #227, opened 2024-06-30
Some type inference cleanups

apoelstra:
Over the weekend I largely rewrote the type inference API to allocate bounds using a slab allocator so that we don't leak memory when we construct reference cycles. These changes seemed independently useful and could be PR'd on their own.


src/types/final_data.rs
```diff
@@ -227,12 +226,6 @@ impl Final {
     }
 }
 
-impl From<Arc<Final>> for Type {
```
uncomputable, line 230: 12a188570a1cdd5adf73cef73512da58aa9cec2f: Out of curiosity, what is the nonobvious behavior of `From` that you referenced in the commit message?

uncomputable (approved):
ACK dc83206e8916df3ae36dd4b466afcba968b4af65

src/types/final_data.rs
apoelstra: Actually this one is pretty straightforward -- it is the `From` impls from `Bound` to `Type` that are weird. It takes a bound, creates a new union-bound object (which users do not see at all, but which have crazy nonlocal behavior) and constructs a type which initially has that bound and no others.

In a later PR I may actually make the `Bound` type private and not user-visible. It's not actually used outside of the types module (including types/arrow.rs which actually implements the type inference rules) and it's not obvious what it's suppose to mean for users.

PR #228, opened 2024-06-30
[mega pr, do not merge] types: rewrite inference API to use a slab allocator for type bounds

### LINKED ISSUE #226: Failed type inference should not pollute type bounds
Consider the following unit test

```rust
    #[test]
    fn inference_failure_type_pollution() {
        let unit1 = Arc::<ConstructNode<Core>>::unit(); // A -> 1a
        let unit2 = Arc::<ConstructNode<Core>>::unit(); // B -> 1b
        // Constructing a disconnect operator will bind A to 2^256 x B (which
        // will succeed) and then bind 1a to B x 1b (which will fail), so the
        // construction of the disconnect operator should fail.
        Arc::<ConstructNode<Core>>::disconnect(&unit1, &Some(Arc::clone(&unit2))).unwrap_err();
        // But because the construction of the disconnect node failed, it shouldn't
        // pollute the unit; it SHOULD NOT leave A with a product bound.
        assert_eq!(
            unit1.arrow().source.finalize().unwrap(),
            types::Final::unit(),
        );  
    }   
```

This fails currently, because type inference works by applying bounds one after the other. In the case of combinators like `disconnect`, it may happen that one bound succeeds while the following one fails, leaving the types of the program in an inconsistent state.

This isn't a super high priority because under "normal" usage once a type inference step fails you should throw the whole program away and restart, and because I have a set of huge PRs coming in that'll reorganize the type checker to no longer leak memory, but it is causing my fuzzer to slow down.

apoelstra:
Our existing type inference engine assumes a "global" set of type bounds, which has two bad consequences: one is that if you are constructing multiple programs, there is no way to "firewall" their type bounds so that you cannot accidentally combine type variables from one program with type variables from another. You just need to be careful. The other consequence is that if you construct infinitely sized types, which are represented as a reference cycle, the existing inference engine will leak memory.

A third issue, though I am unsure whether it leads to actual bugs, is that the type inference engine can only lock data at the granularity of individual type bounds, meaning that multi-step parts of type inference are not done atomically, and if you try to construct a program in a multithreaded context, it is possible that strange/unexpected things will happen.

To avoid this, it is better that each program under construction has an independent "type inference context" which holds a slab allocator. When the context is dropped, which happens when the initial constructed context and all the program nodes that reference it are dropped, it deallocates all the type bounds in one shot, rather than chasing Arcs and failing to notice cycles.

This is a 2000-line diff but the vast majority of the changes are "API-only" stuff where I was moving types around and threading new parameters through dozens or hundreds of call sites. I did my best to break everything up into commits such that the big-diff commits don't do much of anything and the real changes happen in the small-diff ones to make review easier.

Builds on #227.

This does **not** fix #226 although it sets the stage for "checkpointing" the type inference context, which should allow a reasonably-efficient way to undo changes when things go wrong.

### DIFF
--- a/jets-bench/benches/elements/data_structures.rs
+++ b/jets-bench/benches/elements/data_structures.rs
@@ -4,9 +4,8 @@
 use bitcoin::secp256k1;
 use elements::Txid;
 use rand::{thread_rng, RngCore};
-pub use simplicity::hashes::sha256;
 use simplicity::{
-    bitcoin, elements, hashes::Hash, hex::FromHex, types::Type, BitIter, Error, Value,
+    bitcoin, elements, hashes::Hash, hex::FromHex, types::{self, Type}, BitIter, Error, Value,
 };
 use std::sync::Arc;
 
@@ -57,7 +56,8 @@ pub fn var_len_buf_from_slice(v: &[u8], mut n: usize) -> Result<Arc<Value>, Erro
     assert!(n < 16);
     assert!(v.len() < (1 << (n + 1)));
     let mut iter = BitIter::new(v.iter().copied());
-    let types = Type::powers_of_two(n); // size n + 1
+    let ctx = types::Context::new();
+    let types = Type::powers_of_two(&ctx, n); // size n + 1
     let mut res = None;
     while n > 0 {
         let v = if v.len() >= (1 << (n + 1)) {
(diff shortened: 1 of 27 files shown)


apoelstra:
I will probably break this into more than 2 PRs, but I'll leave this open for now as the "source of truth" for the whole project.

Issue #229, opened 2024-07-01
Drop on `Node` should not be recursive


apoelstra:
We've done a great job eliminating recursion throughout the library, aside from in the typechecker. One subtle place is in the default `Drop` impl on `Node`. This should be manually implemented as something that uses a post-order iterator to replace each node with `Inner::Unit`, iteratively dropping the original.

I took a crack at this and couldn't figure it out after ten minutes but it's definitely possible.

uncomputable:
Does that mean that dropping a deeply nested node can result in a stack overflow, as of now?

apoelstra:
Yep. It's not too hard to write a test case -- you can construct a `take(take(take(...(unit)...)))` iteratively then do nothing with it but drop, and it'll blow out the stack inside the drop glue for `Arc`.

uncomputable:
Ok, wow. I thought that `Drop` cannot stack-overflow because it is implemented in an iterative manner. Then I thought about that algorithm and I realized that it cannot be done in valid Rust. Every type has a distinct implementation of `Drop` so the generic algorithm that works for all types has to be recursive, it seems. We can avoid recursion by manually implementing `Drop` for a specific type. Is that correct?

apoelstra:
Yes. To avoid recursion you need to manually implement `Drop` in a way that you iterative drop its children leaving the original object "disabled" in some way (e.g. replacing it with a unit node using mem::replace) that the recursive drop logic will then be trivial.

PR #231, opened 2024-07-03
types: use slab allocator for type bounds

apoelstra:
Our existing type inference engine assumes a "global" set of type bounds, which has two bad consequences: one is that if you are constructing multiple programs, there is no way to "firewall" their type bounds so that you cannot accidentally combine type variables from one program with type variables from another. You just need to be careful. The other consequence is that if you construct infinitely sized types, which are represented as a reference cycle, the existing inference engine will leak memory.

To fix this, we need to stop allocating type bounds using untethered `Arc`s and instead use a slab allocator, which allows all bounds to be dropped at once, regardless of their circularity. This should also improve memory locality and our speed, as well as reducing the total amount of locking and potential mutex contention if type inference is done in a multithreaded context.

This is a 2000-line diff but the vast majority of the changes are "API-only" stuff where I was moving types around and threading new parameters through dozens or hundreds of call sites. I did my best to break everything up into commits such that the big-diff commits don't do much of anything and the real changes happen in the small-diff ones to make review easier.

By itself, this PR does **not** fix the issue of reference cycles, because it includes an `Arc<Context>` inside the recursive `Type` type itself. Future PRs will:

* Take a single mutex lock during calls to the top-level `bind` and `unify` calls, so that these all happen atomically, including all recursive calls.
* Add another intermediate type under `Type` which eliminates the `Arc<Context>` and its potential for circular references. Along the way, make the `Bound` type private, which is not really used outside of the types module anyway.
* Do "checkpointing" during type inference that makes node construction atomic; this is #226 which is **not** fixed by this PR.
* (Maybe) move node allocation into the type inference context so that nodes can be slab-allocated as well, which will address #229 "for free" without us figuring out a non-recursive `Drop` impl for `Arc<Node<N>>`.


src/types/arrow.rs
```diff
@@ -18,19 +18,30 @@ use crate::node::{
...
+    pub inference_context: Context,
+}
+
+// Having `Clone` makes it easier to derive Clone on structures
+// that contain Arrow, even though it is potentially confusing
+// to use `.clone` to mean a shallow clone.
+impl Clone for Arrow {
+    fn clone(&self) -> Self {
+        self.shallow_clone()
+    }
```
uncomputable, line 44: 8eeab8f2e6dc61eba299e57b282ecf677a9bce0a: Would it make sense to introduce a `ShallowClone` trait with `Clone` as a subtrait? `ShallowClone` would be almost like a marker trait and it would provide the method `shallow_clone` which may call `clone` by default. If people see that both `ShallowClone` and `Clone` is implemented for a type, then they can infer that the type is cheaply cloneable

src/types/context.rs
```diff
@@ -123,11 +125,37 @@ impl Context {
...
+
+    /// Reassigns a bound to a different bound.
+    ///
+    /// # Panics
+    ///
+    /// Panics if called on a complete type. This is a sanity-check to avoid
+    /// replacing already-completed types, which can cause inefficiencies in
+    /// the union-bound algorithm (and if our replacement changes the type,
+    /// this is probably a bug.
+    /// probably a bug.
```
uncomputable, line 150: 70ba9d2dc8bb3550e77c0972aa623ea8d42b530c: typo (repeating line)

src/types/context.rs
```diff
@@ -243,7 +245,7 @@ pub struct OccursCheckId {
     context: *const Mutex<Vec<Bound>>,
     // Will become an index into the context in a latter commit, but for
     // now we set it to an Arc<BoundMutex> to preserve semantics.
```
uncomputable, line 247: 3275c5b92c75d5a757fcad2095a9a16d435e5c58: Nit: This comment is made obsolete

uncomputable (approved):
ACK 3275c5b92c75d5a757fcad2095a9a16d435e5c58 This PR was easier to review than expected. One change flowed into the other and most of the changes were "API only".

src/types/arrow.rs
apoelstra: Promoted to https://github.com/BlockstreamResearch/rust-simplicity/issues/232. I like this idea but it's separate from this PR really.

cc https://smallcultfollowing.com/babysteps/blog/2024/06/21/claim-auto-and-otherwise/

apoelstra:
>Is that correct?

Correct on both counts.

apoelstra:
Fixed both nits. Will leave the `ShallowClone` stuff to a later PR.

uncomputable (approved):
ACK c22dbaa600c74f220822bbeca519f898ebe253ca

PR #233, opened 2024-07-04
types: introduce `TypeInner` making `Type` the only public "incomplete Simplicity type" type

apoelstra:
Previously we had two types, `Type` and `Bound`, which were mutually recursive. `Type` was also exposed in the public API as a type representing an incomplete Simplicity type. `Bound` was exposed in the public API for a couple of reasons which this PR eliminates.
    
However, because `Type` is both public and unusable without a live type inference context, this means that it needs to contain an `Arc<Context>`. But this means that if we have a self-referential type, this leaks to a somewhat-convoluted reference cycle in which a `Type` holds an `Arc<Context>` which owns a set of `Bound`s which contain the original `Type`. If you are skeptical of this, try running the test harness in valgrind before and after the last commit of this PR -- you will see that previously we were leaking memory in the `memory_leak` unit test but now we are not.

This completes the "type inference overhaul" needed to eliminate the memory leak in infinitely-sized types.

### DIFF
--- a/src/types/context.rs
+++ b/src/types/context.rs
@@ -158,14 +158,40 @@ impl Context {
         lock.reassign_non_complete(bound, new);
     }
 
-    /// Binds the type to a given bound. If this fails, attach the provided
-    /// hint to the error.
+    /// Binds the type to a product bound formed by the two inner types. If this
+    /// fails, attach the provided hint to the error.
     ///
     /// Fails if the type has an existing incompatible bound.
-    pub fn bind(&self, existing: &Type, new: Bound, hint: &'static str) -> Result<(), Error> {
+    ///
+    /// # Panics
+    ///
+    /// Panics if any of the three types passed in were allocated from a different
+    /// context than this one.
+    pub fn bind_product(
+        &self,
+        existing: &Type,
+        prod_l: &Type,
+        prod_r: &Type,
+        hint: &'static str,
+    ) -> Result<(), Error> {
+        assert_eq!(
+            existing.ctx, *self,
+            "attempted to bind existing type with wrong context",
+        );
+        assert_eq!(
+            prod_l.ctx, *self,
+            "attempted to bind product whose left type had wrong context",
+        );
+        assert_eq!(
+            prod_r.ctx, *self,
+            "attempted to bind product whose right type had wrong context",
+        );
+
         let existing_root = existing.bound.root();
+        let new_bound = Bound::Product(prod_l.shallow_clone(), prod_r.shallow_clone());
+
         let mut lock = self.lock();
-        lock.bind(existing_root, new).map_err(|e| {
+        lock.bind(existing_root, new_bound).map_err(|e| {
             let new_bound = lock.alloc_bound(e.new);
             Error::Bind {
                 existing_bound: Type::wrap_bound(self, e.existing),
--- a/src/types/context.rs
+++ b/src/types/context.rs
@@ -19,7 +19,7 @@ use std::sync::{Arc, Mutex, MutexGuard};
 
 use crate::dag::{Dag, DagLike};
 
-use super::{Bound, CompleteBound, Error, Final, Type};
+use super::{Bound, CompleteBound, Error, Final, Type, TypeInner};
 
 /// Type inference context, or handle to a context.
 ///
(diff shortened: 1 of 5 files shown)


src/types/context.rs
```diff
@@ -267,6 +277,16 @@ struct LockedContext<'ctx> {
 }
 
 impl<'ctx> LockedContext<'ctx> {
+    fn alloc_bound(&mut self, ctx_ptr: *const Mutex<Vec<Bound>>, bound: Bound) -> BoundRef {
```
uncomputable, line 315: f2c7522df9ff50b5a262b6348ae3847326d36e13: How do we ensure that `ctx_ptr` points to the same memory as `self`?

This is a private method, so we can ensure that is being called correctly, but maybe it would be worth to add `*const Mutex<Vec<Bound>>` as a member to `LockedContext`. This would give `LockedContext` a consistent view of its context and we wouldn't have to pass `ctx_ptr` in `alloc_bound`.

src/types/context.rs
```diff
@@ -171,10 +171,13 @@ impl Context {
     pub fn bind(&self, existing: &Type, new: Bound, hint: &'static str) -> Result<(), Error> {
         let existing_root = existing.bound.root();
         let mut lock = self.lock();
-        lock.bind(existing_root, new).map_err(|e| Error::Bind {
-            existing_bound: e.existing,
-            new_bound: e.new,
-            hint,
+        lock.bind(existing_root, new).map_err(|e| {
+            let new_bound = lock.alloc_bound(Arc::as_ptr(&self.slab), e.new);
```
uncomputable, line 202: 7e7756ebcdde18b28ed0a2193380b8d7cc065e5c: `new_bound` needs to be allocated because the bound doesn't exist in the slab yet. `Type` consists of a `BoundRef` that lives inside the slab, so we have to allocate. Is that correct?

src/types/mod.rs
```diff
@@ -94,8 +95,8 @@ pub use final_data::{CompleteBound, Final};
 pub enum Error {
     /// An attempt to bind a type conflicted with an existing bound on the type
     Bind {
-        existing_bound: Bound,
-        new_bound: Bound,
+        existing_bound: Type,
+        new_bound: Type,
```
uncomputable, line 99: 7e7756ebcdde18b28ed0a2193380b8d7cc065e5c: I think this change is acceptable. `Type` and `Bound` are conceptually similar.

src/types/context.rs
```diff
@@ -164,14 +164,26 @@ impl Context {
...
+    pub fn bind_product(
+        &self,
+        existing: &Type,
+        prod_l: &Type,
+        prod_r: &Type,
+        hint: &'static str,
+    ) -> Result<(), Error> {
+        assert_eq!(existing.ctx, *self);
+        assert_eq!(prod_l.ctx, *self);
+        assert_eq!(prod_r.ctx, *self);
```
uncomputable, line 195: d42c496c15925caab8cb55b873b5eb744a9d5598: We should add a readable error message and update the function description with a `# Panics` section.

uncomputable (commented):
Reviewed 15d69b21c55577e538ccfffac043c46df431194b Looks great. I left some comments.

src/types/context.rs
apoelstra: Yep exactly.

src/types/context.rs
apoelstra: Sure. I think I had this in an earlier iteration.

src/types/mod.rs
```diff
@@ -200,12 +181,25 @@ impl Bound {
 /// first one.
 #[derive(Clone)]
 pub struct Type {
+    /// Handle to the type context.
     ctx: Context,
+    /// The actual contents of the type.
+    inner: TypeInner,
+}
```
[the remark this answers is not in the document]
apoelstra: Yep.

src/types/context.rs
apoelstra: Addressed.

apoelstra:
Addressed the two actionable comments. Should be good to go.

uncomputable (approved):
ACK dada0f29f9ffcf34cb21c7573726bd524278e8bf

PR #234, opened 2024-07-11
update staging branch

apoelstra:
Draft til #233 goes in, and maybe will split into multiple PRs.


apoelstra:
This should be good to go. I have a followup that updates the benchmarks but it's a bit of a mess right now so this just updates the vendored version.

Will do another followup with the latest changes from upstream; but this is the hard stuff of splitting witness decoding from program decoding so it's still worth reviewing and merging as-is.

apoelstra:
Fixed fuzztest. The windows CI failures are weirder.

apoelstra:
```

254
  INTERNAL COMPILER ERROR in 'C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.40.33807\bin\HostX64\x64\cl.exe'
255
      Please choose the Technical Support command on the Visual C++
256
      Help menu, or open the Technical Support help file for more information
```

Nice.

fuzz/fuzz_targets/c_rust_merkle.rs
```diff
@@ -8,9 +8,22 @@ use simplicity::jet::Elements;
...
 
 fn do_test(data: &[u8]) {
-    let c_result = run_program(data, TestUpTo::CheckOneOne);
+    if data.len() < 4 {
+        return;
+    }
+    let prog_len = (usize::from(data[0]) << 24)
+        + (usize::from(data[1]) << 16)
+        + (usize::from(data[2]) << 8)
+        + usize::from(data[3]);
```
uncomputable, line 17: 0b5c69f638177d5a0dd3916b0681d18f2e72fddd: Does this decode the program length from the Simplicity encoding? If so, let's mention this in a comment.

uncomputable (commented):
ac0cb0c2c37bf63b407ea85835eab52f12ac2973 Verified the vendored C files. The Rust code looks good besides a nit.

The GitHub fuzzer seems to have found a crash.

_The fuzzing doesn't work in my local nix setup, probably because my nix shells are not as reproducible as I thought :/_

fuzz/fuzz_targets/c_rust_merkle.rs
```diff
@@ -78,7 +79,7 @@ mod tests {
     #[test]
     fn duplicate_crash() {
         let mut a = Vec::new();
-        extend_vec_from_hex("d73804000000000000000000000002040404040000000000000040000001ff0000000800380000000000000002040404040000000000000040000001ff0000000800380000000000000000000000000000385cf8ffffff10cdf7bb2cd063000000000000042d343507fffffffffff800404220fffffff800000040420f0000", &mut a);
+        extend_vec_from_hex("00000004ffff", &mut a);
```
uncomputable, line 82: ac0cb0c2c37bf63b407ea85835eab52f12ac2973

> `=====================================================================
> 0000003238000085000000000010ff07fffffffffffffffff0ffc47758e802cb5de0f1e38c07ee3e5975ebf9043433040404040401c71c71c71c71c80404040494061101000000c7711cc70000000000000004000001000100000000000000040409000000000000001f000000000000`

> C accepted code that Rust rejected: bitstream had trailing bytes 0x00...

apoelstra: Ah, it's my responsibility to call closeBitstream in my C-based test harness.

fuzz/fuzz_targets/c_rust_merkle.rs
apoelstra: Sure, I'll add a comment. Actually I think I can improve it.

apoelstra:
Fixed the fuzztest.

fuzz/fuzz_targets/c_rust_merkle.rs
```diff
@@ -8,23 +8,29 @@ use simplicity::jet::Elements;
...
+    let prog_len = {
+        let mut iter = BitIter::from(data);
+        match simplicity::decode::decode_expression::<_, Elements>(&mut iter) {
+            Ok(_) => (iter.n_total_read() + 7) / 8,
+            Err(_) => match data.first() {
+                Some(&n) => core::cmp::min(data.len(), n.into()),
+                None => return,
+            },
+        }
+    };
```
uncomputable, line 26: 8d3eda887b7adaf8a890701e3625db0ac19b6672: Do we compute the length of the program from the fuzzer bytes to help the fuzzer, because most random byte sequences won't have a valid length prefix?

uncomputable (approved):
ACK 8d3eda887b7adaf8a890701e3625db0ac19b6672

uncomputable:
I hope we can fix the Windows build with an upstream update to https://github.com/BlockstreamResearch/simplicity

PR #237, opened 2024-07-24
Value refactor

uncomputable:
Rename the variants of `Value`  and construct {256, 512}-bit values safely.

### DIFF
--- a/jets-bench/src/data_structures.rs
+++ b/jets-bench/src/data_structures.rs
@@ -67,12 +67,12 @@ pub fn var_len_buf_from_slice(v: &[u8], mut n: usize) -> Result<Arc<Value>, Erro
         let v = if v.len() >= (1 << (n + 1)) {
             let ty = &types[n];
             let val = iter.read_value(&ty.final_data().unwrap())?;
-            Value::sum_r(val)
+            Value::right(val)
         } else {
-            Value::sum_l(Value::unit())
+            Value::left(Value::unit())
         };
         res = match res {
-            Some(prod) => Some(Value::prod(prod, v)),
+            Some(prod) => Some(Value::product(prod, v)),
             None => Some(v),
         };
         n -= 1;
@@ -170,7 +170,7 @@ impl SimplicityEncode for SimplicityCtx8 {
             .flat_map(|x| x.to_be_bytes())
             .collect::<Vec<u8>>();
         let mid_state = Value::u256_from_slice(&arr);
-        Value::prod(buf, Value::prod(len, mid_state))
+        Value::product(buf, Value::product(len, mid_state))
     }
 }
 
@@ -178,7 +178,7 @@ impl SimplicityEncode for elements::OutPoint {
     fn value(&self) -> Arc<Value> {
         let txid = Value::u256_from_slice(&self.txid[..]);
         let vout = Value::u32(self.vout);
-        Value::prod(txid, vout)
+        Value::product(txid, vout)
     }
 }
 
@@ -186,13 +186,13 @@ impl SimplicityEncode for elements::confidential::Asset {
     fn value(&self) -> Arc<Value> {
         match self {
             elements::confidential::Asset::Explicit(a) => {
-                Value::sum_r(Value::u256_from_slice(&a.into_inner()[..]))
+                Value::right(Value::u256_from_slice(&a.into_inner()[..]))
             }
             elements::confidential::Asset::Confidential(gen) => {
                 let odd_gen = gen.serialize()[0] & 1 == 1;
                 let x_pt = Value::u256_from_slice(&gen.serialize()[1..]);
                 let y_pt = Value::u1(odd_gen as u8);
-                Value::sum_l(Value::prod(y_pt, x_pt))
+                Value::left(Value::product(y_pt, x_pt))
             }
             elements::confidential::Asset::Null => panic!("Tried to encode Null asset"),
         }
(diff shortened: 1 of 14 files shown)


uncomputable:
Fixed the jets-bench subcrate.

src/value.rs
```diff
@@ -14,36 +14,37 @@ use std::fmt;
...
 pub enum Value {
-    /// Unit value
+    /// Unit value.
+    ///
+    /// The unit value is the only value of the unit type `1`.
+    /// Because there is only one value to choose from,
+    /// this means that the unit value doesn't contain any information.
+    ///
+    /// The unit value is wrapped in left and right values
+    /// to encode information.
```
apoelstra, line 26: In 7459af049bca4c548626a56e350dc074755d5725:

"The unit value must be embedded in a sum type to encode information."

src/value.rs
```diff
@@ -14,36 +14,37 @@ use std::fmt;
...
     Unit,
-    /// Sum value that wraps a left value
+    /// Left value that wraps an inner value.
+    ///
+    /// A left value is the first option for a sum type `A + B`.
+    /// It means we go left and wrap a value of type `A`.
+    /// The choice of going left encodes one bit.
```
apoelstra, line 32: In 7459af049bca4c548626a56e350dc074755d5725:

I find this "go left" terminology confusing. I think it's clearer to just say "Encodes a value of type `A` as a sum type `A + B`, using a bit to specify that the left type is occupied." Similar with the below.

src/value.rs
```diff
@@ -14,36 +14,37 @@ use std::fmt;
...
+    /// The choice of going right encodes one bit.
     Right(Arc<Value>),
-    /// Product value that wraps a left and a right value
+    /// Product value that wraps a left and a right value.
+    ///
+    /// A product value is the only option for a product type `A × B`.
+    /// It means we go both left and right and
+    /// that we wrap values of both type `A` and type `B`.
+    /// There is no choice,
+    /// so the information is contained solely in the wrapped values.
```
apoelstra, line 46: In 7459af049bca4c548626a56e350dc074755d5725:

This "only option" business is definitely confusing. The number of values in a product type is equal to the product of the numbers of values of the factor types. I would drop this paragraph and say "Since there is no ambiguity about which types are occupied, no additional bits are used and the information is contained solely in the wrapped values.". If you want to say this at all.

jets-bench/benches/elements/data_structures.rs
```diff
@@ -160,19 +160,21 @@ impl SimplicityEncode for SimplicityCtx8 {
...
+        let arr: [u8; 32] = self
             .h
             .iter()
             .flat_map(|x| x.to_be_bytes())
-            .collect::<Vec<u8>>();
-        let mid_state = Value::u256_from_slice(&arr);
+            .collect::<Vec<u8>>()
+            .try_into()
+            .unwrap();
+        let mid_state = Value::u256(&arr);
```
apoelstra, line 170: In 010e675ecf7e7990d3699b5f4fbe4c0c72302a9b:

I guess it doesn't matter because we intend to drop this code anyway after #239 but it would be nice to get rid of the unnecessary `Vec` here.

apoelstra:
utACK 1a818692c63e5b0a447e389511e559ec94d4708b. Let me know if you want to address the above comments, and I can ack/merge.

src/value.rs
uncomputable: I like the word "wrap" because we wrap values in Rust all the time. Values cannot be embedded inside types.

uncomputable:
Rebased on master after #239 got merged

jets-bench/benches/elements/data_structures.rs
uncomputable: I switched to using `Value::power_of_two`, which is also internally called by `Value::u256`. We still have to allocate a `Vec`, like on master, but we no longer have to create an extra byte array.

jets-bench/src/data_structures.rs
```diff
@@ -169,14 +169,15 @@ impl SimplicityEncode for SimplicityCtx8 {
             .iter()
             .flat_map(|x| x.to_be_bytes())
             .collect::<Vec<u8>>();
-        let mid_state = Value::u256_from_slice(&arr);
+        debug_assert_eq!(arr.len(), 32);
+        let mid_state = Value::power_of_two(&arr);
```
apoelstra, line 173: In c1527fa7668e0cc423ce5327a222c2cbca1e2485:

I'm confused as to what you're doing here. In the previous version of this commit you used `Value::u256` because this is guaranteed to be 32 bytes. But now you are just adding another runtime assertion. And this still doesn't touch the unnecessary `Vec` in the line above.

uncomputable: I can convert `[u32; 8]` directly into `[u8; 32]` if you want. The code will be more verbose because I will have to initialize the array and then assign values byte by byte. I cannot use Rust's iterator API, AFAIK.

I misunderstood your original comment. I thought you called the `Vec` unnecessary because I converted it into an `[u8; 32]` array on the stack. I removed the array, but you seem to have wanted to remove the `Vec` altogether.

uncomputable:
Removed the unnecessary vector.

apoelstra:
Ok, awesome. ACK 85fd5b37122d8fff5feae29a6651d5a4aa41805b except that I'd prefer that these new functions took arrays by value, not by reference. It is almost certainly slower, and more annoying for the programmer, to be passing 32-byte objects around by reference.

apoelstra:
I did not actually test this or use the github "approve" button. I did not intend my comment to be read as an ACK.

uncomputable:
The [github-merge](https://github.com/bitcoin-core/bitcoin-maintainer-tools/blob/main/github-merge.py) from the bitcoin maintainer tools interpreted the comment as an ACK. Sorry for that. I will open a fixup PR to pass the 32 / 64 bytes by value.

apoelstra:
No worries -- I shouldn't have written ACK in all caps so that the script wouldn't pick it up. And because this is such a minor thing I should've been clearer that I wanted you to respond to it before merging.

PR #239, opened 2024-07-25
benchmarks: updates and cleanups

apoelstra:
This significantly improves and cleans up the benchmarks. This branch was used to generate the data for https://github.com/BlockstreamResearch/simplicity/pull/247.

The main changes are:

* Refactors transaction environment sampling to be more extensible; extends it with the ability to create lots of fee outputs.
* Replaces the recursive `Value` logic for input/output buffers with a new custom `FlatValue` type which uses flat arrays and should be much cleaner.
* Writes new faster/correct samplers for the `FlatValue` type which can call into libsimplicity to do things like computing valid ges (without doing a pubkey multiplication) and computing valid gejs (which previously was just impossible)
* Rewrites the sampling traits to support multiple distributions; runs the benchmarks with different distributions to find the best/worst ones rather than repeatedly benchmarking with adhoc data.

For the transaction environment jets, which include all the transaction intropection jets, this makes minimal changes. A future PR should do something similar for the transaction environment, running multiple benchmarks across multiple distributions. But because these jets are implemented as lookups into a compute-once-per-tx transaction environment structure, I expect their cost to be very low and low-variance, so I did not do this work.


jets-bench/src/check_all_jets.rs
```diff
@@ -0,0 +1,55 @@
...
+
+static CHECKED_JETS: Mutex<Option<BTreeMap<Jet, usize>>> = Mutex::new(None);
+
+pub fn initialize() {
+    // This is insanely inefficient but will find all jets with encodings <= 24 bits
+    // and empirically it takes less than a second to run against the multi-hour
+    // benchmark run, so whatever.
+    //
+    // In our generated Rust code we should add an integer<>jet map, a count of all
+    // jets, an iterator over all jets, etc., but for now we will make do.
```
uncomputable, line 18: 73baa0fc3ffdfe635660ab4b57eeb43951b10404: I'm on it

jets-bench/src/input.rs
```diff
@@ -93,3 +441,934 @@ impl InputSampling {
...
+}
+
+pub struct EqProduct<L>(pub L);
+
+impl<L: InputSample> InputSample for EqProduct<L> {
+    fn n_distributions(&self) -> usize {
+        // both random
+        self.0.n_distributions() * self.0.n_distributions()
+            // both equal, and each a bit-inverse of the other
+            + 2 * self.0.n_distributions()
```
uncomputable, line 705: 1eb9db38d89add11c740e93ec15a21793ce34945
> `n + n + 2 * n`

I'm just skimming this code, but I expected `2n` instead of `4n`.

uncomputable (approved):
ACK 5ee4bf50cdcbf9e490f91b9ce7536189cec6de47 Love the `InputSample` trait and the combination of multiple samples into a bigger sample.

jets-bench/src/input.rs
apoelstra: I don't know what to make of this comment. `GenericProduct` does `n*m` as you would expect and `EqProduct` can do other distributions because it knows both factors have the same size, so it won't behave the same way as `EqProduct<L>`.

I don't see `4n` anywhere in the codebase.

uncomputable: For `EqProduct<L>` I read `self.0.n_distributions() * self.0.n_distributions() + 2 * self.0.n_distributions()` as `4 * self.0.n_distributions()`. For `GenericProduct<L, L>` (with the same generic parameter twice) I read `self.0.n_distributions() * self.1.n_distributions()` as `2 * self.0.n_distributions()`. This is of course wrong, but I hope you see where I was coming from.

PR #240, opened 2024-07-28
Value fixup

uncomputable:
Fixes outstanding issues of #237 which I failed to address before merging.

### DIFF
--- a/jets-bench/src/data_structures.rs
+++ b/jets-bench/src/data_structures.rs
@@ -168,14 +168,14 @@ impl SimplicityEncode for SimplicityCtx8 {
         for (i, byte) in self.h.iter().flat_map(|x| x.to_be_bytes()).enumerate() {
             arr[i] = byte;
         }
-        let mid_state = Value::u256(&arr);
+        let mid_state = Value::u256(arr);
         Value::product(buf, Value::product(len, mid_state))
     }
 }
 
 impl SimplicityEncode for elements::OutPoint {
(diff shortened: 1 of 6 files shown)


jets-bench/src/data_structures.rs
```diff
@@ -185,12 +185,13 @@ impl SimplicityEncode for elements::confidential::Asset {
...
             elements::confidential::Asset::Explicit(a) => {
-                Value::right(Value::u256(&a.into_inner().0))
+                Value::right(Value::u256(a.into_inner().to_byte_array()))
             }
             elements::confidential::Asset::Confidential(gen) => {
                 let ser = gen.serialize();
                 let odd_gen = ser[0] & 1 == 1;
-                let x_bytes: &[u8; 32] = (&ser[1..33]).try_into().unwrap();
+                let mut x_bytes = [0u8; 32];
+                x_bytes.copy_from_slice(&ser[1..33]);
```
apoelstra, line 194: In 753d70a38eb65935bdb798872eb9cf6400f4a4d5:

You can definitely keep the old `try_into().unwrap()` logic. All you need to do is change `&[u8; 32]` to `[u8; 32]`.

Repeated many times.

uncomputable: Addressed

apoelstra (approved):
ACK 5c6a26feb1c549eba767b5844d5a41f2e078cc3c nice\!

PR #241, opened 2024-07-31
simplicity-sys: stop referring to non-exported mkJetCMR function

apoelstra:
As seen when running

    RUSTFLAGS=-Clink-dead-code cargo test --all-features

we are referring to a non-exported function mkJetCMR from the C code. By default rustc prunes this function and never tries to link it, but if you add the -Clink-dead-code rustflag (which may be needed when compiling to link with an external fuzzer) then linking fails.


uncomputable (approved):
ACK 78e14de7e2b7ed25d522f7f04b74225007b23ad0

PR #242, opened 2024-08-08
update libsimplicity to 121b2951d8cca3d73dccd6fcbe721f2195025067

apoelstra:
This update renames a bunch of symbols from x to simplicity_x. This commit has a small number of non-mechanical changes to address that.

The one "judgement call" was to update wrapper.h in accordance with https://github.com/BlockstreamResearch/simplicity/pull/248#issuecomment-2248830161 to cover all the jet wrapper renames.


apoelstra (commented):
Successfully ran local tests on c9e70ed91851850f96f8abc1035b6a25c067df25.

PR #243, opened 2024-08-24
Update libsim to 3daad68dc9e995fefcd5287fa5b15a7bfb13462a

uncomputable:
This update adds an array of all Core / Bitcoin / Elements jets, which is useful for fuzzing.


apoelstra (approved):
ACK d5c3d1c14539486b89d765aee78b50acd4f1b319 successfully ran local tests

PR #244, opened 2024-08-29
Typed  value

uncomputable:
We need to encode values with padding on the Bit Machine. This encoding requires information about the type of the value. It turns out that universally typing all Simplicity values produces the smallest diff, and it is the safest API. This PR refactors the `Value` struct to include type information.

I also have ideas for a `Word` struct that wraps values of the word type. Only these values are safe to use in word jets. I will leave this for a follow-up PR.


uncomputable:
Rebased on latest master

apoelstra (approved):
ACK 5a5a4e24f23bb86a200a9b81d28a4d36d14997ed successfully ran local tests; nice! And does the API changes needed to later replace the `Value` internal representation with a flat bit array, which should greatly speed things up

PR #249, opened 2024-09-27
Web IDE fixes

uncomputable:
Some fixes that will make my life easier in the web IDE. Bump of MSRV to 1.63.0.


src/node/redeem.rs
```diff
@@ -299,7 +299,9 @@ impl<J: Jet> RedeemNode<J> {
             ) -> Result<Value, Self::Error> {
                 let arrow = data.node.data.arrow();
                 let target_ty = arrow.target.finalize()?;
-                self.bits.read_value(&target_ty).map_err(Error::from)
+                self.bits
+                    .decode_padded_value(&target_ty)
```
apoelstra, line 303: In 0719071b8f119db5345a62863a8a96ddf8cc594a:

The redeemnode encoding definitely does not use padded witness values. Does this really not break any tests?

uncomputable: You are right. The witness values are encoded without padding. It's amazing that our unit tests don't fail.

I will introduce a value decoder with padding (for frames, etc.) and a decoder without padding (for witness data, etc.).

uncomputable: The new methods force us to think about the presence / absence of padding, which is good.

apoelstra: Thanks! And yes, absolutely. My original code had made the opposite mistake (assuming everything was using unpadded encoding).

uncomputable:
I moved the value decoding logic into associated methods of the `Value` struct. There are separate methods for compact and for padded encoding. A private trait is used for code de-duplication, which should be a zero-cost abstraction.

I also fixed some clippy lints.

src/value.rs
```diff
@@ -388,6 +389,138 @@ impl<'a> Iterator for PaddedBitsIter<'a> {
...
+
+    fn read_right_padding<I: Iterator<Item = bool>>(
+        bits: &mut I,
+        ty_l: &Final,
+        ty_r: &Final,
+    ) -> Result<(), EarlyEndOfStreamError>;
+}
+
+struct CompactEncoding;
+struct PaddedEncoding;
```
apoelstra, line 407: In 34e1243f36022f88376fcaf810145d539a6aa669:

These should be empty enums. Otherwise they will have constructors.

uncomputable: Thanks for pointing this out. I pushed a fix.

apoelstra (approved):
ACK 9ef06ef0280265432b7e2b1fd82492dd041c0aef successfully ran local tests

PR #250, opened 2024-09-28
Fix value decoding from padded encoding

uncomputable:
Fix a typo that was introduced in #249


apoelstra (approved):
ACK de99f96a21624db3d662b3f2faa5f138cc13baa6 successfully ran local tests; oops

PR #251, opened 2024-10-07
Remove serde dependency

uncomputable:
Remove the `serde_json` dependency when the `serde` feature is disabled. Update libsimplicity.


Cargo.toml
```diff
@@ -14,7 +14,7 @@ rust-version = "1.63.0"
 default = ["elements"]
 elements = ["dep:elements", "bitcoin"]
 test-utils = ["simplicity-sys/test-utils"]
-serde = ["dep:serde", "bitcoin/serde", "elements/serde"]
+serde = ["dep:serde", "bitcoin/serde", "elements/serde", "elements/json-contract"]
```
apoelstra, line 17: In 2a2e7fb84d29f6f315d0bc1db1c6eb584c75f5ee:

Why enable this? Let's just leave it off.

uncomputable: I thought `json-contract` might be useful for some serde stuff, because the feature is enabled by default, but you're right, we probably don't need it.

uncomputable: Fixed

apoelstra:
Looks like the lockfiles need to be updated.

uncomputable:
Lockfiles are updated. This seems to have fixed CI

apoelstra:
I apparently can't test this until we update to ghc 9.8, because my script can't verify that the generated files are correct.

apoelstra:
@uncomputable can you open a new PR which updates libsimplicity?

uncomputable:
I am on it. I was thinking of pushing to this PR.

uncomputable:
I updated the Merkle roots, but for some reason C and Rust still don't agree on them. It is getting late, so I will take this up again tomorrow. Drafting the PR.

apoelstra:
@uncomputable I patched this up locally to create a test transaction. There are two changes you need to make:

* The `Cmr::BITS` array needs to be updated (you can just change them to the value that makes the unit tests pass; and you can check against the C code in precomputed.h if you want to be sure)
* A couple more unit tests have fixed vectors in them, which need to be updated.

uncomputable:
@apoelstra Thanks for pointing to `Cmr::BITS`. I was unsure which hardcoded Merkle root constants were wrong and I missed `Cmr::BITS`. I added a unit test with print statements for future updates.

apoelstra:
utack c85c6f1eee92b885141154e15c14ee7589060be8 except that we should leave the names of the Tmr IVs alone, and keep exporting them as `pub`

uncomputable:
I can add getters to the `Tmr` constants if we need them. The name of the `Tmr` constants changed due to the code that generates them. The new names are consistent with the names of the `Cmr` constants.

apoelstra (approved):
ACK 905bf9b8c20489ff7c921f02447033578f31f2e4 successfully ran local tests

PR #252, opened 2024-10-12
Word jets with word values

### LINKED ISSUE #155: Scribe with word jets
Scribe is pretty useless now that we have word jets. It's basically only useful for values that are not strings of bytes. With that in mind, I propose we redefine scribe to greedily use word jets. That is, if we have a left sum value of some byte string, we use injl on a word jet, and so on. _(We can choose a different name for this operation if it is confusing. But the pure scribe is really only useful for the completeness theorem and for test cases.)_

uncomputable:
Word jets can only host word values. Currently, our code has some assertions here and there that will fail at runtime if a non-word is passed. The assertions are not fully rigorous, especially regarding the maximum word type of `TWO^(2^31)`.

This PR introduces the wrapper type `Word` which is guaranteed to be a valid word. `Word` is used everywhere where a word should be used.

This PR also optimized scribe to use as many word jets as possible. Fixes #155


apoelstra:
Nice! Pretty slick that we can just use TMR comparisons like this to detect when values are words.

Later we may even want to "optimize" the TMR type into an enum where we special-case a few dozen common/special values, so we can compare just a couple bytes rather than 32.

apoelstra (approved):
ACK f4e259a153cb281d1bc75f9ee3044ad1c99fc328 successfully ran local tests

PR #259, opened 2025-01-09
Implement pruning

### LINKED ISSUE #84: Implement pruning
`case` nodes should be replaced with `assertl` or `assertr` if their left or right branches, respectively, are not used during execution. Pruning should happen during finalization.

Pruning is non-trivial in conjunction with sharing, since inferred types may change. Witness targets may shrink and IMRs may change, recursively leading to more sharing.

uncomputable:
Implement full pruning by running programs on the Bit Machine and tracking (un)used branches. The unused branches are then removed from the program and types are adjusted. Witness data is also adjusted by removing unused bits.

Fixes #84

### DIFF
--- a/src/bit_machine/mod.rs
+++ b/src/bit_machine/mod.rs
@@ -8,14 +8,15 @@
 
 mod frame;
 
+use std::collections::HashSet;
 use std::error;
 use std::fmt;
 use std::sync::Arc;
 
-use crate::analysis;
 use crate::jet::{Jet, JetFailed};
 use crate::node::{self, RedeemNode};
 use crate::types::Final;
+use crate::{analysis, Imr};
 use crate::{Cmr, FailEntropy, Value};
 use frame::Frame;
 
@@ -204,13 +205,44 @@ impl BitMachine {
         Ok(())
     }
 
-    /// Execute the given program on the Bit Machine, using the given environment.
+    /// Execute the given `program` on the Bit Machine, using the given environment.
     ///
-    /// Make sure the Bit Machine has enough space by constructing it via [`Self::for_program()`].
-    pub fn exec<J: Jet + std::fmt::Debug>(
+    ///  ## Precondition
+    ///
+    /// The Bit Machine is constructed via [`Self::for_program()`] to ensure enough space.
+    pub fn exec<J: Jet>(
         &mut self,
         program: &RedeemNode<J>,
         env: &J::Environment,
+    ) -> Result<Value, ExecutionError> {
+        self.exec_with_tracker(program, env, &mut NoTracker)
+    }
+
+    /// Execute the given `program` on the Bit Machine and track executed case branches.
+    ///
+    /// If the program runs successfully, then two sets of IMRs are returned:
+    ///
+    /// 1) The IMRs of case nodes whose _left_ branch was executed.
+    /// 2) The IMRs of case nodes whose _right_ branch was executed.
+    ///
+    /// ## Precondition
+    ///
+    /// The Bit Machine is constructed via [`Self::for_program()`] to ensure enough space.
+    pub(crate) fn exec_prune<J: Jet>(
+        &mut self,
+        program: &RedeemNode<J>,
+        env: &J::Environment,
+    ) -> Result<SetTracker, ExecutionError> {
+        let mut tracker = SetTracker::default();
+        self.exec_with_tracker(program, env, &mut tracker)?;
+        Ok(tracker)
+    }
+
+    fn exec_with_tracker<J: Jet, T: CaseTracker>(
+        &mut self,
+        program: &RedeemNode<J>,
+        env: &J::Environment,
+        tracker: &mut T,
     ) -> Result<Value, ExecutionError> {
         enum CallStack<'a, J: Jet> {
             Goto(&'a RedeemNode<J>),
@@ -316,12 +348,14 @@ impl BitMachine {
                             self.fwd(1 + a.pad_right(b));
                             call_stack.push(CallStack::Back(1 + a.pad_right(b)));
                             call_stack.push(CallStack::Goto(right));
+                            tracker.track_right(ip.imr());
                         }
                         (node::Inner::Case(left, _), false)
                         | (node::Inner::AssertL(left, _), false) => {
                             self.fwd(1 + a.pad_left(b));
                             call_stack.push(CallStack::Back(1 + a.pad_left(b)));
                             call_stack.push(CallStack::Goto(left));
+                            tracker.track_left(ip.imr());
                         }
                         (node::Inner::AssertL(_, r_cmr), true) => {
                             return Err(ExecutionError::ReachedPrunedBranch(*r_cmr))
@@ -470,6 +504,60 @@ impl BitMachine {
             Ok(())
         }
     }
+}
+
+/// A type that keeps track of which case branches were executed
+/// during the execution of the Bit Machine.
+///
+/// The trait is implemented for [`SetTracker`], which does the actual tracking,
+/// and it is implemented for [`NoTracker`], which is a dummy tracker that is
+/// optimized out by the compiler.
+///
+/// The trait enables us to turn tracking on or off depending on a generic parameter.
+trait CaseTracker {
+    /// Track the execution of the left branch of the case node with the given `imr`.
+    fn track_left(&mut self, imr: Imr);
+
+    /// Track the execution of the right branch of the case node with the given `imr`.
+    fn track_right(&mut self, imr: Imr);
+}
+
+/// Tracker of executed left and right branches for each case node.
+#[derive(Clone, Debug, Default)]
+pub(crate) struct SetTracker {
+    left: HashSet<Imr>,
+    right: HashSet<Imr>,
+}
+
+impl SetTracker {
+    /// Access the set of IMRs of case nodes whose left branch was executed.
+    pub fn left(&self) -> &HashSet<Imr> {
+        &self.left
+    }
+
+    /// Access the set of IMRs of case nodes whose right branch was executed.
+    pub fn right(&self) -> &HashSet<Imr> {
+        &self.right
+    }
+}
+
+#[derive(Copy, Clone, Debug)]
+struct NoTracker;
+
+impl CaseTracker for SetTracker {
+    fn track_left(&mut self, imr: Imr) {
+        self.left.insert(imr);
+    }
+
+    fn track_right(&mut self, imr: Imr) {
+        self.right.insert(imr);
+    }
+}
+
+impl CaseTracker for NoTracker {
+    fn track_left(&mut self, _: Imr) {}
+
+    fn track_right(&mut self, _: Imr) {}
 }
 
 /// Errors related to simplicity Execution
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -88,8 +88,6 @@ pub enum Error {
     Execution(bit_machine::ExecutionError),
     /// Witness iterator ended early
     NoMoreWitnesses,
-    /// Finalization failed; did not have enough witness data to satisfy program.
-    IncompleteFinalization,
     /// Tried to parse a jet but the name wasn't recognized
     InvalidJetName(String),
     /// Policy error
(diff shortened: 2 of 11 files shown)


src/value.rs
```diff
@@ -300,6 +300,76 @@ impl Value {
             n,
         })
     }
+
+    /// Prune the value down to the given type.
+    ///
+    /// The pruned type must be _smaller than or equal to_ the current type of the value.
+    /// Otherwise, this method returns `None`.
```
apoelstra, line 356: In 6436f6c221c37570cf37dc9557edb373a33019a6:

Actually, there are no checks for compatibility. Probably if you give it a bad type, the method will panic.

I would suggest just saying "the output is undefined and the function is likely to panic" and changing the `Option` return to just return `Self`.

src/value.rs
```diff
@@ -300,6 +300,76 @@ impl Value {
...
+                Task::MakeProduct => {
+                    let r_value = output.pop().unwrap();
+                    let l_value = output.pop().unwrap();
+                    output.push(Value::product(l_value, r_value));
+                }
+            }
+        }
+
+        debug_assert_eq!(output.len(), 1);
+        output.pop()
```
apoelstra, line 420: In 6436f6c221c37570cf37dc9557edb373a33019a6:

...and I would change this `debug_assert_eq` to an `assert_eq`. It's a super cheap check, we're only doing it once, and it'll catch some forms of "the passed type is not smaller than the original type".

src/value.rs
```diff
@@ -291,6 +291,55 @@ impl Value {
...
+    /// Get the default value of the given type.
+    ///
+    /// The default value serializes to a string of zeroes.
+    ///
+    /// ## Construction
+    ///
+    /// - `default( 1 )` = `()`
+    /// - `default( A + B )` = `default(A)`
+    /// - `default( A × B )` = `default(A) × default(B)`
+    pub fn default_of_type(ty: &Final) -> Self {
```
apoelstra, line 303: In 8784adaa9fc42b7b53e076ec7cbb257b04ae8b92:

Can we call this `zero_value` or something like that? I think "default" misleadingly implies that the returned value is important.

src/lib.rs
```diff
@@ -98,19 +100,20 @@ pub enum Error {
 impl fmt::Display for Error {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match self {
-            Error::Decode(ref e) => fmt::Display::fmt(e, f),
+            Error::Decode(ref e) => write!(f, "{e}"),
```
apoelstra, line 101: In 72a190f36c143626572f4679e258cc1d34dbd6fb:

You are changing this to override any format specifications on `f` (e.g. alternate display). This is a mild reduction in functionality (e.g. it prevents pretty-printing these variants) and also isn't mentioned in the commit message. What is the purpose of this?

src/value.rs
uncomputable: The type checking happens as the value is split: There are sub-tasks of the form "prune value V to type T". We match over T.

If T is unit, then the result is simply unit. V is discarded.
If T is a sum type, then V is deconstructed if it is a left or right value. A new subtask is spawned. If  V is neither left or right, then the entire algorithm returns `None`.
If T is a product type, then V is deconstructed if it is a product. Two subtasks are spawned. Otherwise, the algorithm returns `None`.

I think this is somewhat of a proof that the method never panics or enters undefined behavior.

src/value.rs
uncomputable: I think this assertion is guaranteed to be true for all inputs.

src/value.rs
uncomputable: I borrowed Rust's `Default::default` terminology, where the default integer values are also zeroes. However, Rust describes this trait as _"A trait for giving a type a useful default value."_ which implies that each implementation should think about which default value is the most useful. I, on the other hand, always return the value that serializes to zeroes, so I can see that the name "default" can be misleading.

`Value::zero` or `Value::all_zeroes` could be better.

src/lib.rs
uncomputable: I didn't know about this, so thank you for pointing this out. I thought `write!(f, "{e}")` is more readable and is expanded into `fmt::Display::fmt(e, f)`.

src/value.rs
uncomputable: I added unit tests for cases where the pruned type is not smaller than or equal to the value type.

uncomputable:
Rebased and addressed the comments.

src/value.rs
apoelstra: >I think this is somewhat of a proof that the method never panics or enters undefined behavior.

Oh, I see. I missed the `?`s in the algorithm, which are actually error return paths. Ok, never mind, I'm happy.

src/value.rs
apoelstra: Yep, thanks! See above -- I missed the `?`s and though that there were no error returns up to this point.

src/value.rs
apoelstra: I think that using 0 as the "default" is only a convention for numeric types. For containers, the "empty container" is the default value. This is already an overload of two things that mean something quite different.

If stdlib were designed by people who could do math, zero would be an associated constant on `Add` (and one an associated constant on `Mul`), but I was told "this is not Haskell" whenever I suggested that things be sane or consistent. (And also there were no associated constants back then, so zero/one would've needed to be *methods* and that would've been kinda ugly.) So instead we have a `+` operator with no identity which is routinely used for noncommutative operations and you just have to guess what anything means. Might as well be C++ with side-effecting shift operators..

Anyway, for non-numerics and non-containers there is no convention and it's nonobvious to users what the "default value" is supposed to be. So I think we should avoid the trait except in those two places.

src/lib.rs
apoelstra: If you want, you can replace `fmt::Display::fmt(e, f)` with `e.fmt(f)` and Rust knows to call the right `fmt`. (I don't know how it knows this. It's quite mysterious to me.)

src/bit_machine/mod.rs
```diff
@@ -472,6 +506,37 @@ impl BitMachine {
...
+    fn track_left(&mut self, imr: Imr) {
+        self[0].insert(imr);
+    }
+
+    fn track_right(&mut self, imr: Imr) {
+        self[1].insert(imr);
+    }
+}
+
+impl CaseTracker for () {
```
apoelstra, line 534: In b90069e5b7c4be7c93a12e57c494a11aa45c91da:

We can do this in a followup PR, but we really ought to create a dedicated dummy type for this rather than using `()`.

It's not so important for a private trait but it's cleaner.

src/node/redeem.rs
```diff
@@ -273,6 +274,162 @@ impl<J: Jet> RedeemNode<J> {
...
+        impl<J: Jet> Converter<Redeem<J>, Witness<J>> for Pruner<J> {
+            type Error = std::convert::Infallible;
+
+            fn convert_witness(
+                &mut self,
+                data: &PostOrderIterItem<&RedeemNode<J>>,
+                witness: &Value,
+            ) -> Result<Option<Value>, Self::Error> {
+                let pruned = witness.prune(&data.node.arrow().target);
+                debug_assert!(
```
apoelstra, line 310: In 7d93c829eed0ca9935c595214c5fa526564f4bf8:

I'd prefer these were full asserts. It's an extremely cheap check and if we mess up somehow so that it fails it will result in very mysterious failure modes.

src/node/redeem.rs
```diff
@@ -273,6 +274,162 @@ impl<J: Jet> RedeemNode<J> {
...
+                    .map_witness(Option::<Value>::clone);
+                let retyped = WitnessData::from_inner(&self.inference_context, converted_inner)
+                    .expect("pruned types should check out if unpruned types check out");
+                Ok(retyped)
+            }
+        }
+
+        struct Finalizer<J>(PhantomData<J>);
+
+        impl<J: Jet> Converter<Witness<J>, Redeem<J>> for Finalizer<J> {
```
apoelstra, line 366: In 7d93c829eed0ca9935c595214c5fa526564f4bf8:

This maybe should also be a followup PR, because I'm unsure of correctness, but are you sure we need to do this in two complete passes? Or could we do combine the pruner and finalizer into one?

Even if the answer is no it might be worthwhile to try it, fuzz for programs that pass on one but not the other, and use them as unit tests..

src/node/witness.rs
```diff
@@ -151,21 +151,49 @@ impl<J: Jet> WitnessNode<J> {
...
                 &mut self,
-                _: &PostOrderIterItem<&WitnessNode<J>>,
+                data: &PostOrderIterItem<&WitnessNode<J>>,
                 wit: &Option<Value>,
             ) -> Result<Value, Self::Error> {
                 if let Some(ref wit) = wit {
                     Ok(wit.shallow_clone())
                 } else {
-                    Err(Error::IncompleteFinalization)
+                    // We insert a zero value into unpopulated witness nodes,
```
apoelstra, line 78: In 40e5ab946fd5dfaa7fd97bd936d3918a3a7919bc:

I guess, if we had hidden nodes in our `Node` enum, we could handle this more cleanly (and also get rid of the "you must populate every disconnect node" requirement below)?

src/policy/satisfy.rs
```diff
@@ -91,139 +95,203 @@ impl<Pk: ToXOnlyPubkey> Satisfier<Pk> for elements::LockTime {
...
+    /// This can happen when the [`Satisfier::lookup_asm_program`] method is incorrectly implemented.
+    AssemblyFailed(crate::bit_machine::ExecutionError),
+}
+
+/// A witness expression wrapped in `Ok` if satisfaction was successful
+/// or wrapped in `Err` if satisfaction failed.
+///
+/// This information is necessary for the policy satisfier to select the correct
+/// disjunction (`or`) and threshold (`thres`) branches.
+type SatResult = Result<Arc<WitnessNode<Elements>>, Arc<WitnessNode<Elements>>>;
```
apoelstra, line 117: In c5b5f0299adf43012ee054899850ce31c1121003:

I think we should just use a `(Arc<WitnessNode<Elements>>, bool)` for this type. This would let us get rid of the `ok_if` method and all the calls to `is_ok`. And part of the commit message :).

src/policy/satisfy.rs
```diff
@@ -91,139 +95,203 @@ impl<Pk: ToXOnlyPubkey> Satisfier<Pk> for elements::LockTime {
...
+    ///
+    /// This can happen when the [`Satisfier::lookup_asm_program`] method is incorrectly implemented.
+    AssemblyFailed(crate::bit_machine::ExecutionError),
+}
+
+/// A witness expression wrapped in `Ok` if satisfaction was successful
+/// or wrapped in `Err` if satisfaction failed.
+///
+/// This information is necessary for the policy satisfier to select the correct
+/// disjunction (`or`) and threshold (`thres`) branches.
```
apoelstra, line 116: In c5b5f0299adf43012ee054899850ce31c1121003:

nit: `threshold` should be shortend to `thresh`.

apoelstra:
Done reviewing c1ffa03d6516fc11e672d4ee9fb06eef420b296e. Looking pretty good!

src/policy/satisfy.rs
uncomputable: I tried to make the code as idiomatic as possible, but I might have overdone it here :)

src/node/redeem.rs
uncomputable: I think there are non-local effects in the type inference. We convert the program in post-order, so a witness node etc. cannot know the pruned type that it will receive. _This also means that I implemented the pruning of witness values incorrectly, because I used the unpruned type. I will fix this and extend the unit tests._

uncomputable:
I addressed the comments and fixed the pruning of witness values. I realized that pruned witness data is different from unpruned witness data. This is now reflected in the unit tests.

apoelstra (approved):
ACK 912197936f38aa71cbfbd6b60fa144bc637a48f4; successfully ran local tests; sorry for the delay

PR #260, opened 2025-01-11
clippy fixes: make some lifetimes anonymous and add send/sync to types::BoundRef

### LINKED ISSUE #258: Arcs cannot be sent across threads
The `Arc`s that we use for the inference context cannot be sent between threads. (New clippy lint)

```
warning: usage of an `Arc` that is not `Send` and `Sync`
  --> src/types/context.rs:64:19
   |
64 |             slab: Arc::new(Mutex::new(vec![])),
   |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: `Arc<Mutex<Vec<Bound>>>` is not `Send` and `Sync` as `Mutex<Vec<Bound>>` is neither `Send` nor `Sync`
   = help: if the `Arc` will not used be across threads replace it with an `Rc`
   = help: otherwise make `Mutex<Vec<Bound>>` `Send` and `Sync` or consider a wrapper type such as `Mutex`
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#arc_with_non_send_sync
   = note: `#[warn(clippy::arc_with_non_send_sync)]` on by default
```

The problem stems from `BoundRef`, which includes a raw pointer.

```
`*const Mutex<Vec<types::Bound>>` cannot be sent between threads safely
   --> src/types/context.rs:42:19
    |
42  |     assert_send::<Context>();
    |                   ^^^^^^^ `*const Mutex<Vec<types::Bound>>` cannot be sent between threads safely
    |
    = help: within `UbInner<BoundRef>`, the trait `Send` is not implemented for `*const Mutex<Vec<types::Bound>>`
    = help: the trait `Send` is implemented for `Mutex<T>`
```

Let's discuss a fix.

apoelstra:
`types::BoundRef` is an internal type which includes a sort of weak pointer to an `Arc` containing a slab allocation used by the type inference engine.

It is `Send` and `Sync` by construction but the compiler does not recognize this because it includes a raw pointer. Add explicit implementations with a safety comment explaining why it is okay.

Fixes #258

### DIFF
--- a/src/bit_encoding/decode.rs
+++ b/src/bit_encoding/decode.rs
@@ -133,7 +133,7 @@ enum DecodeNode<J: Jet> {
     Word(Word),
 }
 
-impl<'d, J: Jet> DagLike for (usize, &'d [DecodeNode<J>]) {
+impl<J: Jet> DagLike for (usize, &'_ [DecodeNode<J>]) {
     type Node = DecodeNode<J>;
 
     fn data(&self) -> &DecodeNode<J> {
--- a/src/bit_encoding/encode.rs
+++ b/src/bit_encoding/encode.rs
@@ -35,7 +35,7 @@ impl<'n, N: node::Marker> Disconnectable<EncodeNode<'n, N>> for EncodeNode<'n, N
     }
 }
 
-impl<'n, N: node::Marker> DagLike for EncodeNode<'n, N> {
+impl<N: node::Marker> DagLike for EncodeNode<'_, N> {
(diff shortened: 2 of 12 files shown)


apoelstra (commented):
On 5696096d4c011eaa8dae178ec5a002961c1ceeed successfully ran local tests

uncomputable (approved):
ACK 5696096d4c011eaa8dae178ec5a002961c1ceeed

PR #263, opened 2025-01-28
Correctly free null pointers

### LINKED ISSUE #262: rust_free does not handle NULL the way free does
In the C stdlib you can call `free` on a NULL pointer and it'll be a no-op.

Our code tries to subtract 16 from this, overflows, and then attempts to deallocate the overflowed "address".

uncomputable:
Fixes #262


apoelstra (approved):
ACK 7cf1a9049758fee530be129c54940817f6954f45; successfully ran local tests; thanks! These are great improvements to the docs

PR #264, opened 2025-02-06
Virtual hidden nodes

uncomputable:
This PR adds the wrapper type `Hiding<N>` which allows the use of hidden nodes during the construction of a program. The hidden nodes exist only in the wrapper and the `Inner` struct remains untouched _(therefore "virtual hidden nodes")_.

We use `Hiding<N>` to simplify the policy satisfier and to represent missing assembly fragments, which removes one error path.


uncomputable:
PR fails because of outdated github actions which are updated in #261

src/merkle/mod.rs
```diff
@@ -15,6 +15,12 @@ use crate::Value;
 use hashes::{sha256, Hash, HashEngine};
 use std::fmt;
 
+/// Trait for types that have a Commitment Merkle Root.
+pub trait Merkelized {
```
apoelstra, line 19: In e54c81c286f773a8f1b8727a9fe96e168ccb52d5:

I would call this trait `Cmr`...

src/node/mod.rs
```diff
@@ -407,6 +407,18 @@ where
     }
 }
 
+impl<N: Marker> Merkelized for Node<N> {
+    fn cmr(&self) -> Cmr {
+        self.cmr
+    }
+}
+
+impl<N: Marker> Merkelized for Arc<Node<N>> {
```
apoelstra, line 418: In e54c81c286f773a8f1b8727a9fe96e168ccb52d5:

...but do you even need this trait? It seems like `Borrow<Node<N>>` would work in its place.

src/node/hiding.rs
```diff
@@ -0,0 +1,226 @@
...
+            result: Err(cmr),
+            ctx: self.ctx.shallow_clone(),
+        }
+    }
+
+    /// Access the non-hidden node contained in the wrapper.
+    ///
+    /// Return `None` if the wrapper contains a hidden node.
+    pub const fn as_node(&self) -> Option<&N> {
+        match &self.result {
```
apoelstra, line 63: In 60166205dcb0a46cacba79c3306a8b30505a38b8:

This function can be `self.result.as_ref().ok()`.

src/node/hiding.rs
```diff
@@ -0,0 +1,226 @@
...
+/// If the root node of a program is hidden, then the program is invalid.
+///
+/// ## Retrieval of final program
+///
+/// When the program is fully constructed,
+/// then it can be retrieved via the [`Hiding::get_node`] method.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Hiding<N> {
+    result: HidingResult<N>,
+    ctx: Context,
```
apoelstra, line 38: In 60166205dcb0a46cacba79c3306a8b30505a38b8:

Where is `ctx` used?

uncomputable: The context is needed for `CoreConstructible::inference_ctx`.

src/node/mod.rs
uncomputable: I am running into compile errors about unconstrained type parameters.

```
error[E0207]: the type parameter `M` is not constrained by the impl trait, self type, or predicates
  --> src/node/hiding.rs:73:6
   |
73 | impl<M, N: Borrow<Node<M>>> Hiding<N> {
   |      ^ unconstrained type parameter
```

Here, the `cmr` and `hide` methods fail to compile.

```rust
impl<M: Marker, N: Borrow<Node<M>>> Hiding<N> {
    fn cmr(&self) -> Cmr {
        match &self.result {
            Ok(node) => node.borrow().cmr(),
            Err(cmr) => *cmr,
        }
    }

    /// Convert the node inside the wrapper into a hidden node with the same CMR.
    pub fn hide(self) -> Self {
        match self.result {
            Ok(node) => Self::hidden(node.borrow().cmr(), self.ctx),
            Err(..) => self,
        }
    }
}
```

I was able to fix the `cmr` and `hide` methods, but now I cannot implement `CoreConstructible`.

```rust
impl<N> Hiding<N> {
    fn cmr<M: Marker>(&self) -> Cmr
    where
        N: Borrow<Node<M>>,
    {
        match &self.result {
            Ok(node) => node.borrow().cmr(),
            Err(cmr) => *cmr,
        }
    }

    /// Convert the node inside the wrapper into a hidden node with the same CMR.
    pub fn hide<M: Marker>(self) -> Self
    where
        N: Borrow<Node<M>>
    {
        match self.result {
            Ok(node) => Self::hidden(node.cmr(), self.ctx),
            Err(..) => self,
        }
    }
}
```

uncomputable: Maybe there is a good workaround. I could write a macro that implements code for `Node<N>` and for `Arc<Node<N>>`, but then a trait like `Merkelized` is more canonical.

uncomputable:
Made fuzzing CI work again and addressed the comments.

src/node/mod.rs
apoelstra: No, what you've done is fine. Can you add a comment though, saying that this could be replaced with `Borrow<Node<N>>` except for language limitations in the trait checker. In particular it is not possible to specify `forall<N> T: Borrow<Node<N>>`.

Sorry for leading you on the wild goose chase.

src/node/hiding.rs
apoelstra: Ahh, ok. So now I'm worried about type inference here. I will post a top-level comment. This might be OK but my intuition is that hidden nodes should somehow return a "dummy inference context" while non-hidden nodes can just pass through to `self.result.inference_context()`.

apoelstra:
Ok, after thinking about this more, I think that the type inference story in this PR is correct. But let me write out some thoughts.

In this PR there are three ways to construct a `Hiding` node:

* Calling `Hiding::hidden` and explicitly passing a CMR and a type inference context, which is unused except as an accessor
* Calling `Hiding::from` on some `CoreConstructible` node, which simply wraps the node and stores a copy of its inference context arc.
* Doing the same but then calling `hide`, which drops the original node, keeping its CMR and the handle to its inference context.

When we construct programs/expressions, we always do so in post-order. This is enforced by our API which simply doesn't let you create nodes until their children exist, which in turn means that when you create a new node, its children "haven't seen each other" until the moment of creation. What I mean by this is that the children's sets of type inference variables will be disjoint, and during creation of the new node, bounds may be created that entangle them. (Sharing complicates this a bit but let's ignore that for now.)

The rule for type inference is that hidden nodes do not have type variables. So when you put a hidden node into a combinator with anything else, there are no new type bounds and the other child's type variables are unaffected. In fact, they are not just unaffected but they are hidden -- the new node will be hidden, and therefore also have "no type inference variables", and any variables that its children might've had before being hidden are effectively gone. (Again, with sharing maybe these variables are still accessible elsewhere, but we can ignore that.)

This code **is correct** in that when you, for example, create a new `comp` combinator, if either child is hidden we manually combine their CMRs, preserve the type inference engine, and do nothing else. And only if neither child is hidden, then we pass through to `CoreConstructible::comp` on the two children, which will add bounds between the children's type inference variables.

**However**, it makes me a little nervous that in the hiding case, we carefully preserve the type inference variables/bounds/arrows/etc of the children by preserving the type inference context. I wish there was a way that we could say "hidden node, no types to see here".

apoelstra:
API-wise I don't see any good way to do what I'm describing -- we could drop the `ctx` member of `Hiding` at the expense of adding a `types::Context::new_dummy` constuctor and updating the context-consistency checks to say that a dummy is "the same context" as any other one, and so on, but the result would be a lot of extra complicated code which might even be more confusing than what you've got here.

Could you put my above comment, or maybe some subset of it, as a block comment in the hiding.rs file? Maybe as a doccomment on the private `ctx` field?

apoelstra:
0b2ae765218813363c7ad434fbd6011a9f900e40 looks good other than my above comments, which are exclusively "can you add more comments".

uncomputable:
In the end, the hidden nodes in this PR are simulated and the rule that they must not have type inference variables does not apply. We do throw away expressions and they types when we hide them, but that should not be a problem.

I agree that it is confusing that hidden nodes (even if they are simulated) have a type inference context. However, as long as `CoreConstructible::inference_context` exists, no one use how this method is used. For soundness, the same context should be returned for all nodes of the same program.

uncomputable:
I updated the documentation of `Hiding` and added comments that explain soundness.

src/node/hiding.rs
```diff
@@ -0,0 +1,245 @@
...
+/// 4. Any node with non-hidden children remains unchanged.
+///
+/// The program can be extracted from the wrapper when construction is finished.
+/// The program is invalid if the root node is "hidden".
+///
+/// ## Wrapping
+///
+/// A node can be wrapped via [`Hidden::from`] to add hiding support.
+/// A wrapped node can be converted into a "hidden" node via [`Hidden::hide`].
+/// Finally, a "hidden" node can be manually created via [`Hidden::hidden`].
```
apoelstra, line 27: In 8bc68ac1663d0c47bb745a87df168e54a7113be4:

In several places in these docs you say `Hidden` but the type is called `Hiding`.

uncomputable: Good catch. Thanks for pointing out the typo.

apoelstra (approved):
ACK 29dd9ea0d51e429da57b2bf990005ca4e8b491cd; successfully ran local tests; awesome, thanks!

PR #265, opened 2025-02-07
Remove WitnessNode

uncomputable:
Remove `WitnessNode` by merging its code with the nearly identical `ConstructNode`.

Builds upon #264


apoelstra:
In ce012c54f041cda5a560e660fb0c826c02ac844e:

I'm really confused by the claim that you can remove an error variant by just putting incorrect data into the program. Can you elaborate why this is reasonable?

apoelstra:
Oh, I see, this is only in the `SimpleFinalizer` which is basically a test-only structure which can already easily be cajoled into putting invalid data into a program. Yeah, we definitely shouldn't have error variants in the top-level `Error` enum that are only used by this thing.

uncomputable:
> putting incorrect data into the program

My approach is to insert reasonable default data into a program during finalization. The data does not change the CMR of the program. The user will see how the program fails on the Bit Machine and be able to make adjustments based on that.

apoelstra:
> > 
> 
> My approach is to insert reasonable default data into a program during finalization. The data does not change the CMR of the program. The user will see how the program fails on the Bit Machine and be able to make adjustments based on that.

It's not "reasonable default data". It's unremarkable, possibly-valid data that was put into a program without the user's knowledge and will result in incredibly confusing errors. The user will not "see how the program fails on the Bit Machine". If they are lucky the program will fail at all, and then they will need to sift through their witness data and find the magic zero bits that were fabricated by this library.

I agree that we should strive to make debugging and errors as simple as possible. I do not believe that it will ever be possible for debugging to be so simple that it's OK for the development toolchain to be silently inserting bad data into real programs.

It's fine if this is test-only code or code that we intend to rip out. (As noted in its doccomment, this finalizer already does no type-checking even though this is now easily possible with our API, and it also can't handle disconnect, and it doesn't attempt any pruning, etc etc). But if you are saying that this is something you seriously want to do, then NACK.

uncomputable:
I'm surprised by the backlash since the idea of reducing error paths during finalization is a continuation from the zero values from #259. In a test program made-up default values might make a program run on the Bit Machine, but if the same happens in a production program then your program is insecure. The library will never accidentally make up a valid signature, etc. But yeah, it would be good to inform the user which spending paths were taken.

In any case, this PR only touches `SimpleFinalizer` which is used in unit tests. The merger of `WitnessNode` into `CommitNode` has nothing to do with finalization error handling.

apoelstra:
The zero values in #259 are only used in places where we expect that they will never be visible to the user (because they will be converted to hidden nodes and deleted). There is a giant block comment explaining this. And furthermore, I had hoped that we would be able to get rid of this once we had hidden nodes again.

apoelstra:
Can you add a block comment to the `SimpleFinalizer` saying that this finalizer is only expected to be used for tests and demos, that synthesizing dummy data inside a finalizer is not reasonable production behavior, and that we will revisit this at some point in the future?

uncomputable:
I added a disclaimer to `SimpleFinalizer` that is should not be used in production and why.

I will use `Hiding` to handle missing paths during finalization in a follow-up PR. That doesn't introduce unexpected default values.

apoelstra (approved):
ACK 61b7ddc5cc95a67d54df19ec36586856da502bc4; successfully ran local tests

Issue #146, opened 2023-06-29
Values as byte vectors


uncomputable:
It might be easier to represent values as byte vectors along with a `Final` type. Most values (that we interact with) are byte strings or slight variations thereof. We could destruct the value by using `Final` if needed. TMRs can be used to detect powers of two.

PR #266, opened 2025-02-10
heavily optimize the `Value` type

apoelstra:
This PR has many commits but only one I would say is "big". This logic is all pretty-heavily covered by tests, and I added more tests, as well as benchmarks, so hopefully you don't need to review the fiddly bit-manipulation stuff too hard.

It also leaves the `value` module in a somewhat messier state than when it started -- we have a new `ValueRef` type which duplicates some functionality of `Value` but isn't exposed in the public API; we now implement `DagLike` on `Value` itself rather than `&Value`, but using it is pretty slow compared to using it on `ValueRef`. The `Eq` and `Ord` traits are now implemented manually, but `Eq` requires a bit of computation and `Ord`, despite being implemented manually, still exposes a pretty-arbitrary ordering on values which have different types.

So it is probably worthwhile to follow up on this with a PR that overhauls the public API and tries to provide a cleaner interface.

**However**, this PR also introduces some massive speedups. It introduces new benchmarks and has their output in some commit messages. The highlights are as follows:

Prior to this PR:
```
    benches::bench_value_create_64k               ... bench:   2,565,318.90 ns/iter (+/- 40,140.60)
    benches::bench_value_create_64k_compact       ... bench: 214,621,296.60 ns/iter (+/- 23,731,924.55)
    benches::bench_value_display_64k              ... bench:  17,526,983.30 ns/iter (+/- 339,118.95)
````

With this PR:
```
    benches::bench_value_create_64k               ... bench:     875,039.00 ns/iter (+/- 5,007.70)
    benches::bench_value_create_64k_compact       ... bench:      45,258.88 ns/iter (+/- 169.22)
    benches::bench_value_display_64k              ... bench:   9,685,460.80 ns/iter (+/- 48,509.31)
```

In other words, directly creating a value which is a 64 kilobyte (not kilobit) bitstring used to take 2.5ms; displaying took 17ms; and parsing it from a bit iterator took **214** milliseconds. Now these take 0.875ms, 9ms and 0.045ms, respectively. These are a 2.8, 1.8, and 4700x speedup respectively.

The latter operation in particular, creating values from bitstrings and types, has been a major bottleneck in my fuzzer and is the motivation for this PR.

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -5,49 +5,74 @@
 //! Simplicity processes data in terms of [`Value`]s,
 //! i.e., inputs, intermediate results and outputs.
 
-use crate::dag::{Dag, DagLike, NoSharing};
+use crate::dag::{Dag, DagLike};
 use crate::types::{CompleteBound, Final};
+use crate::BitIter;
 
 use crate::{BitCollector, EarlyEndOfStreamError};
+use core::{cmp, fmt, iter};
 use std::collections::VecDeque;
-use std::fmt;
 use std::hash::Hash;
 use std::sync::Arc;
 
 /// A Simplicity value.
-#[derive(Clone, Eq, PartialEq, Ord, PartialOrd, Hash)]
+#[derive(Clone)]
 pub struct Value {
-    inner: ValueInner,
+    /// The underlying data, in "padded" bit-encoded form, with the leftmost
+    /// bits encoded in the most-significant bits. So e.g. right(unit) is 0x80.
+    ///
+    /// We use the padded form, even though it is space-inefficient (sometimes
+    /// significantly so), for a couple reasons: first, this is the representation
+    /// used by the bit machine, so we want encoding/decoding to be as fast as
+    /// possible. Secondly, when destructuring a value (which we need to do during
+    /// scribing and pruning, at least) it is quite difficult to destructure a
+    /// compact-encoded product, since we would need to determine the compact
+    /// bitlengths of the children, and these quantities are not readily available.
+    inner: Arc<[u8]>,
+    /// An offset, in bits, at which the actual data starts. This is useful
+    /// because it allows constructing sub-values of a value without needing
+    /// to construct a new `inner` with all the bits offset.
+    bit_offset: usize,
+    /// The Simplicity type of the value.
     ty: Arc<Final>,
 }
 
-/// The inner structure of a Simplicity value.
-#[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-enum ValueInner {
-    /// The unit value.
-    ///
-    /// The unit value is the only value of the unit type `1`.
-    /// It must be wrapped in left and right values to encode information.
-    Unit,
-    /// A left value.
-    ///
-    /// A left value wraps a value of type `A` and its type is the sum `A + B` for some `B`.
-    /// A `false` bit encodes that a left value is wrapped.
-    Left(Arc<Value>),
-    /// A right value.
-    ///
-    /// A right value wraps a value of type `B` and its type is the sum `A + B` for some `A`.
-    /// A `true` bit encodes that a right value is wrapped.
-    Right(Arc<Value>),
-    /// A product value.
-    ///
-    /// A product value wraps a left value of type `A` and a right value of type `B`,
-    /// and its type is the product `A × B`.
-    /// A product value combines the information of its inner values.
-    Product(Arc<Value>, Arc<Value>),
+// Because two equal values may have different bit offsets, we must manually
+// implement the comparison traits. We do so by first comparing types, which
+// is constant overhead (this just compares TMRs). If those match, we know
+// the lengths and structures match, so we then compare the underlying byte
+// iterators.
+impl PartialEq for Value {
+    fn eq(&self, other: &Self) -> bool {
+        self.ty == other.ty && self.raw_byte_iter().eq(other.raw_byte_iter())
+    }
+}
+impl Eq for Value {}
+
+impl PartialOrd for Value {
+    fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+        Some(self.cmp(other))
+    }
+}
+impl Ord for Value {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        self.ty
+            .cmp(&other.ty)
+            .then_with(|| self.raw_byte_iter().cmp(other.raw_byte_iter()))
+    }
 }
 
-impl DagLike for &'_ Value {
+impl core::hash::Hash for Value {
+    fn hash<H: core::hash::Hasher>(&self, h: &mut H) {
+        b"Simplicity\x1fValue".hash(h);
+        self.ty.hash(h);
+        for val in self.raw_byte_iter() {
+            val.hash(h);
+        }
+    }
+}
+
+impl DagLike for Value {
     type Node = Value;
 
     fn data(&self) -> &Value {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

**Comment by apoelstra** [comment]
Drafting overnight (unless I can fix this in the next couple minutes).

Can crash it with the following regression test
```diff

diff --git a/src/value.rs b/src/value.rs
index 525b163..20d069c 100644
--- a/src/value.rs
+++ b/src/value.rs
@@ -854,4 +854,20 @@ mod tests {
     }

+    #[test]
+    fn prune_regression_1() {
+        // Found this when fuzzing Elements; unsure how to reduce it further.
+        let nontrivial_sum = Value::product(
+            Value::right(Final::two_two_n(4), Value::u16(0)),
+            Value::u8(0),
+        );
+        // Formatting should succeed and have no effect.
+        let _ = format!("{nontrivial_sum}");
+        // Pruning should succeed and have no effect.
+        assert_eq!(
+            nontrivial_sum.prune(nontrivial_sum.ty()),
+            Some(nontrivial_sum)
+        );
+    }
+
     #[test]
     fn prune() {
```

apoelstra (commented):
On d7a9d2cde3fe5408ae2acbdef2a68b5dbeabab24 successfully ran local tests

apoelstra (commented):
On 987a8a20fadc1334a19a649bc458b6543095fb6a successfully ran local tests

src/value.rs
```diff
@@ -927,6 +928,24 @@ mod tests {
...
+    }
+
+    #[test]
+    fn padded_round_trip() {
+        let v = Value::u64(0xff00_00ff_ff00_00ff);
+        let (bits, _) = v.iter_padded().collect_bits();
+        let mut iter = BitIter::new(bits.into_iter());
+        let new_v = Value::from_padded_bits(&mut iter, &v.ty).unwrap();
+        assert_eq!(v, new_v);
+    }
```
uncomputable, line 1361: 790b38a2abe7883c393ab8be1e14906feb5c5112: Value encoding roundtrips are candidates for future fuzz tests

src/value.rs
```diff
@@ -5,69 +5,231 @@
...
+        Some(self.cmp(other))
+    }
+}
+impl Ord for Value {
+    fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+        self.ty
+            .cmp(&other.ty)
+            .then_with(|| self.raw_byte_iter().cmp(other.raw_byte_iter()))
+    }
+}
```
uncomputable, line 71: cbbe9b5a3e7982c5b4c85b28cc37996fbcb0c7b2: In a follow-up PR, we should remove the `Ord` implementation. It makes no semantic sense to compare values of different types. Rather, `PartialOrd` could return `None` if `self.ty != other.ty`.

src/value.rs
```diff
@@ -5,69 +5,231 @@
...
+
+impl core::hash::Hash for Value {
+    fn hash<H: core::hash::Hasher>(&self, h: &mut H) {
+        "[simplicity value]".hash(h);
```
uncomputable, line 75: cbbe9b5a3e7982c5b4c85b28cc37996fbcb0c7b2: This is a don't care value, but maybe we should use `"Simplicity\x1fValue"` to keep consistency with the Merkle roots.

src/value.rs
```diff
@@ -140,27 +330,76 @@ impl Value {
...
     pub fn to_left(&self) -> Option<Self> {
-        match &self.inner {
-            ValueInner::Left(inner) => Some(inner.shallow_clone()),
-            _ => None,
+        if self.first_bit() == Some(false) {
+            if let Some((lty, rty)) = self.ty.as_sum() {
+                let sum_width = 1 + cmp::max(lty.bit_width(), rty.bit_width());
+                Some(Self {
+                    inner: Arc::clone(&self.inner),
+                    bit_offset: self.bit_offset + sum_width - lty.bit_width(),
```
uncomputable, line 454: cbbe9b5a3e7982c5b4c85b28cc37996fbcb0c7b2: `bit_offset: self.bit_offset + self.ty.bit_width() - lty.bit_width()`

The variable `sum_width` is not required.

src/value.rs
```diff
@@ -140,27 +330,76 @@ impl Value {
...
     pub fn to_right(&self) -> Option<Self> {
-        match &self.inner {
-            ValueInner::Right(inner) => Some(inner.shallow_clone()),
-            _ => None,
+        if self.first_bit() == Some(true) {
+            if let Some((lty, rty)) = self.ty.as_sum() {
+                let sum_width = 1 + cmp::max(lty.bit_width(), rty.bit_width());
+                Some(Self {
+                    inner: Arc::clone(&self.inner),
+                    bit_offset: self.bit_offset + sum_width - rty.bit_width(),
```
uncomputable, line 472: cbbe9b5a3e7982c5b4c85b28cc37996fbcb0c7b2: `bit_offset: self.bit_offset + self.ty.bit_width() - rty.bit_width()`

src/value.rs
```diff
@@ -416,90 +695,88 @@ impl Value {
...
             }
         }
         Ok(())
     }
 }
-
 /// An iterator over the bits of the padded encoding of a [`Value`].
 #[derive(Debug, Clone, Eq, PartialEq, Hash)]
-struct PaddedBitsIter {
+struct CompactBitsIter {
```
uncomputable: cbbe9b5a3e7982c5b4c85b28cc37996fbcb0c7b2: This documentation talks about padded encoding, but the iterator handles compact encoding.

src/value.rs
```diff
@@ -5,69 +5,231 @@
...
+pub struct PreOrderIter<'v> {
+    inner: iter::Take<BitIter<RawByteIter<'v>>>,
+}
+
+impl Iterator for PreOrderIter<'_> {
+    type Item = bool;
+    fn next(&mut self) -> Option<Self::Item> {
+        self.inner.next()
+    }
+}
```
uncomputable, line 223: cbbe9b5a3e7982c5b4c85b28cc37996fbcb0c7b2: Is this type useful? AFAICS, we can simply remove it. The only method that calls it returns an opaque `impl Iterator` type.

src/value.rs
```diff
@@ -960,27 +960,57 @@ impl Value {
...
+                    result_stack.push(Value::u2(bits.read_u2()?.into()));
+                }
+                State::ProcessType(ty) if ty.tmr() == Tmr::POWERS_OF_TWO[2] => {
+                    let u4 = (u8::from(bits.read_u2()?) << 2) + u8::from(bits.read_u2()?);
+                    result_stack.push(Value::u4(u4));
+                }
+                State::ProcessType(ty) => {
+                    // The POWERS_OF_TWO array is somewhat misnamed; the ith index contains
+                    // the TMR of TWO^(2^n). So e.g. the 0th index is 2 (a bit), the 1st is
+                    // u2, then u4, and the 3rd is u8.
```
uncomputable, line 978: 987a8a20fadc1334a19a649bc458b6543095fb6a: This comment should be in the documentation of `POWERS_OF_TWO` / we should rename `POWERS_OF_TWO` in a follow-up commit.

src/value.rs
```diff
@@ -947,8 +947,8 @@ impl Padding for PaddedEncoding {
 }
 
 impl Value {
-    fn from_bits<I: Iterator<Item = bool>, P: Padding>(
-        bits: &mut I,
+    fn from_bits<I: Iterator<Item = u8>, P: Padding>(
+        bits: &mut BitIter<I>,
```
uncomputable, line 951: 386622530d3f956ca19d74feed7759b3c69fbf3a: This breaks downstream code in the web IDE Bit Machine. However, since `Value` now contains `Arc<[u8]>` it makes sense to use an iterator over something like `[u8]`. In the web IDE, I can collect the bits in `Vec<u8>` and create a `BitIter` from that.

uncomputable:
Fixes #146

uncomputable (changes requested):
987a8a20fadc1334a19a649bc458b6543095fb6a: Awesome work. Storing values in a format that is compatible with the Bit Machine and with the most common value type -- words -- is a great improvement.

I reviewed the code and left some comments. Most are nits. There are a few things that I think should be changed before we merge.

src/value.rs
```diff
@@ -132,19 +126,18 @@ impl Value {
...
-        self.iter_padded().count()
+        self.ty.bit_width()
     }
 
     /// Check if the value is a nested product of units.
     /// In this case, the value contains no information.
     pub fn is_empty(&self) -> bool {
-        self.pre_order_iter::<NoSharing>()
-            .all(|value| matches!(&value.inner, ValueInner::Unit | ValueInner::Product(..)))
+        self.ty.bit_width() == 0
```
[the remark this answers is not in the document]
apoelstra: If the padded bitwidth is 0, then the compact bitwidth will also be zero :).

We could add a dedicated "is_zero_size" accessor to `Final` if that would make you more comfortable.

src/value.rs
apoelstra: Yes -- I tried this, but we then lose `Ord` on `Word` (which maaybe makes sense, you can compare them numerically though you have to decide the relationship between 0u4 and 0u16 etc), and on `Node` and on `Policy`... so it's far-reaching.

I think the right approach would be to change `Policy` to implement `Ord` manually in alphabetical order (we do this in rust-miniscript now; it was a PITA to implement) and then we can drop `Ord` on the smaller types.

In any case, out of scope for this PR.

src/value.rs
apoelstra: Sure!

src/value.rs
apoelstra: ooo nice catch

src/value.rs
apoelstra: Yeah, lemme double-back and check on this. This was one of the first things I implemented and the PR has been almost entirely rewritten since then.

src/value.rs
apoelstra: Yeah, can you just collect into a `Vec<u8>` for now and raise an issue about it? After doing this I really want to revisit the `BitIter` API here. It's awkward in several ways.

src/value.rs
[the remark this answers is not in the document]
uncomputable: My comment also concerned line 129 `self.ty.bit_width()` above.

I don't think we need special `is_zero_size` methods

src/value.rs
apoelstra: Changed both to return concrete types instead of `impl Iterator`. If the concrete types are available then they're much nicer for the user -- they are sized, we can implement additional traits on them, they can store them in structs, etc.

apoelstra:
Addressed every comment labelled on cbbe9b5. I believe all the rest are followups or non-actionable.

apoelstra (commented):
On be8d437d0149697285be42b9732cd222955ce7b9 successfully ran local tests

src/value.rs
uncomputable: We can store opaque `impl` types in structs using generics, can't we?

In any case, now `PreOrderIter` is exposed via the API and slightly more useful. Good enough.

uncomputable (approved):
ACK be8d437d0149697285be42b9732cd222955ce7b9

src/value.rs
apoelstra: > We can store opaque `impl` types in structs using generics, can't we?

I'm pretty sure not. You can try it in the playground.

PR #269, opened 2025-02-10
Followups to #266 -- more `Value` optimizations

### LINKED ISSUE #268: Rename Tmr::POWERS_OF_TWO
The `POWERS_OF_TWO` array is somewhat misnamed; the ith index contains the TMR of TWO^(2^n). So e.g. the 0th index is 2 (a bit), the 1st is u2, then u4, and the 3rd is u8.

We should rename `POWERS_OF_TWO` and clarify its documentation.

apoelstra:
This cleans up the `Value` API by making `ValueRef` public and using it in place of the slow and redundant `Value::to_*` methods.

It also renames `Tmr::POWERS_OF_TWO` which was noticed as a confusing name.

It adds a `types::Final::has_padding` accessor and uses it to dramatically speed up bit decoding for values.

During pruning, it notices when there is nothing to do (when pruning a value to the type that it already has) and shortcuts.

Unlike #266, this one leaves the API in better shape than it was before, and should be reasonably easy to review. It hsa a lot of commits but they are all pretty small.

Fixes #268

### DIFF
--- a/src/value.rs
+++ b/src/value.rs
@@ -9,7 +9,7 @@ use crate::dag::{Dag, DagLike};
 use crate::types::{CompleteBound, Final};
 use crate::BitIter;
 
-use crate::{BitCollector, EarlyEndOfStreamError, Tmr};
+use crate::{BitCollector, EarlyEndOfStreamError};
 use core::{cmp, fmt, iter};
 use std::collections::VecDeque;
 use std::hash::Hash;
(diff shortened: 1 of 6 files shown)


apoelstra (commented):
On e025850c1cbf1c03df9624619c392ac6220fed5e successfully ran local tests

src/value.rs
```diff
@@ -629,43 +629,11 @@ impl Value {
...
-                Task::MakeProduct => {
-                    let r_value = output.pop().unwrap();
-                    let l_value = output.pop().unwrap();
-                    output.push(Value::product(l_value, r_value));
-                }
-            }
+        Self {
+            inner: Arc::from(vec![0; (ty.bit_width() + 7) / 8]),
+            bit_offset: 0,
+            ty: Arc::new(ty.clone()),
```
uncomputable, line 635: 843fb2bd63c8446f56df93a6575f2c5bc63e3864: Nice catch.

src/value.rs
```diff
@@ -845,44 +845,8 @@ impl Iterator for CompactBitsIter<'_> {
...
-        Ok(())
-    }
-}
-
 impl Value {
-    fn from_bits<I: Iterator<Item = u8>, P: Padding>(
+    fn from_bits<I: Iterator<Item = u8>>(
         bits: &mut BitIter<I>,
         ty: &Final,
     ) -> Result<Self, EarlyEndOfStreamError> {
```
uncomputable, line 823: 00729173ea8d8b3044c3c034a0b546745c2ac63e: The body of `from_bits` should be moved to `from_compact_bits` and `from_bits` should be removed.

uncomputable (commented):
e025850c1cbf1c03df9624619c392ac6220fed5e: Very nice. Please remove `from_bits`, which is no longer necessary.

apoelstra:
Sure. Added a commit which does this.

apoelstra (commented):
On de20a3b005e6089527cbf6a78fc1d1a45598964f successfully ran local tests

uncomputable (approved):
ACK de20a3b005e6089527cbf6a78fc1d1a45598964f

PR #270, opened 2025-02-12
update libsimplicity to c54a8db0e4c69a571aa67c3bc3188eb33ca67dcb

apoelstra:
This matches Elements d020546243677f8490685cae06da626f971ba4a7.


uncomputable:
Is this version compatible with what's running on Liquid testnet? We should have at least one branch that is compatible.

uncomputable:
@apoelstra We should pick this PR back up

apoelstra (commented):
On a647f022c0053f345adaae759a99aa28cb0388b5 successfully ran local tests

apoelstra:
Whew. @uncomputable this is ready for review.

fuzz/fuzz_targets/c_rust_merkle.rs
```diff
@@ -46,11 +46,11 @@ fn do_test(data: &[u8]) {
             }
             println!("{}", rust_data.cmr());
             println!("{}", rust_data.amr());
-            println!("{}", rust_data.imr());
+            println!("{}", rust_data.ihr());
```
uncomputable, line 49: e4522021dc59d682901a5a5a2982a65c4ba7d4b1: @apoelstra @roconnor-blockstream Are IMRs entirely replaced by IHRs? I thought, some use cases are called IMR and some are called IHR, in which case a blind find-and-replace also renames use cases where IMR is still the valid term.

uncomputable (commented):
a647f022c0053f345adaae759a99aa28cb0388b5 Looks good, except for the question of keeping IMRs in the valid places

fuzz/fuzz_targets/c_rust_merkle.rs
apoelstra: Grepping the Simplicity codebase for `IMR` and `Identity Merkle Root` shows some scattered remaining references (e.g. in the tech report it appears no renaming has been done) but nothing that looks like a principled distinction.

I believe the literal text `imr_` stays in the preimage of the IV for IHR computation for witness and disconnect nodes, but this codebase never sees the preimage text. We just use precomputed IVs that we get from the C code.

apoelstra:
Cool -- gonna go ahead and merge this. If we need to do further (un-)renames we can do it in a followup PR.

uncomputable (approved):
ACK a647f022c0053f345adaae759a99aa28cb0388b5 I opened an issue to handle the distinction IMR / IHR later.

PR #271, opened 2025-02-13
bit machine: fix some overflows with massive types

apoelstra:
There probably are other things we can break with this test program, but these are the ones I encountered just now.

### DIFF
--- a/src/bit_machine/limits.rs
+++ b/src/bit_machine/limits.rs
@@ -1,0 +1,122 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Bit Machine Resource Limits
+//!
+//! Implementation of the Bit Machine, without TCO, as TCO precludes some
+//! frame management optimizations which can be used to great benefit.
+//!
+
+use core::fmt;
+use std::error;
+
+/// The maximum number of cells needed by the bit machine.
+///
+/// This roughly corresponds to the maximum size of any type used by a
+/// program, in bits. In a blockchain context this should be limited by
+/// the transaction's weight budget.
+///
+/// The limit here is an absolute limit enforced by the library to avoid
+/// unbounded allocations.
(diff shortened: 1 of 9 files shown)


apoelstra (commented):
On 72e1dc3dc17bb0830e71f693b672667aae4dfd12 successfully ran local tests

src/bit_machine/limits.rs
```diff
@@ -0,0 +1,120 @@
...
+            Ok(())
+        }
+    }
+
+    /// Helper function to check every value and sum for being within bounds.
+    pub(super) fn check_program(
+        source_ty_width: usize,
+        target_ty_width: usize,
+        bounds: &crate::NodeBounds,
+    ) -> Result<(), Self> {
```
uncomputable, line 100: 2110376b0556fe2bbc1e2a5e79f04ed68f446ba3: Is there a reason why the argument list is not simply `program: &RedeemNode<J>` ?

uncomputable (commented):
72e1dc3dc17bb0830e71f693b672667aae4dfd12: Nice. Putting limitations on the program size will prevent our fuzzers from producing slow units.

src/bit_machine/limits.rs
apoelstra: Not really, no. IIRC I didn't want to type `<J: Jet>`. I don't remember why I thought that typing `crate::NodeBounds` would be any less annoying.

Let me know if you want to change it.

uncomputable: Let's change the parameter to `&RedeemNode<J>`, since that is the only node type that knows its bounds.

apoelstra (commented):
On b2d83e2127e8df50c5747e264439d76154eb2957 successfully ran local tests

uncomputable (approved):
ACK b2d83e2127e8df50c5747e264439d76154eb2957

PR #276, opened 2025-03-20
Fix Windows

### LINKED ISSUE #274: Tests fail on Windows
The library behaves in strange ways on Windows, which makes the unit tests fail. We should fix the code and enable Windows in CI.

See #272 (where Windows was not enabled in CI for this reason)

uncomputable:
Fixes #274

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -107,7 +107,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-latest]
+        os: [macos-latest, windows-latest]
     steps:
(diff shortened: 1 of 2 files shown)


apoelstra:
May be able to rebase this now that #281 is in. Weird that these failures only happen on Windows though..

uncomputable:
Rebased

uncomputable:
I had to run simplicity-sys unit tests before simplicity-lang unit tests for size and align errors show up in the log. CI passes now.

Rebased onto master.

apoelstra (approved):
ACK 9ae18c68486b675d1acff327ef179158968ce494; successfully ran local tests

PR #283, opened 2025-03-26
Backport #271: integer overflow checks

apoelstra:
Backports several integer overflow checks to the currently-released version of 0.3. These can be used to crash rust-simplicity which is making it hard to write regression fuzztests.

Also changes the version to 0.3.1 so we can release immediately.

Should be easy enough to review with `git range-diff pr/271/head...pr/283/head`.

### DIFF
--- a/Cargo-recent.lock
+++ b/Cargo-recent.lock
@@ -435,7 +435,7 @@ dependencies = [
 
 [[package]]
 name = "simplicity-lang"
-version = "0.3.0"
+version = "0.3.1"
 dependencies = [
  "bitcoin",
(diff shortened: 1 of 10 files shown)


apoelstra:
cc @uncomputable I think this will be a quick review and should let us get #280 in.

uncomputable (approved):
ACK 34aa968a7d5642ff3aabbe8c80df05ee7c7f40f1

apoelstra:
Had to disable diff-checking the copy of libsimplicity vendored in the depend/ directory, since I can't build the old libsimplicity with a modern nixpkgs.

PR #284, opened 2025-03-26
rename `FirstPassIhr` to `Imr`

### LINKED ISSUE #282: Clarify distinction between IMR / IHR
What is the definition of IMR (identity merkle root) in contrast to an IHR (identity hash root)? #270 renamed all usages of `IMR` to `IHR`. However, https://github.com/BlockstreamResearch/simplicity/pull/277 retains some usages of `IMR`.

> What we had been calling the IMR in this code is actually hash of the triple of the identity merkle root, and the type roots of the input and output types.

We might have to rename some usages of `IHR` back to `IMR` in case the latter is the right term.

apoelstra:
In #270 we renamed IMR (Identity Merkle Root) to the IHR (Identity Hash Root). We did this via mass-rename which turned the `FirstPassImr` object into `FirstPassIhr`.

The correct terminogy is now that `FirstPassImr` is simply `Imr`. 
This object continues to be a mostly-internal-only object which is used in service of computing the `Ihr`, which is what users are likely to interact with.

Fixes #282

### DIFF
--- a/src/merkle/ihr.rs
+++ b/src/merkle/ihr.rs
@@ -9,20 +9,20 @@ use hashes::sha256::Midstate;
 
 use super::{bip340_iv, compact_value, FailEntropy};
 
-/// Identity Merkle root (first pass)
+/// Identity Merkle Root
 ///
 /// A Merkle root that commits to a node's combinator, its witness data (if present),
 /// and recursively its children. Used as input to the [`Ihr`] type which is probably
 /// actually what you want.
 #[derive(Copy, Clone, PartialEq, Eq, Hash, PartialOrd, Ord)]
-pub struct FirstPassIhr(Midstate);
+pub struct Imr(Midstate);
 
-impl_midstate_wrapper!(FirstPassIhr);
+impl_midstate_wrapper!(Imr);
 
-/// Identity Merkle root
+/// Identity Hash Root
 ///
-/// A Merkle root that commits to a node's combinator, its witness data (if present),
-/// its source and target types, and recursively its children.
+/// A Merkle root that commits to a node's [`Imr`] (which recursively commits to its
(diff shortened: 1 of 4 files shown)


src/merkle/ihr.rs
```diff
@@ -9,30 +9,30 @@ use hashes::sha256::Midstate;
 
 use super::{bip340_iv, compact_value, FailEntropy};
 
-/// Identity Merkle root (first pass)
+/// Identity Merkle Hash
```
roconnor-blockstream, line 12: Hash -> Root

src/merkle/ihr.rs
```diff
@@ -230,9 +230,9 @@ impl FirstPassIhr {
 impl Ihr {
     /// Do the second pass of the IHR computation. This must be called on the result
     /// of first pass.
-    pub fn compute_pass2(first_pass: FirstPassIhr, ty: &FinalArrow) -> Ihr {
+    pub fn compute_pass2(imr: Imr, ty: &FinalArrow) -> Ihr {
```
roconnor-blockstream, line 233: Without a "pass 1" this could potentially be renamed `compute_ihr`.

roconnor-blockstream (changes requested):

src/merkle/ihr.rs
apoelstra: Ah, I like that.

apoelstra: Changed it to `from_imr`.

uncomputable (approved):
ACK 296abdd1a711f431b0cad935aa5127655b42b6cc That makes a lot of sense. I'm happy that we could get rid of `FirstPassImr`. In a way, we foreshadowed the need for `IHR` in the Rust code.

PR #287, opened 2025-04-05
Fix fuzz

uncomputable:
Fix some tiny things that I found in simplicity-fuzz.


apoelstra (approved):
ACK 463779ec4478fa8c847066268c1eacd0b7bc893a; successfully ran local tests; thanks!!

PR #288, opened 2025-04-09
Generic BitMachine execution tracker

m-kus:
This PR generalizes `CaseTracker` to `ExecTracker` that now supports tracking jet calls (in addition to case branches). It is also now possible to run BitMachine with a custom tracker created outside of the crate.

This mechanim allows to implement jet tracer which pretty-prints jet arguments/result for each invocation, see https://github.com/m-kus/simfony/pull/2/files#diff-13e8b158f84f88059e0bd8631dfa9113112e30d7b8c21a15413d0ef36d2ea180

As a follow up it is proposed to further extend `ExecTracker` trait to allow visitor pattern on case/assert expressions. That would enable debug logging, similar to https://github.com/uncomputable/simfony-webide/blob/adb60d4e60c0d6b019ea13f64b1a166da3467d8d/src/function.rs#L145

Having these two features (jet calls tracking and debug symbols) combined it's fairly simple to output weighted stack traces (e.g. in pprof) and build a Simfony profiler.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,3 +119,16 @@ jobs:
         run: |
           cargo test --locked --manifest-path simplicity-sys/Cargo.toml
           cargo test --locked --workspace --all-features
+
+  docs:
+    name: Docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Crate
+        uses: actions/checkout@v4
+      - name: Checkout Toolchain
+        uses: dtolnay/rust-toolchain@nightly
+      - name: Check that documentation builds without errors
+        env:
+          RUSTDOCFLAGS: "--cfg docsrs -D warnings -D rustdoc::broken-intra-doc-links"
+        run: cargo doc -j1 --all-features
--- a/src/bit_machine/mod.rs
+++ b/src/bit_machine/mod.rs
@@ -20,6 +20,7 @@ use crate::types::Final;
 use crate::{analysis, Ihr};
 use crate::{Cmr, FailEntropy, Value};
 use frame::Frame;
+use simplicity_sys::ffi::UWORD;
 
 pub use self::limits::LimitError;
 
@@ -242,7 +243,12 @@ impl BitMachine {
         Ok(tracker)
     }
 
-    fn exec_with_tracker<J: Jet, T: CaseTracker>(
+    /// Execute the given `program` on the Bit Machine, using the given environment and tracker.
+    ///
+    ///  ## Precondition
+    ///
+    /// The Bit Machine is constructed via [`Self::for_program()`] to ensure enough space.
+    pub fn exec_with_tracker<J: Jet, T: ExecTracker<J>>(
         &mut self,
         program: &RedeemNode<J>,
         env: &J::Environment,
@@ -371,7 +377,7 @@ impl BitMachine {
                     }
                 }
                 node::Inner::Witness(value) => self.write_value(value),
-                node::Inner::Jet(jet) => self.exec_jet(*jet, env)?,
+                node::Inner::Jet(jet) => self.exec_jet(*jet, env, tracker)?,
                 node::Inner::Word(value) => self.write_value(value.as_value()),
                 node::Inner::Fail(entropy) => {
                     return Err(ExecutionError::ReachedFailNode(*entropy))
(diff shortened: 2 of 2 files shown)


src/bit_machine/mod.rs
```diff
@@ -494,13 +494,17 @@ impl BitMachine {
...
+        let (input_read_frame, input_buffer) = unsafe { get_input_frame(self, input_width) };
         let (mut output_write_frame, output_buffer) = unsafe { get_output_frame(output_width) };
 
         let jet_fn = jet.c_jet_ptr();
         let c_env = J::c_jet_env(env);
         let success = jet_fn(&mut output_write_frame, input_read_frame, c_env);
 
+        if cfg!(feature = "trace") {
+            println!("{:?} {:?} -> {:?}", jet, input_buffer, output_buffer);
+        }
```
uncomputable: f6312648cd8d3c4f4649ae01639a146992dce8c9: I feel that we shouldn't add print statements to the code, even when they are feature-gated. Printing could be handled via a trait that is implemented in a downstream crate.

uncomputable (commented):
concept ack: 5d77a6e8d50e27147fb4e8c0cb339d5fec59899f Can the simfony stack trace be implemented with `CaseTracker`? This should be our goal.

src/bit_machine/mod.rs
m-kus: Oh sorry, that must be in the previous commit, I should squash them

apoelstra:
This looks great to me. It's a bit of a sharp API -- your input and output buffers are passed as arrays of `UWORD`s and to actually read the data you need to extract a bitstream from them (and use the jet's type signature to know where to start) but I think this is fine. It avoids any expensive code needing to be run in the normal no-tracker case. We might want to provide some utility functions to help with this or something.

Will ACK this.

uncomputable (dismissed):
ACK 4821b6537c3578714924efb84b36ac83694aa365 I think raw `&[UWORD]` slices are too low level to provide helpful introspection, but I will leave refactors to future PRs.

m-kus:
Using `&[UWORD]` slices is indeed quite low level, but I agree with @apoelstra that it helps to avoid overhead for the no-tracker case. Also it is quite simple to convert to `simplicity::BitIter` and then leverage existing parsers from the Simfony crate to go from bits to high-level types.

src/bit_machine/mod.rs
```diff
@@ -548,20 +569,24 @@ impl SetTracker {
 #[derive(Copy, Clone, Debug)]
 struct NoTracker;
```
apoelstra, line 570: In 4821b6537c3578714924efb84b36ac83694aa365:

I believe we should make `NoTracker` (and probably `SetTracker`) public now.

src/bit_machine/mod.rs
```diff
@@ -510,20 +523,28 @@ impl BitMachine {
     }
 }
 
-/// A type that keeps track of which case branches were executed
-/// during the execution of the Bit Machine.
+/// A type that keeps track of Bit Machine execution.
 ///
-/// The trait is implemented for [`SetTracker`], which does the actual tracking,
+/// The trait is implemented for [`SetTracker`], that tracks which case branches were executed,
 /// and it is implemented for [`NoTracker`], which is a dummy tracker that is
```
apoelstra, line 529: In 4821b6537c3578714924efb84b36ac83694aa365

In my local CI I run

```bash
export 'RUSTDOCFLAGS=--cfg docsrs -D warnings -D rustdoc::broken-intra-doc-links'
cargo doc -j1 --all-features
```

which errors out about the references to `SetTracker` and `NoTracker` which are private types, since these are public docs. We should do this in the real CI as well.

src/bit_machine/mod.rs
m-kus: Fixed

src/bit_machine/mod.rs
m-kus: Added CI job

apoelstra:
@m-kus if you are motivated, to fix the doc test you need to run with the nightly compiler because stable doesn't have the docsrs flags we need for this test.

If you are less motivated you can drop the CI test and I will do it at some point :)

apoelstra:
Great, looks good! I have queued this on my local CI box. There is one rust-bitcoin PR in front of it so it'll hopefully finish in the next hour.

Very cool that you're developing on an arm box! We definitely appreciate the extra testing we get from you doing this..

apoelstra (approved):
ACK 36ff9613917a4b2f339499a9d3d2f96a0c59a319; successfully ran local tests

apoelstra:
@uncomputable can I go ahead and merge this?

.github/workflows/main.yml
```diff
@@ -119,3 +119,16 @@ jobs:
...
           cargo test --locked --workspace --all-features
+
+  docs:
+    name: Docs
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Crate
+        uses: actions/checkout@v4
+      - name: Checkout Toolchain
+        uses: dtolnay/rust-toolchain@nightly
```
uncomputable, line 130: 36ff9613917a4b2f339499a9d3d2f96a0c59a319: an unpinned nightly version might lead to flaky CI, but it should be ok for checking if docs build.

uncomputable (approved):
ACK 36ff9613917a4b2f339499a9d3d2f96a0c59a319

.github/workflows/main.yml
apoelstra: We already have one :). I figured we could pin both instances at once. (Which we should do!)

PR #290, opened 2025-04-14
Version 0.5

uncomputable:
Cut a new release, most prominently with fixes to the type system.


uncomputable:
It turned out that we hadn't published 0.4 yet. No need for 0.5 yet.

PR #293, opened 2025-05-01
Track dbg! calls

m-kus:
This PR extends the `ExecTracker` trait to support tracking debug calls (identity nodes generated by `dbg!` macro in Simfony). It is enabled by feature `debug` to avoid the decoding overhead by default.

Ported from https://github.com/uncomputable/simfony-webide/blob/adb60d4e60c0d6b019ea13f64b1a166da3467d8d/src/function.rs#L144

Example of usage: https://github.com/keep-starknet-strange/stark-symphony/blob/37b0a1892c69ea3b0f2d1c8088563e7fae7c3825/simfony-cli/src/tracker.rs#L65

Sample output
image

### DIFF
--- a/src/bit_machine/mod.rs
+++ b/src/bit_machine/mod.rs
@@ -347,10 +347,23 @@ impl BitMachine {
                     call_stack.push(CallStack::Goto(left));
                 }
                 node::Inner::Case(..) | node::Inner::AssertL(..) | node::Inner::AssertR(..) => {
-                    let choice_bit = self.read[self.read.len() - 1].peek_bit(&self.data);
+                    let in_frame = &self.read[self.read.len() - 1];
+                    let choice_bit: bool = in_frame.peek_bit(&self.data);
 
                     let (sum_a_b, _c) = ip.arrow().source.as_product().unwrap();
                     let (a, b) = sum_a_b.as_sum().unwrap();
+
+                    if tracker.is_track_debug_enabled() {
+                        if let node::Inner::AssertL(_, cmr) = ip.inner() {
+                            let mut bits = in_frame.as_bit_iter(&self.data);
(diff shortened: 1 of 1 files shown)


apoelstra:
In 7c43d07c12a47906fb1de6af35f8d0f2c2f93127:

Rather than feature-gating it, can you add a new trait method `enable_track_debug(&self) -> bool` which has a default impl of `false`, and gate the expensive code on that?

Also sorry for the late review -- I assumed from the name that this PR was much bigger than it was!

m-kus:
Done!

Haha no worries, minimised footprint was the goal :)

src/bit_machine/mod.rs
```diff
@@ -588,6 +613,13 @@ impl<J: Jet> ExecTracker<J> for NoTracker {
     fn track_right(&mut self, _: Ihr) {}
 
     fn track_jet_call(&mut self, _: &J, _: &[UWORD], _: &[UWORD], _: bool) {}
+
+    fn track_dbg_call(&mut self, _: &Cmr, _: Value) {}
+
+    fn is_track_debug_enabled(&self) -> bool {
+        // Set flag to test frame decoding in unit tests
+        cfg!(test)
```
m-kus, line 621: Not ideal, but I don't have a better idea

apoelstra: Hehe, I think it's ok. Agree that it looks pretty weird, but the nice thing is that `cfg(test)` is only ever set within this project, so we don't need to worry about this weirdness somehow being exposed to our dependencies.

apoelstra: I guess the "right" way to do it is to add a whole nother `Tracker` type for use in tests. But I think this is fine for now.

apoelstra (approved):
ACK 2806157d481c571cfb2714e7252fe1b1a6394a05; successfully ran local tests

PR #294, opened 2025-06-27
Add optional arguments budget and env to run_program test helper

m-kus:
`run_program` test helper is quite useful because it allows to run programs in (almost) the same way they would be run in Elements/Liquid node. The execution behaviour should ideally be identical to `BitMachine`,  but there is a possibility of a divergence and using this helper is a good way to catch it.

This PR adds two optional arguments to allow specifying actual budget and environment. This prevents segmentation fault for some programs.


apoelstra (approved):
ACK 7529af3dac2e44babbcd888e05caf897ff97975a; successfully ran local tests; thanks!

PR #295, opened 2025-06-27
ci: fix clippy lints and a couple other small things

apoelstra:
* Fix clippy lints
* Enable a few more lints related to casting
* Fix CI to use the `nightly-version` version of clippy so that we won't break as new lints come out


canndrew:
ACK 48634a8

PR #296, opened 2025-07-10
update libsimplicity to cad9a63292ea8fdd5c1615e0322986f6f996d42f

apoelstra:
Updates libsimplicity for the recent refactorings for the elements/bitcoin split, and also to a commit that uses the new Haskell deps so that I can run my local CI on it.

PR #297, opened 2025-07-14
simplicity-sys: separate out linker symbols in C library by version name

apoelstra:
First several commits (all the ones that don't update libsimplicity) from #296.

### DIFF
--- a/simplicity-sys/Cargo.toml
+++ b/simplicity-sys/Cargo.toml
@@ -8,6 +8,8 @@ documentation = "https://docs.rs/simplicity-sys/"
 description = "FFI bindings to libsimplicity"
 edition = "2021"
 rust-version = "1.78.0"
+build = "build.rs"
+links = "rustsimplicity_0_4"
 
 [build-dependencies]
 cc = "1.0.83"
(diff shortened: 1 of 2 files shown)


apoelstra:
cc @canndrew can you review this? Probably fine to just check that it seems half-sensible. I'm not sure it's worth your time to understand the vagaries of C/Rust FFI.

apoelstra:
For future reference, *this* set of fixes are there because otherwise when we update libsimplicity, our fuzztest (which attempts to compile both the local version of simplicity and the 0.3 release on crates.io) fails to build due to symbol conflicts. This failure manifests in CI and you can reproduce it locally (with some difficulty, since you need to rebase #296 on the commit before this one, or just obtain commit 
ad7a60b998e7ebdbf1c054c36d78240ad645918f from somewhere) by going into the fuzz directory and running

```
RUSTFLAGS=--cfg=fuzzing cargo build --bin construct_type
```

You can see these failures (for now) at https://github.com/BlockstreamResearch/rust-simplicity/actions/runs/16205734263/job/45755435923?pr=296 [and also I have uploaded the logs here so they won't be deleted](https://github.com/user-attachments/files/21233846/logs_41525644690.zip)

apoelstra:
cc @canndrew can you give this a cursory review and an ack?

The other one is not ready because upstream is generating code with broken paths, but this one is good to go.

PR #298, opened 2025-07-23
ci: copy clippy config from main Cargo.toml to simpcli Cargo.toml

apoelstra:
We have the "uninlined format args" lint whitelisted in the main crate but apparently not in simpcli.


canndrew:
ACK ab87c71

PR #299, opened 2025-07-29
introduce base64 encoding and decoding for all nodes

apoelstra:
This also refactors our bit encoding code to no longer be specific to `CommitNode` and RedeemNode`; it introduces some string display methods and `from_str` methods which are able to parse hex witness data alongside base64 programs.


apoelstra:
cc @canndrew this is the last PR before we can start cutting releases

apoelstra (commented):
On 96839b7f36166247b092de98f2b95149cd154cfa successfully ran local tests

PR #300, opened 2025-07-29
bump version of simplicity-lang to 0.5.0

apoelstra:
Also start maintaining a changelog. This library is used by several other projects and it's not changing rapidly anymore.


apoelstra:
Tagged and published.

apoelstra:
Tagged and published.

PR #292, opened 2025-04-26
Automated daily update to rustc (to nightly-2025-08-08)

apoelstra:
Automated update to Github CI workflow `rust.yml` by [create-pull-request](https://github.com/peter-evans/create-pull-request) GitHub action

PR #301, opened 2025-08-02
node/mod.rs: fix new clippy lint

apoelstra:
Should fix #292.


apoelstra (commented):
On 5e9c106a4b6df1cf7b97716f583db54cab618647 successfully ran local tests

PR #305, opened 2025-08-31
types: use GhostCell in place of mutexes in union-bound and type context

apoelstra:
I recently learned about the [GhostCell](https://crates.io/crates/ghost-cell) crate, and its associated paper from ICFP 2021. There are a couple of similar projects from that era, notably `qcell` which predates the published version of ghostcell by a bit, but none of these are used much, as evidenced by the "Dependents" tab on crates.io.

I learned about them from a blog post about ["generativity"](https://arhan.sh/blog/the-generativity-pattern-in-rust/) which has a pattern that makes GhostCells a little more ergonomic-looking at the cost of using macros. (I don't think this is worthwhile and I didn't use them). 

But the macro syntax isn't really the point. The key quote from the blog post is this

>The fundamental purpose of the generativity pattern is not necessarily to improve performance, but to statically require that individual values come from or refer to the same source. The performance benefits are a symptom of this requirement. 

Anyway, GhostCells are extremely powerful and I don't know why they are not more popular. In effect, they allow synchronizing a single zero-sized token, and gating access to an arbitrarily-sized set of otherwise-disconnected objects on them. So if you have a `&mut Token` you can get  mutable access to all of your objects.

This is exactly the situation we have with our type inference context and our union bound algorithm: we have a `Context` object which has a `Mutex` controlling access to a slab of allocated bounds, and then we have a whole pile of `UbElements` which are connected in a forest structure that the Rust compiler doesn't really understand, but where we morally have exclusive access exactly when we have exclusive access to the slab. `GhostCell` lets us tell the compiler this.

This PR

* Eliminates all the mutexes in the type checker except the one in the `Context` object that holds the slab
* Fixes multiple bugs related to inconsistent contexts, including one in the policy module which made it impossible to actually use the asm fragment functionality
* Adds an invariant lifetime `'brand` which acts as a tag on all `Type`s, `ConstructNode`s and `WitnessNode`s which prevent them being combined if they come from different type inference contexts
* Requires the user to do node construction inside a closure which is passed to `Context::with_context`, rather than constructing a context object which then sorta leaks into the ether by being contained in an `Arc` carried by each node

The latter item sounds like a big ergonomic hit, but in practice it's not a big deal because it doesn't affect `CommitNode` or `RedeemNode`.

The use of a lifetime rather than a marker type means that while every type context has a "distinct type" for `Type`, `Arrow`, `ConstructNode`, etc., the distinction is erased after typechecking. So compilation speed and program size will not be adversely affected. But conversely, the compiler error messages are pretty unclear when you use this incorrectly.


uncomputable:
This looks very promising. Are we ready to say that the type inference code is deadlock free, at this point?

apoelstra:
@uncomputable yep!

apoelstra:
Went over this in detail in person with @canndrew. He will ACK when he gets to his computer, I think.

apoelstra:
One followup item is that `Incomplete::from_bound_ref` takes a `Context` but it should take a `LockedContext` (or rather, a `ContextInner`). This is a private method and requires a locked context, so better to require the caller to have already obtained the lock than to have lots of `.lock` sites. I think.

apoelstra (commented):
On 060455ac8254f18d5d7d5472e0f7fdaf12ee3a7f successfully ran local tests

PR #52, opened 2023-01-06
Remove unneeded static

sanket1729:

PR #307, opened 2025-09-05
various FFI cleanups and soundness fixes

apoelstra:
Currently our transaction environment construction is quite complicated, and (as a consequence) unsound in a couple of places. This was implemented in PRs #49 through #52 in the early days of this crate with very little review, using a strategy copied from Haskell but not really appropriate for Rust.

In particular:

* We do not need custom marshalling functions in Rust because we can directly define `repr(C)` structs
* ...in doing so, we may use Rust references in place of raw pointers, as long as the C code upholds the Rust aliasing rules (which it does, at least for these read-only POD types); in fact we may use optional references in place of nullable pointers
* ...and we can use arrays and pointers to arrays in place of pointers to primitive types which are supposed to point to a specific number of elements
* ...but we must make sure that any data we produce a pointer or reference to outlives the pointer/reference!

There was at least one instance where we were creating a temporary variable and extracting a pointer from it, which would then outlive the object it pointed into. (This occurred for the annex, though in dead code, and for the pegin genesis hash, in live code that has never been used.)

As an aside, we used empty enums in a couple places for opaque types. We should not do this, according to the nomicon.


apoelstra (commented):
On ffad4a0cefa3a91305c426788c3c308f706cc752 successfully ran local tests

apoelstra (commented):
On 6678851afa5ba801d28e3d0dfcda59d01c4b9926 successfully ran local tests

apoelstra:
cc @canndrew with this one and #308 I think we can cut a new release (or maybe we can go all the way to bitcoin support, we'll see how things are progressing on libsimplicity)

canndrew:
ACK 405487e

PR #308, opened 2025-09-05
reduce MSRV to 1.74.0

apoelstra:
We increased our MSRV to 1.78.0 because (a) we needed 1.70 or higher for libfuzzer, and (b) LWK is using 1.78.0.

However, we don't expect LWK to be the primary user of this crate -- if we want to copy somebody, we should copy rust-bitcoin, who recently moved up from 1.63 to 1.74.


apoelstra (commented):
On f16771c894f730ad86333f68c25e2a152251cdb9 successfully ran local tests

PR #316, opened 2025-10-14
Add historical benchmark results

roconnor-blockstream:
These benchmark logs make up (or in the future will make up) the consensus critical cost values of jets used in various applications. Adding these logs will be useful so that we have a place to reference them.


apoelstra (approved):
ACK f37d2e9e58da6bbf03813a7adc5e4e6950ebf087; successfully ran local tests

PR #320, opened 2025-11-21
benches: fix divmod_128_64 distribution

apoelstra:
The `div_mod_128_64` jet has preconditions and shortcuts its operation if these are not met. To benchmark correctly we need to sample according to those preconditions.

### DIFF
--- a/jets-bench/src/input.rs
+++ b/jets-bench/src/input.rs
@@ -282,7 +282,7 @@ impl FlatValue {
     }
 
     /// Creates an iterator over the bits of the value
-    pub fn bit_iter(&self) -> FlatValueBitIter {
+    pub fn bit_iter(&self) -> FlatValueBitIter<'_> {
         FlatValueBitIter {
             inner: self,
             n_read: 0,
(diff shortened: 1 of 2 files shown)


apoelstra:
The new distribution is consistently slower than the uniform one. But all results are in a 40-60ns range; I see maybe a 10% slowdown rather than the 2x (or maybe 4x) that I would naively expect.

But I added printfs to the C code to confirm that with the new distribution we are no longer hitting the "shortcut" case, while with the uniform distribution we are. So this is working. I think this is just a case where the marshalling cost is overwhelming the actual computation cost.

jets-bench/src/input.rs
```diff
@@ -1322,6 +1322,66 @@ impl InputSampleExactSize for CheckSigSignature {
...
+            // It has two preconditions:
+            //     1. The high bit of b must be set.
+            //     2. ah (the high 64 bits of a) is strictly less than b.
+            //
+            // See the implementation of `simplicity_div_mod_128_64` in libsimplicity C/jets.c
+            // for how these are implemented.
+
+            // To implement this, we sample two random 64-bit strings, with one of them
+            // having its high bit forced to 1. The higher of the two (which will always
+            // have its high bit 1) is ah, and the other one is b.
```
roconnor-blockstream: The higher of the two is b.

PR #321, opened 2025-11-26
add CONTRIBUTING.md

apoelstra:
If this looks good I'll replicate it for SimplicityHL and rust-elements. (simplicity-webide, simplicity-unchained, and other less-mature projects should probably have a different, less strict, policy.)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,34 @@
+ # PR Structure
+
+ All changes must be submitted in the form of pull requests. Direct pushes
+ to master are not allowed.
+
+ Pull requests:
+
+ * should consist of a logical sequence of clearly defined independent changes
+ * should not contain commits that undo changes introduced by previous commits
+ * must consist of commits which each build and pass unit tests (we do not
+   require linters, formatters, etc., to pass on each commit)
+ * must not contain merge commits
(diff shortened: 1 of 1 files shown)


delta1 (approved):
ACK 2fab9ff39f44db553f520c9db38675e0a53afabf

LGTM

KyrylR (approved):

apoelstra:
For SimplicityHL we probably want to be more friendly. For this crate, which is pretty low-level, I'd rather not be. In neither case do we want to find ourselves running a kindergarten for low-compentence "contributors" trying to pad their resumes by wasting our time.

We also do not want any code of conduct, let alone one with explicitly named conduct police. If you think there's value in one we can use the one-line ESR [If you are more annoying to work with than your contributions justify, you'll be ejected.](https://x.com/esrtweet/status/1971768345188844003) one. Maybe it would discourage would-be conduct police from joining the project.

apoelstra:
> Would it be possible to hook the `github-merge.py` into github merge button as an action? I am guessing not since it requires a signature.

No. No matter what Github will put a bright green "screw up the repo" button on every single PR and you cannot disable it without also disabling pushing normal merge commits.

> But would be nice.

Yep.

PR #323, opened 2025-12-06
Add `prune_with_tracker` method

gerau:
Add `prune_with_tracker` method to `node::redeem::RedeemNode` to allow passing custom execution trackers during node pruning. The existing `prune` method is now a wrapper around this new method; it passes `SetTracker` to preserve the original behavior.

Also modify the `bit_machine::ExecTracker` trait to include two new methods: `contains_left` and `contains_right`. These check if the left or right branch of a case node contains a given `Ihr`, allowing us to replicate the `SetTracker` logic inside old `prune` method.

Removed `bit_machine::BitMachine::exec_prune`, because after the change, it appears that this function is no longer used anywhere. Since it is visible only within the crate, I assume it is safe to remove.

### DIFF
--- a/src/bit_machine/mod.rs
+++ b/src/bit_machine/mod.rs
@@ -223,26 +223,6 @@ impl BitMachine {
         self.exec_with_tracker(program, env, &mut NoTracker)
     }
 
-    /// Execute the given `program` on the Bit Machine and track executed case branches.
-    ///
-    /// If the program runs successfully, then two sets of IHRs are returned:
-    ///
-    /// 1) The IHRs of case nodes whose _left_ branch was executed.
-    /// 2) The IHRs of case nodes whose _right_ branch was executed.
-    ///
-    /// ## Precondition
-    ///
-    /// The Bit Machine is constructed via [`Self::for_program()`] to ensure enough space.
-    pub(crate) fn exec_prune<J: Jet>(
-        &mut self,
-        program: &RedeemNode<J>,
-        env: &J::Environment,
-    ) -> Result<SetTracker, ExecutionError> {
-        let mut tracker = SetTracker::default();
-        self.exec_with_tracker(program, env, &mut tracker)?;
-        Ok(tracker)
-    }
-
     /// Execute the given `program` on the Bit Machine, using the given environment and tracker.
     ///
     ///  ## Precondition
@@ -544,6 +524,12 @@ impl BitMachine {
 ///
 /// The trait enables us to turn tracking on or off depending on a generic parameter.
 pub trait ExecTracker<J: Jet> {
+    /// Check if left branch of the case node contains given `ihr`.
+    fn contains_left(&self, ihr: &Ihr) -> bool;
+
+    /// Check if right branch of the case node contains given `ihr`.
+    fn contains_right(&self, ihr: &Ihr) -> bool;
+
     /// Track the execution of the left branch of the case node with the given `ihr`.
     fn track_left(&mut self, ihr: Ihr);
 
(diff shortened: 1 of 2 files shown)


KyrylR:
By adding the ability to pass a tracker here in `prune_with_tracker`, you also delegate responsibility to the user to know the library’s internals; in particular, the user is now responsible for making these work:

```rust
    /// Check if left branch of the case node contains given `ihr`.
    fn contains_left(&self, ihr: &Ihr) -> bool;

    /// Check if right branch of the case node contains given `ihr`.
    fn contains_right(&self, ihr: &Ihr) -> bool;

    /// Track the execution of the left branch of the case node with the given `ihr`.
    fn track_left(&mut self, ihr: Ihr);

    /// Track the execution of the right branch of the case node with the given `ihr`.
    fn track_right(&mut self, ihr: Ihr);
 ```
 
Which, in my opinion, is not the right way to go.

By looking more closely at the `prune` function, I see why the tracker was not passed before, though that does not mean we should sacrifice developer experience.

Instead, I would propose adding more capabilities to the `SetTracker`.

You can add optional `Sink` functionality and have a function like `prune_with_debug`, where you would pass this new structure.

Here is a rough idea of how sink logic can be handled:

```rust
fn default_debug_sink(label: &str, value: &dyn core::fmt::Display) {
    println!("DBG: {label} = {value}");
}

fn default_jet_trace_sink(jet: Elements, args: &[Value], result: &Value) {
    print!("{jet:?}(");
    for (i, a) in args.iter().enumerate() {
        if i > 0 {
            print!(", ");
        }
        print!("{a}");
    }
    println!(") = {result}");
}

    /// Create a new tracker bound to the provided debug symbol table.
    #[must_use]
    pub fn new(debug_symbols: &'a DebugSymbols) -> Self {
        Self {
            debug_symbols,
            debug_sink: None,
            jet_trace_sink: None,
        }
    }

    /// Enable forwarding of debug!() calls to the provided sink.
    #[must_use]
    pub fn with_debug_sink<F>(mut self, sink: F) -> Self
    where
        F: FnMut(&str, &dyn core::fmt::Display) + 'a,
    {
        self.debug_sink = Some(Box::new(sink));
        Self { ..self }
    }

    /// Enable the default debug!() sink that prints to stdout.
    #[must_use]
    pub fn with_default_debug_sink(self) -> Self {
        self.with_debug_sink(default_debug_sink)
    }

    /// Enable forwarding of jet call traces to the provided sink.
    #[must_use]
    pub fn with_jet_trace_sink<F>(mut self, sink: F) -> Self
    where
        F: FnMut(Elements, &[Value], &Value) + 'a,
    {
        self.jet_trace_sink = Some(Box::new(sink));
        Self { ..self }
    }

    /// Enable the default jet trace sink that prints to stdout.
    #[must_use]
    pub fn with_default_jet_trace_sink(self) -> Self {
        self.with_jet_trace_sink(default_jet_trace_sink)
    }
    
    // ... 

    fn track_jet_call(
        &mut self,
        jet: &Elements,
        input_buffer: &[UWORD],
        output_buffer: &[UWORD],
        _: bool,
    ) {
        if let Some(sink) = self.jet_trace_sink.as_mut()
            && let (Ok(args), Ok(result)) = (
                parse_args(*jet, input_buffer),
                parse_result(*jet, output_buffer),
            )
        {
            sink(*jet, &args, &result);
        }
    }

    fn track_dbg_call(&mut self, cmr: &Cmr, value: simplicityhl::simplicity::Value) {
        if let Some(sink) = self.debug_sink.as_mut()
            && let Some(tracked_call) = self.debug_symbols.get(cmr)
            && let Some(Either::Right(debug_value)) =
                tracked_call.map_value(&StructuralValue::from(value))
        {
            sink(debug_value.text(), &debug_value.value());
        }
    }

    fn is_track_debug_enabled(&self) -> bool {
        self.debug_sink.is_some()
    }
```

So if the user wants, they will be able to construct a structure for getting logs.

We should also make enough effort to write good Rust docs for the new function to make it as easy as possible for the user to see the logs of `prune` execution.

apoelstra:
I propose we do this https://github.com/BlockstreamResearch/rust-simplicity/issues/324

PR #325, opened 2025-12-06
Replace `ExecTracker` trait and add `PruneTracker` extension trait

### LINKED ISSUE #324: New `ExecTracker` API
I propose that we extend and thoroughly document the `ExecTracker` trait. Currently it has several issues:

* It can be used to track left and right but cannot guide pruning (or directly be used to understand pruning)
* It can track jets but no other calls
* When tracking jets, it provides the input and output frames as `&[UWORD]` slices with no guidance on how to interpret them
* It can track `dbg` constructions (which are just `assert`s) where the CMR of the hidden node has some meaning...
* ...but only when the `is_track_debug_enabled` flag is provided, which is weird and unergonomic, but needed to avoid the overhead of frame->value conversions
* It has basically no documentation

Obviously we should improve the documentation. Beyond that, I think I can fix all the issues with the following design: #325

apoelstra:
Adds the ability for the execution tracker to track every visited node, to view the "input" (top read frame) of every node and the "output" (top write frame after execution) of every terminal node. Allows it to read this data in the form of a bit iterator, which is much easier to work with than the `&[UWORD]` that the old jet interface provided. Allows tracking "debug" nodes without special-purpose methods that enable/disable costly conversions.

Also adds a `PruneTracker` extension trait which can be used to introspect or manipulate the pruning process.

See #323 and #324 for motivation.

Fixes #324.

### DIFF
--- a/src/bit_machine/tracker.rs
+++ b/src/bit_machine/tracker.rs
@@ -131,3 +131,73 @@ impl<J: Jet> ExecTracker<J> for NoTracker {
         }
     }
 }
+
+/// Tracker that just outputs all its activity to stderr.
+#[derive(Clone, Debug, Default)]
+pub struct StderrTracker {
+    exec_count: usize,
+    inner: SetTracker,
+}
+
+impl StderrTracker {
+    /// Constructs a new empty [`StderrTracker`], ready for use.
+    pub fn new() -> Self {
+        Self::default()
+    }
+}
+
+impl<J: Jet> ExecTracker<J> for StderrTracker {
+    fn visit_node(&mut self, node: &RedeemNode<J>, input: super::FrameIter, output: NodeOutput) {
+        let input_val = Value::from_padded_bits(&mut input.clone(), &node.arrow().source)
+            .expect("input from bit machine will always be well-formed");
+        eprintln!(
+            "[{:4}] exec {:10} {}",
+            self.exec_count,
+            node.inner(),
+            node.arrow()
+        );
+        eprintln!("       input {input_val}");
+        match output.clone() {
+            NodeOutput::NonTerminal => { /* don't bother describing non-terminal output */ }
+            NodeOutput::JetFailed => eprintln!("    JET FAILED"),
+            NodeOutput::Success(mut output) => {
+                let output_val = Value::from_padded_bits(&mut output, &node.arrow().target)
+                    .expect("output from bit machine will always be well-formed");
+                eprintln!("      output {output_val}");
+            }
+        }
+
+        if let crate::node::Inner::AssertL(_, cmr) = node.inner() {
+            // SimplicityHL, when compiling in "debug mode", tags nodes by inserting
+            // synthetic AssertL nodes where the "cmr" is actually a key into a lookup
+            // table of debug information. An implementation of ExecTracker within
+            // the compiler itself might do a lookup here to output more useful
+            // information to the user.
+            eprintln!("      [debug] assertL CMR {cmr}");
+        }
+
+        ExecTracker::<J>::visit_node(&mut self.inner, node, input, output);
+        self.exec_count += 1;
+        eprintln!();
+    }
+}
+
+impl<J: Jet> PruneTracker<J> for StderrTracker {
+    fn contains_left(&self, ihr: Ihr) -> bool {
+        if PruneTracker::<J>::contains_left(&self.inner, ihr) {
+            true
+        } else {
+            eprintln!("Pruning unexecuted left child of IHR {ihr}");
+            false
+        }
+    }
+
+    fn contains_right(&self, ihr: Ihr) -> bool {
+        if PruneTracker::<J>::contains_right(&self.inner, ihr) {
+            true
+        } else {
+            eprintln!("Pruning unexecuted right child of IHR {ihr}");
+            false
+        }
+    }
+}
--- a/src/bit_machine/mod.rs
+++ b/src/bit_machine/mod.rs
@@ -22,7 +22,9 @@ use crate::{Cmr, FailEntropy, Value};
 use frame::Frame;
 
 pub use self::limits::LimitError;
-pub use self::tracker::{ExecTracker, NoTracker, NodeOutput, PruneTracker, SetTracker};
+pub use self::tracker::{
+    ExecTracker, NoTracker, NodeOutput, PruneTracker, SetTracker, StderrTracker,
+};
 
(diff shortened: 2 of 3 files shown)


uncomputable (dismissed):
utACK [6e34c441aa46bf3d9631c8cd0c1afe3a089e3ff6](https://github.com/BlockstreamResearch/rust-simplicity/commit/6e34c441aa46bf3d9631c8cd0c1afe3a089e3ff6)

KyrylR:
I believe we can close this PR: https://github.com/BlockstreamResearch/rust-simplicity/pull/323 in favor of this one 

I like the change too, the only part that is missing IMO is a "default" (basic) implementation of the tracker that the end dev can use 

If we can add "default" tracker and mention it in the docs for the `prune_with_tracker` function, it would fully cover the initial need for opening the #323

apoelstra (commented):
On e22f8d9a3430d98da8f014d312e303c92bad635e successfully ran local tests; ready for review

KyrylR:
Here is a PoC implementation of tracker that I could consider a default one for the SimplicityHL:

```rust
impl ExecTracker<Elements> for DefaultTracker<'_> {
    fn visit_node(&mut self, node: &RedeemNode<Elements>, input: FrameIter, output: NodeOutput) {
        self.inner.visit_node(node, input.clone(), output.clone());

        match node.inner() {
            Inner::Jet(jet) => {
                if let Some(sink) = self.jet_trace_sink.as_mut()
                    && let NodeOutput::Success(output_iter) = output
                {
                    let mut output_frame = output_iter.clone();
                    output_frame.next();

                    let parsed_result = Value::reconstruct(
                        &StructuralValue::from(
                            SimValue::from_padded_bits(
                                &mut (output_frame.clone()),
                                &node.arrow().target,
                            )
                            .expect("output from bit machine will always be well-formed"),
                        ),
                        &target_type(*jet)
                            .resolve(|_: &AliasName| -> Option<ResolvedType> { None })
                            .expect("target type of built-in jet always resolves"),
                    );

                    let mut input_frame = input.clone();
                    input_frame.next();

                    if let Ok(args) = parse_args_from_frame(*jet, &mut input_frame)
                        && let Some(result) = parsed_result
                    {
                        sink(*jet, &args, &result);
                    }
                }
            }
            // Handle debug calls (encoded as AssertL with CMR as key into debug symbols)
            Inner::AssertL(_, cmr) => {
                if let Some(sink) = self.debug_sink.as_mut()
                    && let Some(tracked_call) = self.debug_symbols.get(cmr)
                {
                    let mut input_frame = input.clone();
                    input_frame.next();

                    if let Ok(input_val) =
                        SimValue::from_padded_bits(&mut input_frame, &node.arrow().target)
                        && let Some(Either::Right(debug_value)) =
                            tracked_call.map_value(&StructuralValue::from(input_val))
                    {
                        sink(debug_value.text(), debug_value.value());
                    }
                }
            }
            _ => {}
        }
    }
}

fn parse_args_from_frame(jet: Elements, input_frame: &mut FrameIter) -> Result<Vec<Value>> {
    let jet_source_types = source_type(jet);
    if jet_source_types.is_empty() {
        return Ok(vec![]);
    }

    let arguments_blob = SimValue::from_padded_bits(input_frame, &jet.source_ty().to_final())
        .expect("input from bit machine will always be well-formed");

    let mut args = Vec::with_capacity(jet_source_types.len());
    collect_args(&arguments_blob.as_ref(), args.capacity(), &mut args)?;

    Ok(args
        .into_iter()
        .zip(jet_source_types.iter())
        .map(|(argument, aliased_type)| {
            Value::reconstruct(&argument.into(), &resolve_type(aliased_type))
                .expect("correct structure of the value of compiled program is guaranteed")
        })
        .collect())
}

/// Traverses a product and collects the arguments.
fn collect_args(node: &ValueRef, num_args: usize, args: &mut Vec<SimValue>) -> Result<()> {
    if num_args == 0 {
        return Ok(());
    }

    if num_args == 1 {
        args.push(node.to_value());

        Ok(())
    } else if let Some((left, right)) = node.as_product() {
        args.push(left.to_value());

        collect_args(&right, num_args - 1, args)
    } else {
        Err(anyhow!(
            "unexpected value structure while collecting arguments"
        ))
    }
}

fn resolve_type(aliased_type: &AliasedType) -> ResolvedType {
    aliased_type
        .resolve(|_: &AliasName| -> Option<ResolvedType> { None })
        .expect("target type of built-in jet always resolves")
}
```

apoelstra:
>Yeah, though, the thing that we are doing here is going to be used there, so this what I wanted to test

Okay, but there's nothing I can do in this PR to do this. I can open a draft PR to SimplicityHL (or you can, using your above code).

The reason you need to advance a bit is because the `AssertL` combinator is actually a `Case` combinator, which takes a bit of input to decide which branch to take. But this bit is "meaningless" and always 0 because it's an assertion.

KyrylR:
Overall I like the change, it also made a few things clearer for me, thanks

src/bit_machine/tracker.rs
```diff
@@ -0,0 +1,205 @@
...
+
+impl<J: Jet> ExecTracker<J> for StderrTracker {
+    fn visit_node(
+        &mut self,
+        node: &RedeemNode<J>,
+        mut input: super::FrameIter,
+        output: NodeOutput,
+    ) {
+        let input_val = Value::from_padded_bits(&mut input, &node.arrow().source)
+            .expect("input from bit machine will always be well-formed");
```
KyrylR:
```suggestion
    ) {
    	self.inner.visit_node(node, input.clone(), output.clone());
        
        let input_val = Value::from_padded_bits(&mut input, &node.arrow().source)
            .expect("input from bit machine will always be well-formed");
```

Is it ok, that we do not have `self.inner.visit_node(node, input.clone(), output.clone());` here?

apoelstra: oo, no, it's not! Good catch.

apoelstra:
Addressed @KyrylR's comments.

KyrylR (approved):

apoelstra (commented):
On ed79edf021343f58f14c272cb10ec9b827e3a66c successfully ran local tests

KyrylR:
ACK 3bf7cefdd1d4d15893f48e4a5114ecdea43d01fd

apoelstra (commented):
On 3bf7cefdd1d4d15893f48e4a5114ecdea43d01fd successfully ran local tests

PR #327, opened 2025-12-11
Bump `simplicity-lang` version

KyrylR:


KyrylR:
> We also need to do a major version bump because replacing the ExecTracker trait was a breaking change.

done 

> I'm not sure we need to break out all the benchmark changes since they're not visible to users of the release.

I just decided to go without it, as currently everything in the CHANGELOG is user related

> I'm also unsure whether we want to improve the Display output of Value first.

We can just publish this one so I can proceed with SimplicityHL. I already want to use this change in `simplicity-contracts`

KyrylR:
reopened here: https://github.com/BlockstreamResearch/rust-simplicity/pull/329

PR #329, opened 2025-12-11
Bump `simplicity-lang` version

KyrylR:


apoelstra:
> I fully regenerated the lock file... let me know whether I should revert to just changing the version of `rust-simplicity`

I'll see if I can build with this. In future please just update the version of rust-simplicity. (Though we should make some effort to keep Cargo-recent.lock up to date I guess. I think we should have a weekly CI cronjob for it.)

apoelstra:
Tagged and published.

PR #339, opened 2026-01-24
Add regression test for #337

apoelstra:
This will not pass CI until we fix the bug; this adds a unit test and a fuzz test which demonstrates the issue.

Note that #338 only fixes one of the two regression tests.


apoelstra:
Oops, I forgot about this. We fixed the bug. These regression tests and fuzztests are useful.

apoelstra:
Lol, maybe we didn't fix the whole bug. Let me revisit this.

PR #356, opened 2026-03-30
fix: get_padding for larger costs and padding lengths

delta1:
`Cost::get_padding` previously always assumed only 1 byte for compactsize encoding when calculating required padding size.

For larger differences in cost/budget, this incorrectly resulted in an additional 1 or 2 bytes of padding depending on the difference.

I found this when calculating padding for the SimplicityHL hash loop example, where rust-simplicity was calculating a 7426 byte annex padding while libsimplicity required a 7424 byte padding, since the compactsize encoding requires 2 additional bytes.

See ElementsProject/elements#1539

### DIFF
--- a/src/analysis.rs
+++ b/src/analysis.rs
@@ -152,14 +152,33 @@ impl Cost {
             return None;
         }
 
-        // Two bytes are automatically added to the encoded witness stack by adding the annex:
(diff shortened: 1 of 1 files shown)


apoelstra (approved):
ACK 7871509198ebc386694148f461d6cc1e40fadcce; successfully ran local tests

PR #357, opened 2026-04-01
Lock `human_encoding` module under a feature

ivanlele:
This PR introduces a new feature flag, `human_encoding`, which locks the module of the same name.

It is a followup to the discussion in #354.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ default = ["elements"]
 elements = ["dep:elements", "bitcoin"]
 test-utils = ["simplicity-sys/test-utils"]
 serde = ["dep:serde", "bitcoin/serde", "elements/serde"]
+human_encoding = ["dep:logos"]
 
 [lib]
 name = "simplicity"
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -16,6 +16,7 @@ default = ["elements"]
 elements = ["dep:elements", "bitcoin"]
 test-utils = ["simplicity-sys/test-utils"]
 serde = ["dep:serde", "bitcoin/serde", "elements/serde"]
+human_encoding = []
```
apoelstra: In e0a6a621302233be09d460afe0696338ebfbdb51:

This should have `dep:logos` in it, right?

Cargo.toml
```diff
@@ -40,6 +41,7 @@ getrandom = { version = "0.2", features = ["js"] }
 simplicity-sys = { version = "0.6.2", path = "./simplicity-sys", features = [
     "test-utils",
 ] }
+logos = "0.15"
```
apoelstra: In 774aa7b6f6460d01d653b7027d62cede2af2b04d:

We should just feature-gate those tests (I know there are a lot of them, but that's okay because we mostly test with `--all-features`) behind the `human_encoding` feature.

ivanlele: The deed is done

apoelstra (approved):
ACK a86927efcf3552bafd6e889643645a57cc92d9c7; successfully ran local tests; thanks for the quick iteration!

PR #360, opened 2026-04-28
Migrate execution to dynamic dispatch

ivanlele:
This PR closes #349 by moving all non-object-safe Jet supertraits onto Box<dyn Jet> and migrating the project’s internal generic model to dynamic dispatch.

### DIFF
--- a/src/node/commit.rs
+++ b/src/node/commit.rs
@@ -12,31 +12,27 @@ use super::{
 };
 
 use std::io;
-use std::marker::PhantomData;
 use std::sync::Arc;
 
 #[derive(Copy, Clone, PartialEq, Eq, PartialOrd, Ord, Debug, Hash)]
-pub struct Commit<J> {
+pub struct Commit {
     /// Makes the type non-constructible.
     never: std::convert::Infallible,
-    /// Required by Rust.
-    phantom: std::marker::PhantomData<J>,
 }
 
-impl<J: Jet> Marker for Commit<J> {
-    type CachedData = Arc<CommitData<J>>;
+impl Marker for Commit {
+    type CachedData = Arc<CommitData>;
     type Witness = NoWitness;
     type Disconnect = NoDisconnect;
     type SharingId = Ihr;
-    type Jet = J;
 
-    fn compute_sharing_id(_: Cmr, cached_data: &Arc<CommitData<J>>) -> Option<Ihr> {
+    fn compute_sharing_id(_: Cmr, cached_data: &Arc<CommitData>) -> Option<Ihr> {
         cached_data.ihr
     }
 }
 
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
-pub struct CommitData<J> {
+pub struct CommitData {
     /// The source and target types of the node
     arrow: FinalArrow,
     /// The IMR of the node if it exists. This is distinct from its IHR; it is
@@ -48,13 +44,9 @@ pub struct CommitData<J> {
     /// The IHR of the node if it exists, meaning, if it is not (an ancestor of)
     /// a witness or disconnect node.
     ihr: Option<Ihr>,
-    /// This isn't really necessary, but it helps type inference if every
-    /// struct has a \<J\> parameter, since it forces the choice of jets to
-    /// be consistent without the user needing to specify it too many times.
-    phantom: PhantomData<J>,
 }
 
-impl<J: Jet> CommitData<J> {
+impl CommitData {
     /// Accessor for the node's arrow
     pub fn arrow(&self) -> &FinalArrow {
         &self.arrow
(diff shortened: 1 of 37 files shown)


ivanlele:
@apoelstra  The main voodoo lives in `stc/bit_machine/mod.rs`:

```rust
let typed_jet = jet
    .as_ref()
    .as_any()
    .downcast_ref::<JE::Jet>()
    .ok_or(ExecutionError::JetTypeMismatch)?;
jet_result = self.exec_jet(typed_jet, env);
```

Here we downcast the `dyn Jet` into a concrete type so it can be matched inside `c_jet_ptr`.

With the move to dynamic dispatch, we lose some benefits, a few places now require additional generic boilerplate (prime example is: `RedeemNode::decode::<_, _, Core>`). But, we also received some: working with plain Node types becomes simpler since they no longer carry generics.

In a follow up PR, I'm thinking to  add runtime checks for `dyn Jet` within `Context`, to somehow compensate for the loss of compile time checks (runtime performance will take a hit unfortunatly)

apoelstra:
At a conference and have only done a cursory review so far. Overall this looks great! Amazing to get rid of so many generics. Frustrating to have extra generics on `Redeem::decode` but we expected this (and conceptually I think it's necessary to do it this way, or an equivalent way like taking a `decode_jet` closure).

> @apoelstra The main voodoo lives in `stc/bit_machine/mod.rs`:

Neat. We can revisit this in a followup. Maybe we can remove the generic from `BitMachine` somehow by doing dangerous things with c jet ptrs? Unsure. Would likely require more changes on the Haskell side. Let's definitely not try to do it in this PR, which is already super big, so for now this solution is good.

> In a follow up PR, I'm thinking to add runtime checks for dyn Jet within Context

Yeah, this is a good idea (but should be in a followup). We may want a `JetType` enum or something which can be returned from the `DynJet` trai. Unsure.

src/node/commit.rs
```diff
@@ -179,52 +170,51 @@ impl<J: Jet> CommitNode<J> {
...
     pub fn unfinalize_types<'brand>(
         &self,
         inference_context: &types::Context<'brand>,
-    ) -> Result<Arc<ConstructNode<'brand, J>>, types::Error> {
-        struct UnfinalizeTypes<'a, 'brand, J: Jet> {
+    ) -> Result<Arc<ConstructNode<'brand>>, types::Error> {
+        struct UnfinalizeTypes<'a, 'brand> {
             inference_context: &'a types::Context<'brand>,
-            phantom: PhantomData<J>,
+            phantom: PhantomData<()>,
```
apoelstra: In f33074e744b84cfd1b1804fe2b8d0421212768ce:

Lol. You should just delete this phantomdata.

src/node/construct.rs
```diff
@@ -73,42 +72,42 @@ impl<'brand, J: Jet> ConstructNode<'brand, J> {
...
         self.finalize_types_non_program()
     }
 
     /// Convert a [`ConstructNode`] to a [`CommitNode`] by finalizing all of the types.
     ///
     /// Does *not* sets the source and target type of this node to unit.
-    pub fn finalize_types_non_program(&self) -> Result<Arc<CommitNode<J>>, types::Error> {
-        struct FinalizeTypes<J: Jet>(PhantomData<J>);
+    pub fn finalize_types_non_program(&self) -> Result<Arc<CommitNode>, types::Error> {
+        struct FinalizeTypes(PhantomData<()>);
```
apoelstra: In f33074e744b84cfd1b1804fe2b8d0421212768ce:

Same here (drop the phantom)

src/node/construct.rs
```diff
@@ -386,52 +365,47 @@ impl<'brand, J> CoreConstructible<'brand> for ConstructData<'brand, J> {
...
             arrow: Arrow::witness(inference_context, NoWitness),
-            phantom: PhantomData,
         }
     }
 }
 
-impl<'brand, J: Jet> JetConstructible<'brand, J> for ConstructData<'brand, J> {
-    fn jet(inference_context: &types::Context<'brand>, jet: J) -> Self {
+impl<'brand> JetConstructible<'brand> for ConstructData<'brand> {
+    fn jet(inference_context: &types::Context<'brand>, jet: &dyn Jet) -> Self {
```
apoelstra, line 391: In f33074e744b84cfd1b1804fe2b8d0421212768ce:

Let's do this in a followup, but I'd like to explore making the `jet` function parametric over `J: Jet` rather than taking a `dyn Jet`.

src/node/commit.rs
ivanlele: Done, and the next one too. Probably should've worn my glasses when I was going over it one last time :d

apoelstra (approved):
ACK 0fb7c510b5f28a3b85a82a9436ba57e3cc7a8c4c; successfully ran local tests

PR #361, opened 2026-05-04
Runtime jet type check

ivanlele:
This PR introduces a runtime jet type check in the Context to prevent unintended behavior.


ivanlele:
@apoelstra At first I thought about returning a concrete error type, but the `jet` function is pretty low-level, so that would mean changing all callers to handle a `Result`. I can do that if you think it’s better.

The current approach is still better than what we had before, we catch type mismatches at construction time instead of during execution.

apoelstra:
Yeah, I think panicking is fine. It's definitely a programming error if jets wind up getting mixed. We'll likely need some experience with the API to figure out what tools we need to provide to minimize chances of this error.

apoelstra (approved):
ACK 24ce63a19d393ac8fef360ed7b8d965fa1d3ae73; successfully ran local tests

PR #362, opened 2026-05-05
fix(bitmachine): fix frame data offsets send to tracker (fixes Execution reached a pruned branch)

stringhandler:
Fixes some bugs where the data sent from the bitmachine to trackers would not start at the current cursor, resulting in some random junk being seen by the tracker. This doesn't have an effect on full programs, but when pruning this seems to generate an incorrect pruning, as seen in #337. 

The following code now prunes and runs correctly:
```rust 
fn add(elt: u32, acc: u32) -> u32 {
    let (_, sum): (bool, u32) = jet::add_32(elt, acc);
    sum
}

fn main() {
    let list: List<u32, 4> = list![5];
    let result: u32 = fold::<add, 4>(list, 0);
    assert!(jet::eq_32(result, 5));
}
```

Whereas previously it would fail on `Execution failed: Execution reached a pruned branch: fd6503b9fd020b8d4ed75deb14cc05bbcb91da392ba4b675f2b2345dabb1cde5`

Some excerpts from StderrTracker:
```
// Before fix
[  45] exec iden 2^32 → 2^32
       input 0x00000002
       output 0x00000005
       
// After fix
   [  45] exec iden 2^32 → 2^32
       input 0x00000005
       output 0x00000005

```

```
/// Before fix
[  46] exec iden (2^64? × 2^32?) × 1 → (2^64? × 2^32?) × 1
       input ((L(ε),R(0x00000005)),ε)
       output ((L(ε),L(ε)),ε)
       
// After fix
[  46] exec iden (2 × 2^32?) × 1 → (2 × 2^32?) × 1
       input ((0b0,R(0x00000005)),ε)
       output ((0b0,R(0x00000005)),ε)
```       
     

I haven't tried it on KyrylR 's specific SHRINCS test on that issue.

NOTE: I have based this off an older commit so I could test it with SimplicityHL, which doesn't compile with the latest `rust-simplicity`

NOTE: the unit test is written by Claude (Happy to remove it if needed)

### DIFF
--- a/src/bit_machine/mod.rs
+++ b/src/bit_machine/mod.rs
@@ -277,8 +277,11 @@ impl BitMachine {
         }
 
         'main_loop: loop {
-            // Make a copy of the input frame to give to the tracker.
+            // Capture read and write frames before the node action, to give to the tracker.
+            // The read frame cursor is where the node's input begins.
+            // The write frame cursor is where the node's output begins.
             let input_frame = self.read.last().map(Frame::shallow_copy);
+            let output_frame = self.write.last().map(Frame::shallow_copy);
             let mut jet_result = Ok(());
 
             match ip.inner() {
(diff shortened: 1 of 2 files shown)


apoelstra:
In a4cbd56339eeb387d4db89e9c0cabb395492cd0e:

Let's just delete `as_bit_iter`. In the one place it's still called it's right after `out_frame.reset_cursor()` so it's equivalent to the new `as_bit_iter_from_cursor`.

src/bit_machine/mod.rs
```diff
@@ -277,8 +277,11 @@ impl BitMachine {
         }
 
         'main_loop: loop {
-            // Make a copy of the input frame to give to the tracker.
+            // Capture read and write frames before the node action, to give to the tracker.
+            // The read frame cursor is where the node's input begins.
+            // The write frame cursor is where the node's output begins.
             let input_frame = self.read.last().map(Frame::shallow_copy);
+            let output_frame = self.write.last().map(Frame::shallow_copy);
```
apoelstra, line 284: In a4cbd56339eeb387d4db89e9c0cabb395492cd0e:

I think this new `output_frame` binding is unnecessary, but I like having it, because it borrows `self.write` for almost the entire interpreter function, guaranteeing that nothing in there modifies `self.write`.

stringhandler:
Updated. I left the name as `as_bit_iter_from_cursor` instead of renaming it to `as_bit_iter()` which may have been cleaner, but this is more explanatory

apoelstra:
In 1a121ba1ba00aa733f536e6d5d0476ad34b80a13:

Can you remove the reference to `as_bit_iter` from the doccomment? I think my local CI will fail on this since it will be rendered as a broken doc-link.

apoelstra (approved):
ACK 9cd34467e91e0f910c1077e1de549c893d29c66e; successfully ran local tests

PR #353, opened 2026-03-17
Revendor libsimplicity

ivanlele:
This PR vendors the latest version of [libsimplicity](https://github.com/BlockstreamResearch/simplicity/commit/d190505509f4c04b1b9193c6739515f9faa18aac)

PR #364, opened 2026-05-06
Bump simplicity version

ivanlele:


ivanlele:
@apoelstra, In the changelog, I tried to keep CI/bench/dev staff out

stringhandler:
There are a few places that say `rust_simplicity_0_6_0` in the source. Do we update these?

apoelstra:
Oops, good catch.

When we did #353 we should've updated the version of simplicity-sys. If we'd updated that version, all these constants would've changed.

We should open a separate PR which bumps the version of simplicity-sys and re-runs `./vendor-simplicity.sh` which should fix this.

ivanlele:
> We should open a separate PR which bumps the version of simplicity-sys and re-runs `./vendor-simplicity.sh` which should fix this.

Created PR #365 for this

ivanlele:
@apoelstra, with #362 merged and #365 approved (only the merge remains), do we have everything needed to roll out the new rust-simplicity version?

About your comment in BlockstreamResearch/SimplicityHL#224, I think the thing I suggested there might make the BlockstreamResearch/SimplicityHL#314 diff even worse. Without an updated rust-simplicity version to implement it, it would require adding many new generic parameters. So I think we first need the new rust-simplicity release before we can continuing iteration after BlockstreamResearch/SimplicityHL#322, which I made yesterday

apoelstra:
Sorry, I have been travelling, hence the delays in these PRs. I will try to get this in today so it's ready for work on Monday.

ivanlele:
> Sorry, I have been travelling, hence the delays in these PRs. I will try to get this in today so it's ready for work on Monday.

Np! Have a good road

apoelstra (approved):
ACK 00d36259c4a84acd060e1d090cee90ca41dbefbf; successfully ran local tests

apoelstra:
Tagged and published.

PR #365, opened 2026-05-06
Bump simplicity-sys to 0.7.0

ivanlele:


apoelstra:
Why 0.7.1 rather than 0.7.0?

ivanlele:
> Why 0.7.1 rather than 0.7.0?

Before was 0.6.2, so I assumed we also include patches

apoelstra:
Whenever we bump the major version we should reset the minor version to 0. So let's do 0.7.0 for this PR.

ivanlele:
> Whenever we bump the major version we should reset the minor version to 0. So let's do 0.7.0 for this PR.

Should be fine now

apoelstra (approved):
ACK 2f2736814226b5b4c1186e418ae2f58a1fc345bb; successfully ran local tests

apoelstra:
Tagged and published.

PR #369, opened 2026-06-17
fix: fix injl typo in core constructible

stringhandler:
Found a simple typo via LLM scanning. 

I don't think it's particularly dangerous as we would have seen this by now. I think this API is hardly used

### DIFF
--- a/src/merkle/cmr.rs
+++ b/src/merkle/cmr.rs
@@ -280,7 +280,7 @@ impl<'brand> CoreConstructible<'brand> for ConstructibleCmr<'brand> {
 
     fn injr(child: &Self) -> Self {
         ConstructibleCmr {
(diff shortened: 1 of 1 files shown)


delta1 (approved):
ACK 9bc2ae5b0d9ad7cf4f29c80d1a14dd44974ee808; tested locally

apoelstra:
Surprising that no unit tests (or actual usage) have been broken by this. This looks like a convenience API for the `policy` module. Amusingly, grepping through that module, it looks like we don't use `injl` or `injr` anywhere!

apoelstra (approved):
ACK 9bc2ae5b0d9ad7cf4f29c80d1a14dd44974ee808; successfully ran local tests

PR #370, opened 2026-06-17
fix: prevent silent overflow in calloc

stringhandler:
Minor overflow prevention.

Overflow can silently occur in release builds. Low priority but simple enough fix.


delta1 (approved):
ACK 11153bb423e4415b212569b1f5a189e93c31b6a8; tested locally

apoelstra:
In 11153bb423e4415b212569b1f5a189e93c31b6a8:

In theory this is still bad because it will allocate less memory than the user expects, which could lead to writing past the end of the buffer. (Or could it? I suppose that the user can't actually write more memory than is addressable, which this saturation should limit at..)

apoelstra (approved):
ACK 11153bb423e4415b212569b1f5a189e93c31b6a8; successfully ran local tests

stringhandler:
It gets `saturating_add`ed later anyway https://github.com/BlockstreamResearch/rust-simplicity/blob/160d782c6d18f93cd32d9624c18f526683a70122/simplicity-sys/src/alloc.rs#L52

PR #373, opened 2026-07-27
clean up Merkle Root types; update bitcoin_hashes to 1.1

apoelstra:
We had a bunch of type confusion in `Cmr`, `Amr`, etc., where we'd represent our merkle roots as `sha256::Midstate`s (which is okay, but will stop working with bitcoin_hashes 1.0 which adds a length counter to the midstate types) and then represent our IVs as the mr type (not okay, and required a bunch of macro hackery to make work).

We want the new bitcoin_hashes 1.0 midstate type, which aside from being more compatible with the upcoming rust-elements release, will let us convert midstates and engines directly into `Ctx8` Simplicity values. So move to that, and clean up the existing types along the way.


simplicity-sys/src/c_jets/c_env/elements.rs
```diff
@@ -148,8 +148,11 @@ extern "C" {
 }
 impl CTxEnv {
     pub fn sighash_all(&self) -> sha256::Hash {
+        // This is a little bit sketchy, directly interpreting a midstate as a sha256 hash. Bit
```
delta1, line 151: typo nit

```suggestion
        // This is a little bit sketchy, directly interpreting a midstate as a sha256 hash. But
```

delta1 (approved):
ACK 2d1f5603ee62c234cf4ef4fd15aa1da86f962ce3; tested locally

stringhandler:
ACK 2d1f5603ee62c234cf4ef4fd15aa1da86f962ce3

simplicity-sys/src/c_jets/c_env/elements.rs
apoelstra: I'm gonna leave this one be since it's an internal comment and the typo doesn't cause confusion. May fix in a followup.

apoelstra:
ACK 2d1f5603ee62c234cf4ef4fd15aa1da86f962ce3; successfully ran local tests

PR #374, opened 2026-08-04
some cleanups and new functionality for Tmr and Value

apoelstra:
Ahead of adding "quoting" functionality where we produce Simplicity fragments that compute the CMRs of other Simplicity fragments, constify a few functions, cleanup some others, and add `Value` constructors from SHA256 contexts.

The next PR will use the new bitcoin-hashes release from https://github.com/rust-bitcoin/rust-bitcoin/pull/6646 which will let us do some more cleanups and constification, and then we can start doing "real" changes.

### DIFF
--- a/src/jet/type_name.rs
+++ b/src/jet/type_name.rs
@@ -5,6 +5,7 @@
 //! Source and target types of jet nodes need to be specified manually.
 
 use crate::types::{self, Final, Type};
+use crate::Tmr;
 use std::cmp;
 use std::sync::Arc;
 
@@ -29,6 +30,18 @@ use std::sync::Arc;
 ///
 #[derive(Clone, PartialEq, Eq, PartialOrd, Ord, Hash, Debug)]
 pub struct TypeName(pub &'static [u8]);
+
+impl PartialEq<Final> for TypeName {
+    fn eq(&self, other: &Final) -> bool {
+        self.tmr() == other.tmr()
+    }
+}
+
+impl PartialEq<TypeName> for Final {
+    fn eq(&self, other: &TypeName) -> bool {
+        self.tmr() == other.tmr()
+    }
+}
 
 impl TypeName {
     /// Convert the type name into a type.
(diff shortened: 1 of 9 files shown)


src/jet/type_name.rs
```diff
@@ -70,6 +83,40 @@ impl TypeName {
         }
     }
 
+    /// Compute the TMR of the type name.
+    pub fn tmr(&self) -> Tmr {
+        let mut stack = Vec::with_capacity(16);
+
+        for c in self.0.iter().rev() {
+            match c {
+                b'1' => stack.push(Tmr::unit()),
```
stringhandler, line 92: These repeated mappings might be a source of errors in future. Not how to clean it up though.

src/jet/type_name.rs
```diff
@@ -106,3 +153,26 @@ impl TypeName {
...
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::jet::{Core, Jet};
+
+    #[test]
+    fn all_jet_tmrs() {
+        for jet in &Core::ALL {
```
stringhandler, line 163: Core is probably sufficient, but there may be some types in Elements that are not in Core.

stringhandler (commented):
ConceptACK. 

Seems fine. Added some comments but they are not blocking

src/jet/type_name.rs
apoelstra: I considered that, but then I'd have to feature-gate the test and it seemed like the effort was not worth the additional coverage.

src/jet/type_name.rs
apoelstra: Yeah, Rust is really bad at expressing abstract algorithms. Not much we can do.

apoelstra:
On 7143d879002631f9d406efa3a1b7ff5b80679422 successfully ran local tests

apoelstra:

apoelstra:
ping

stringhandler:
ACK 7143d879002631f9d406efa3a1b7ff5b80679422 ran cargo test

PR #375, opened 2026-08-10
types: eliminate some blowups when finalizing and dropping types

apoelstra:
Improves `types::Type::finalize` so that it doesn't unshare the type DAG when finalizing. This is important because we need to finalize the types before we can get static bounds, meaning that we can't reject exponentially-sized programs until after this method has succeeded.

Along the way, because I was testing with an extremely deep program found by the Bitcoin Red Team, I ran into stack overflows dropping both `Node` and `types::Incomplete`. This PR replaces those with manually-written non-recursive impls. It also copies the display-limiting logic from `types::Type` to `types::Incomplete`.

### DIFF
--- a/src/types/incomplete.rs
+++ b/src/types/incomplete.rs
@@ -11,6 +11,7 @@
 use crate::dag::{Dag, DagLike, NoSharing};
 use crate::types::union_bound::PointerLike;
 
+use super::context::BoundRefSharing;
 use super::{Bound, BoundRef, Context};
 
 use std::fmt;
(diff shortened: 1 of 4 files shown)


apoelstra:
On 51992e555a0c3d8049c078d892f4bef86bf5a531 successfully ran local tests

apoelstra:

stringhandler (commented):
Quick scan looks fine.

src/types/incomplete.rs
```diff
@@ -56,6 +111,17 @@ impl fmt::Display for Incomplete {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         let mut skip_next = false;
         for data in self.verbose_pre_order_iter::<NoSharing>(None) {
```
stringhandler: Curios, why did we not pass in MAX_DISPLAY_DEPTH here?

apoelstra: Oh, nice, I just forgot.

apoelstra: Fixed.

apoelstra:
On 0400e4198d16931b02694efb4ca865b4052cf3c2 successfully ran local tests

apoelstra:
ping

stringhandler:
ACK 0400e4198d16931b02694efb4ca865b4052cf3c2 ran cargo test

PR #377, opened 2026-08-24
`CoreConstructible`: remove impl for `Arrow`, add `arrow()` accessor

apoelstra:
This PR changes the `CoreConstructible` trait from "a node type that can be constructed from child node(s)" to "a node type that can be constructed from a `types::Arrow` and child node(s)".

This is a conceptually important shift but hopefully not one that requires many code changes. There are a couple goals here:

* It cleans up a bunch of code, where we were using the `CoreConstructible` trait to represent many different kinds of things (nodes, type arrows, "hidden nodes" within policies). This enabled some code reuse, which we didn't even take advantage of, at the cost of some API friction. Lots of unused parameters and unused result return valuse.
* The existing trait is inherently tied to type inference (all the `unit`/`iden`/etc constructors are infallible while all the `comp`/`case`/etc ones return a `Result<Self, types::Error>`. So you can't implement some sort of custom node type that imposes additonal restrictions. This is fine, but we could take better advantage of this. Specifically, I want to introduce infallible constructors where the caller provides a `types::Arrow` and promises that the arrow is correct. (If the caller lies here they will create an ill-typed program which will then fail down the line when they attempt to finalize it or otherwise convert from a `ConstructNode` to a useful kind of node. You can't even serialize a `ConstructNode` so there's not much you can do here.)

Then next PR will introduce these new constructors along with a new `TypePromise` construction which takes advantage of it to do infallible node construction in cases where the Rust compiler knows that the types all match.

### DIFF
--- a/src/merkle/cmr.rs
+++ b/src/merkle/cmr.rs
@@ -1,7 +1,9 @@
 // SPDX-License-Identifier: CC0-1.0
 
 use crate::jet::Jet;
+#[cfg(feature = "elements")]
 use crate::node::{CoreConstructible, DisconnectConstructible, WitnessConstructible};
+#[cfg(feature = "elements")]
 use crate::types::{self, Error};
 use crate::value::Word;
 use crate::{FailEntropy, Tmr};
@@ -254,11 +256,13 @@ impl Cmr {
 /// Wrapper around a CMR which allows it to be constructed with the
 /// `*Constructible*` traits, allowing CMRs to be computed using the
 /// same generic construction code that nodes are.
+#[cfg(feature = "elements")] // only used by policy module
 pub struct ConstructibleCmr<'brand> {
     pub cmr: Cmr,
     pub inference_context: types::Context<'brand>,
 }
 
+#[cfg(feature = "elements")] // only used by policy module
 impl<'brand> CoreConstructible<'brand> for ConstructibleCmr<'brand> {
     fn iden(inference_context: &types::Context<'brand>) -> Self {
         ConstructibleCmr {
@@ -366,6 +370,7 @@ impl<'brand> CoreConstructible<'brand> for ConstructibleCmr<'brand> {
     }
 }
 
+#[cfg(feature = "elements")] // only used by policy module
 impl<'brand, X> DisconnectConstructible<'brand, X> for ConstructibleCmr<'brand> {
     // Specifically with disconnect we don't check for consistency between the
     // type inference context of the disconnected node, if any, and that of
@@ -379,6 +384,7 @@ impl<'brand, X> DisconnectConstructible<'brand, X> for ConstructibleCmr<'brand> 
     }
 }
 
+#[cfg(feature = "elements")] // only used by policy module
 impl<'brand, W> WitnessConstructible<'brand, W> for ConstructibleCmr<'brand> {
     fn witness(inference_context: &types::Context<'brand>, _witness: W) -> Self {
         ConstructibleCmr {
@@ -393,6 +399,7 @@ mod tests {
     use super::*;
 
     use crate::node::{ConstructNode, CoreConstructible};
+    use crate::types;
 
     use std::str::FromStr;
     use std::sync::Arc;
--- a/src/human_encoding/named_node.rs
+++ b/src/human_encoding/named_node.rs
@@ -8,8 +8,7 @@ use crate::node::{
     self, Commit, CommitData, CommitNode, Construct, ConstructData, Constructible as _, Converter,
     CoreConstructible as _, Inner, NoDisconnect, NoWitness, Node,
 };
-use crate::types;
-use crate::types::arrow::{Arrow, FinalArrow};
+use crate::types::{self, Arrow, FinalArrow};
 use crate::{encode, ConstructNode, Value};
 use crate::{BitWriter, Cmr, Ihr};
 
--- a/src/merkle/amr.rs
+++ b/src/merkle/amr.rs
@@ -2,7 +2,7 @@
 
 use crate::jet::Jet;
 use crate::merkle::compact_value;
-use crate::types::arrow::FinalArrow;
+use crate::types::FinalArrow;
 use crate::value::Word;
 use crate::{Cmr, Tmr, Value};
 use hashes::sha256::Midstate;
--- a/src/merkle/ihr.rs
+++ b/src/merkle/ihr.rs
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: CC0-1.0
 
 use crate::jet::Jet;
-use crate::types::arrow::FinalArrow;
+use crate::types::FinalArrow;
 use crate::value::Word;
 use crate::{Cmr, Tmr, Value};
 use hashes::sha256::Midstate;
--- a/src/node/commit.rs
+++ b/src/node/commit.rs
@@ -7,8 +7,8 @@ use crate::{encode, types, Value};
 use crate::{Amr, BitIter, BitWriter, Cmr, DecodeError, Ihr, Imr};
 
 use super::{
-    Construct, ConstructData, ConstructNode, Constructible, Converter, Inner, Marker, NoDisconnect,
-    NoWitness, Node, Redeem, RedeemNode,
+    Construct, ConstructData, ConstructNode, Converter, Inner, Marker, NoDisconnect, NoWitness,
+    Node, Redeem, RedeemNode,
 };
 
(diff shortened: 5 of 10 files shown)


stringhandler (approved):
utACK 8896e7861c8f6c6ab6bedccf9caad001d63e2189

Maybe we should add a CI job with `--no-default-features` to catch this in future?

src/node/hiding.rs
```diff
@@ -30,103 +36,144 @@ use crate::{Cmr, FailEntropy, HasCmr, Word};
...
         }
     }
 }
 
-// We need `N: CoreConstructible` to access the inference context.
-// Because of this, implementations of `{Jet, Disconnect, Witness}Constructible`
-// for `Hiding<N>` require `N: CoreConstructible`.
-impl<'brand, N: CoreConstructible<'brand>> From<N> for Hiding<'brand, N> {
+// We implement From<N> for arbitrary N, to avoid but the expectation is that this is
+// FIXME revisit this. does it compile?
```
stringhandler: Is this still relevant?

stringhandler (approved):
utACK 1fef62fa9c6f65ff015cb16ed758ed09c1b2e2fa

stringhandler:
> Maybe we should add a CI job with --no-default-features to catch this in future?

Added in #380

src/node/hiding.rs
apoelstra: Oops, no, this is several iterations old. Removed.

stringhandler:
utACK 617c6975c8f6070b4c1f5973e4605f82da26ebb8

delta1:
I reviewed with Claude, which has proposed a potential issue along with a [regression test](https://github.com/delta1/rust-simplicity/commit/7481f2d5f1c53c79c2bcb9e98e849846bf6fbb04) + [fix](https://github.com/delta1/rust-simplicity/commit/27615e89051c95409fc0f40f779a59b6c49ced0b) pushed to [my branch here](https://github.com/delta1/rust-simplicity/tree/2026-08/constructible) 

It seems to be a valid finding, but I'm not sure if the fix is the best way to do it? 

---

Regression: hidden-branch satisfaction produces a program whose CMR no longer matches the committed policy.

The Hiding rewrite implements `injl/injr/take/drop_/assertl/assertr/disconnect` via `map_ref(_result)`, which preserves a hidden child's CMR unchanged. But a simulated hidden subtree must update its CMR as it passes through each combinator — as the pre-rewrite code did (`hidden_cloned_ctx(Cmr::drop(*cmr))`, etc.). Without this, when an unsatisfiable branch is hidden and flows through drop_ (in `policy::serialize::or/thresh_summand`) and then a case that becomes an assertion, the assertion commits the wrong (unwrapped) CMR.

Impact: `policy.satisfy(...)` yields a program whose Merkle root differs from `policy.commit()`, i.e. it would spend a different contract than the one committed to. The satisfied program still executes, so the existing tests don't catch it.

Repro (asserting `commit().cmr() == satisfy().cmr()` for an Or with a hidden branch) passes on base and fails on this PR.

Fix: thread a `cmrfn: impl FnOnce(Cmr) -> Cmr` through `map_ref/map_ref_result` and apply the matching transform at each call site (`Cmr::injl/injr/take/drop`, `Cmr::case(...)` for `assertl/assertr`, `Cmr::disconnect`). With that, all tests (incl. the new regression test) pass and clippy is clean.

apoelstra:
> a case that becomes an assertion, the assertion commits the wrong (unwrapped) CMR

Ah yep. Thanks for the regression test. I'll think a bit and see if I can come up with a cleaner solution than Claude's "just fish a closure through the whole API".

apoelstra:
On 617c6975c8f6070b4c1f5973e4605f82da26ebb8 successfully ran local tests

apoelstra:
Ah, Claude's solution is not "fish a closure through the whole API". It adds an extra closure to a couple utility functions that already take closures and are localized to src/node/hiding.rs. Its fix is almost byte for byte identical to what I would've done (so I am going to take it without crediting it, though I will credit it for the regression test).

apoelstra:
On 6a3561c766729e4f2eae642c51ca06925590a1c6 successfully ran local tests

apoelstra:
ping. this is ready for review

apoelstra:

src/node/hiding.rs
```diff
@@ -30,103 +36,147 @@ use crate::{Cmr, FailEntropy, HasCmr, Word};
...
+    ) -> Hiding<'brand, M> {
+        use core::convert::Infallible;
+        match self.map_ref_result::<_, Infallible>(|node| Ok(mapfn(node)), cmrfn) {
+            Ok(res) => res,
+            Err(inf) => match inf {},
         }
     }
 
+    /// If the node is not hidden, apply a function to the underlying node. If it is hidden,
+    /// do nothing.
```
stringhandler, line 74: Is this comment correct, it's different to above.

src/node/hiding.rs
```diff
@@ -30,103 +36,147 @@ use crate::{Cmr, FailEntropy, HasCmr, Word};
...
+        &self,
+        mapfn: impl FnOnce(&N) -> Result<M, Err>,
+        cmrfn: impl FnOnce(Cmr) -> Cmr,
+    ) -> Result<Hiding<'brand, M>, Err> {
+        Ok(Hiding {
+            inner: match self.inner {
+                HidingInner::Node(ref n) => HidingInner::Node(mapfn(n)?),
+                HidingInner::Hidden { cmr, ref arrow } => HidingInner::Hidden {
+                    cmr: cmrfn(cmr),
+                    arrow: arrow.shallow_clone(),
```
stringhandler, line 86: In 36fc4ce28ef2c273d8ce3e5d0ebdbd5ff9d093fd

Should this not be `Arrow::hidden(arrow.inference_context())`, i.e. new vs clone.

stringhandler (commented):
In 36fc4ce28ef2c273d8ce3e5d0ebdbd5ff9d093fd

(LLM assisted review)

src/merkle/cmr.rs
```diff
@@ -379,7 +376,7 @@ impl<'brand, X> DisconnectConstructible<'brand, X> for ConstructibleCmr<'brand>
     fn disconnect(left: &Self, _right: &X) -> Result<Self, Error> {
         Ok(ConstructibleCmr {
             cmr: Cmr::disconnect(left.cmr),
-            inference_context: left.inference_context.shallow_clone(),
+            arrow: left.arrow.shallow_clone(),
```
stringhandler, line 379: In 6a3561c766729e4f2eae642c51ca06925590a1c6

This is different from all the other methods. I don't know it well enough to understand if this is a problem, now or in future, but maybe better to create a distinct `Arrow::disconnect`.

src/node/hiding.rs
apoelstra: Yes. `map_ref` just passes through to `map_ref_result` after doing a bit of `Infallible` boilerplate for the user. So they are essentially the same method and get the same doccomment.

src/node/hiding.rs
apoelstra: Yeah, I think so. I think `clone` is probably fine but it's hard to reason about. Better to just say "fresh hidden node means a fresh inference variable". This would also match `zip_ref`.

src/merkle/cmr.rs
apoelstra: Yeah, this code is wrong. Lemme see if I can just delete it or something.

The damn policy module..

apoelstra: Yeah I can just delete this block of code. When/if we add disconnect support to the Policy module we can put it back.

apoelstra:
Addressed both of @stringhandler's nits.  Ready for rereview.

src/merkle/cmr.rs
apoelstra: Oh, hah, I did actually add an `Arrow::disconnect` in a later commit (which would appear in a followup PR).

But I think it's actually better to delete this. It's not totally obvious to me what the semantics should be and it's easy to put back when we have time to think about it.

apoelstra:
NACK 6a3561c766729e4f2eae642c51ca06925590a1c6; needs changes from Mike

apoelstra:
On c5dd9fb331fea2e828e7b937b8c044a222ef854b successfully ran local tests

stringhandler:
utACK c5dd9fb331fea2e828e7b937b8c044a222ef854b

PR #379, opened 2026-08-25
fix: fix pruning not using max sharing

stringhandler:
Fix for https://github.com/BlockstreamResearch/SimplicityHL/issues/391

As described in the issue, there are cases when a pruned program still contains some unpruned nodes. 

The test included fails when the pruning uses Internal sharing rather than MaxSharing.

### DIFF
--- a/src/node/redeem.rs
+++ b/src/node/redeem.rs
@@ -428,9 +428,12 @@ impl RedeemNode {
         // 2) Prune out unused case branches.
         // Because the types of the pruned program may change,
         // we construct a temporary witness program with unfinalized types.
+        // Convert with `MaxSharing`: the tracker keys branch usage by IHR, so
+        // converting IHR-equal nodes separately makes each copy keep the branches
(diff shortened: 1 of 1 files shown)


apoelstra:
Can you run `cargo fmt` on this?

stringhandler:
done.

apoelstra:
Can you also split the unit test into a separate commit (fix first, then test, so that all commits pass, but locally I'll delete the fix to see if the test fails)?

stringhandler:
done.

apoelstra:
Nice, thanks! Okay, this is definitely a bug (I wasn't sure it was possible to construct an unshared `RedeemNode` from the public API, but clearly it is). Hopefully it fixes the downstream issue.

apoelstra (approved):
ACK 35a49132463a7812cf425ae5c19fddf9d01addf1; successfully ran local tests

PR #380, opened 2026-08-25
ci: add cargo hack to test all features

stringhandler:
Adds cargo hack to the Github job to check that all combinations of the features compile.


apoelstra:
Nice! Failure is fixed by #377.

FYI cargo-rbmt can do this as well, but cargo-hack is short and sweet and we don't have any "special" feature combinations in this library, nor do we need to prune the matrix for speed reasons.

apoelstra:
Drafting until #377 is in.

apoelstra:
Can you rebase on #377?

PR #381, opened 2026-08-26
cargo: bump rust-simplicity version

KyrylR:
Since 0.8.0, there are breaking public changes (quick AI summary):
- Final::two_two_n now returns Result instead of its previous behavior.
- Policy::Assembly / the Assembly fragment was removed.
- Public Elements/hash types move from Elements 0.25 and bitcoin_hashes 0.14 to Elements 0.27 and bitcoin_hashes 1.2.

---

Therefore I would go for the minor version bump

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,21 @@
-# Unreleased
+# 0.9.0 - 2026-08-26
 
-* Add `Value` and `Type` constructors for variable-length buffers and SHA256 `Ctx8` contexts;
-  change the signature of `Final::two_two_n` to return a result rather than panicking when `n`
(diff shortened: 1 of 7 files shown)


stringhandler:
I think @apoelstra is still in the middle of adding canonical CMRs so might be better to wait

apoelstra:
Yeah, I'd like to sneak #377 in there at least. I spoke to Kyrylo out of band and he said he could wait a bit. (But if not, it's no big deal to just do two releasese.)

apoelstra:
Is this a dupe of #383?

KyrylR:
Yes, closing

PR #382, opened 2026-09-11
simplicity sys: prepare simplicity-sys 0.8.0 release

KyrylR:
TL;DR: we are bumping because of braking change of bitcoin_hashes dep

Ran the vendor command with the following command:

```
nix shell nixpkgs#gnused nixpkgs#coreutils \                      
  --command bash ./vendor-simplicity.sh \
  ./depend /tmp/simplicity-vendor-check.PIsQKe/upstream
```

On this revision of Simplicity: https://github.com/BlockstreamResearch/simplicity/commit/ba967b1a0d57e069c37cd4041b8bae53f12fd456

It is my first time doing such bump, so I am not entirely convinced that it is necessary, but AI reasoning (which is below) looks legit to me 

## Summary by GPT-6 Astra 

Bump simplicity-sys to 0.8.0 ahead of the next Rust Simplicity release

[The hash dependency update](https://github.com/BlockstreamResearch/rust-simplicity/pull/373) moved bitcoin_hashes from 0.14 to 1.x. Since simplicity-sys exposes these types in its public API, this is a breaking change and requires a version bump

Update the versioned C symbols, corresponding Rust bindings, and Cargo linking metadata to the 0.8 namespace. Update the workspace dependency requirements and lockfile accordingly. The vendored Simplicity revision and C implementation remain unchange

Verified that rerunning the vendoring script produces no Git diff


apoelstra:
Can you also add a `pub use hashes;` to `src/lib.rs`?

I didn't realize that we were exporting `hashes` types in our public API, but we are. So we should let users type `simplicity_sys::hashes` to get "the version of `hashes` used by simplicity-sys".

apoelstra (approved):
ACK 731ae176b73b7ee8ef308111fc2cb2ca82b759b4; successfully ran local tests

KyrylR:
Tagged

KyrylR:
@apoelstra I am not an owner of the sys package, so I cannot publish it

apoelstra:
Deleted the tag, which should point to the merge commit (which does not exist yet).

apoelstra:
Tagged and published.

